Issue #6, opened 2016-12-06
Define item areas using CollisionArea2D instead of using image masks


StraToN:
Currently, Escoria only allows users to define "actionable" areas by creating TextureButtons with a Click mask image parameter set.
![Screen1_texturebutton](http://i.imgur.com/doccssR.png)
I personally use an Area2D node with CollisionArea2D child instead:
![Screen2_collisionarea2d](http://i.imgur.com/VRRRFqd.png)

Note that in my example above, the defined area is part of the background, it is not a pickable object. Of course if the object was pickable I'd use either a Sprite or a TextureButton.

While I understand the benefits of the current method, it has two important disadvantages:
- First, you need to create a texture for the item, even if it is simply a non-pickable part of the background.
- Second, the mask image is longer to edit than a simple CollisionPolygon2D directly editable in Godot editor.

Area2D with a CollisionArea2D can have the same behaviour as TextureButton, but a bit more flexible. 
Thus I propose 2 solutions:
1. Keep the TextureButton method and support the second method.
2. Don't use any of them and create a new "Item" node (using editor plugin) which would be a fusion of both methods (ie: 2 in 1)

Olteon:
Good point.

StraToN:
I'm currently working on this issue. Solution 1 is considered, which basically consists in a new "item_area2d.gd" script. Will PR this very soon.

PR #8, opened 2017-03-07
New scripts for background items, input management and player masking

StraToN:


device/globals/item_background.gd
```diff
@@ -0,0 +1,349 @@
...
+export var use_custom_z = false
+
+var anim_notify = null
+var anim_scale_override = null
+
+var ui_anim = null
+
+var event_table = {}
+
+#func _set(name, val):
```
akien-mga: It's probably best not to add commented out code in the first place?

StraToN: Well, that code is duplicated from global/item.gd. Plain copy/paste, so I didn't feel like modifying. 
However, it could be interesting to make some code cleaning everywhere.

StraToN:
Wait before merge, I fixed a bug in mask_light2d.gd which I need to commit first.

StraToN:
This PR should fix #6

PR #4, opened 2016-10-31
use tscn format

c-cesar:
As tscn format is the new default, and more conveniet for git, I converted the old scn files to new format.

PR #15, opened 2017-04-19
Support a secondary right-click action for items

fleskesvor:
This commit enables a simple two-click interface by exposing a secondary action on items. The clicked() function in the game node takes a third parameter, which is the event.button_index when called from item nodes, but defaults to 1 (BUTTON_LEFT) when called from the other .gd scripts which calls the function (background.gd and global_vm.gd).

The example scene has been updated to support this interface, with the exception of the box item, which currently supports three actions, so the puzzle would need changing.


fleskesvor:
I have also updated box.esc, since there's now only use+look actions, instead of the original pick_up+open+look, so trying to pick up the box or open the newspaper is no longer possible.

fleskesvor:
Added an example event script that can be added to the box by setting "use" and "look" on "Action" and "Secondary Action".

StraToN:
Very interesting indeed, I like it, it seems to behave a bit more like SCUMM, which is cool.
I'll test this now.

fleskesvor:
I've found out that the secondary action is never applied when used on inventory items, so that needs sorting out. I'm not able to combine items in the inventory either, but I'm not sure yet if I've configured that correctly from the editor.

StraToN:
@fleskesvor Do you want to correct it before I merge?

fleskesvor:
I won't be able to make any corrections for several hours, but either way is fine with me. If you believe those effects are mostly beneficial, I'm fine with you merging it now. Otherwise, I'll try to get a new commit in tonight (CET).

StraToN:
@fleskesvor Well, we're not in a hurry ;) Let's just wait so you can fix it nice and good.

fleskesvor:
I added secondary action to inventory items and added an example event script to the ice cream in the inventory. Decided to give the extra parameter in interact() in game.gd an extra parameter rather than set the value from everywhere it's called, since that's a lot of places. Default values work exactly the same as in Python, right? I've tested, and can't see that I've broken anything, but there seems to be lots of corner case handling in the scripts that I haven't grasped fully yet.

Maybe it would be better if clicked() also took a default value of MOUSE_LEFT for the button_index parameter, since we're only passing on the action button_index from one of the places the function is called from anyway. If so, background.gd and global_vm.gd wouldn't have to be changed at all.

fleskesvor:
Everythi

fleskesvor:
An issue that is more pronounced with this interface, that is already present in Escoria (on a fresh clone) with the standard verb interface, is that actions which trigger player dialogue needs an extra click to "reset", so to speak. Eg., if you look at the can with the look action enabled from the interface, you can't immediately look at the ice cream. You have to make the dialogue go away first. It feels a bit clunky once you're aware of it.

fleskesvor:
I gave the new input variable "button_index" in game.gd's `clicked()` function a default value (BUTTON_LEFT), as I'd already done with `interact()`. I also moved the changes in test.scn plus some test scripts to a separate test branch, [test/secondary-item-action](https://github.com/fleskesvor/escoria/tree/test/secondary-item-action), to reduce the number of changes in this PR, and thus make it easier to review.

StraToN:
@fleskesvor Can you fix the conflict? I'll try to fix #4 so we can have it working on Godot 3.0. I'd love to have this PR merge soon after ^^.

fleskesvor:
Yeah, I'll fix the conflict, tidy it up a bit and squash it into one commit. Sorry for forgetting about this one for so long.

fleskesvor:
I've fixed the conflict and cleaned up the commit history. I'm going to check if it's working as intended now.

fleskesvor:
I've updated my [test branch](https://github.com/fleskesvor/escoria/tree/test/secondary-item-action) too, and it works as intended in my testing.

One possible gotcha though, is that if 'use' is your secondary action, you must also use right-click to confirm that 'use' action when combining items, which might not be intuitive. It's the reason why I've defined 'look' as the secondary action and 'pick_up'/'use' as the primary action in my test branch.

PR #19, opened 2017-04-27
Programmatically set RectangleShape2D on background area

fleskesvor:
The new background_area and item_background area scripts are great, but using them is more difficult than it needs to be. This is especially problematic since the new features are mostly undocumented, but having to first set the shape, then figure out how to set a transform and extents is also tricker than the old method of using a TextureFrame for background input, which only requires the user to attach a script.

What this addition does is to automatically set the shape on the Area2D which is used for background input, given that a node named "background" exists. For an example use case, see https://github.com/fleskesvor/escoria/tree/feat/simpler-background-area-test, which also uses the item_background script to make the mountain in the test scene respond to user input.

### DIFF
--- a/device/globals/background_area.gd
+++ b/device/globals/background_area.gd
@@ -25,9 +25,6 @@ func _enter_tree():
 		var extents = Vector2(size.x / 2, size.y / 2)
 		var transform = Matrix32(Vector2(1, 0), Vector2(0, 1), extents)
 
-		printt(extents)
-		printt(transform)
-
 		var shape = RectangleShape2D.new()
 		shape.set_extents(extents)
 		add_shape(shape)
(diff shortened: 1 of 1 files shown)


StraToN:
Very nice !
I need to verify these scripts actually work together right: I encounter some problems with ```item_background.gd``` as input clicks are actually catched by both ```background_area.gd``` AND ```background_item.gd```. The result is, if you left-click on a background item, the player moves to its interact_pos. But if you click on it a second time, the player moves to mouse input position.

fleskesvor:
Thanks. :) 

Yeah, I've noticed that too. Might be worth opening an issue about it, to highlight the problem, and perhaps get some eyes on it?

PR #32, opened 2017-09-02
Merge item_background.gd into item.gd

### LINKED ISSUE #16: Rework item.gd and item_background.gd to avoid duplicated code
These 2 scripts duplicate code, that is bad. We should simplify using an item_base.gd script to be extended by item_background.gd.

fleskesvor:
Fixes #16.

See [test/merge_item_scripts](https://github.com/fleskesvor/escoria/tree/test/merge_item_scripts) for updated test scene with TextureFrames and an Area2D using the same item.gd script.

This PR also fixes the path to terrain.gd.


akien-mga:
CC @StraToN. Looks like a good simplification.

StraToN:
All good indeed, that's just what we needed actually. `item.gd` can now be applied on both Area2D nodes and also TextureFrame nodes.

Issue #30, opened 2017-08-23
Multiple Talking Emotions


lavaduderDev:
I've been using set_state for my emotions, but every time my character uses the default "talk" animation. There are work arounds for this (like using set_state every time a character speaks/stops speaking.), but no default way to have multiple talk emotions.

Will you add the ability for multiple talk animations?

punto-:
Is this on the player character, or on a character that uses the "item.gd"
script? On an item, you can change the name of the talk animation on the
item, using the state animation. It's the "talk_animation" property on the
item script.

lavaduderDev:
Sorry for the long wait. This is for both the player and item.

Okay what I mean by multiple emotions.
I have one character, 
He has a default animation and default tallking animation. 
I also have a angry animation, and a angry talking animation.
I want to be able to interchange the default & angry talking animations in esc script.
That way when the character is done being angry. He will look normal.

When you say talk_animation property, are you talking about the script variable "talk_animation", because I can only set it to ONE talking animation at a time.

fleskesvor:
This should be solved by #147. You can add a second `AnimationPlayer`, "angry", with your angry sprites and use `set_state player angry` to make the player angry. Change back using eg. `set_state player animation`.

PR #33, opened 2017-09-19
Experimental support for player costumes

fleskesvor:
The SCUMM engine supported [costumes](http://wiki.scummvm.org/index.php/SCUMM/Technical_Reference/Costume_resources). I've been thinking that it should be relatively simple to support something similar in Escoria, using Godot's animation player, so I've made a rudimentary implementation by adding a new AnimationPlayer to the test player character and the necessary functions to trigger a costume change from an .esc script.

I've added the new images as a new set of animation frames within the same AnimatedSprite, and the key thing to note with this implementation is that one must add key frames on the animation attribute too, so that it's possible to switch between the old "default" set and the new "hat" set that I added.

As I see it, this wouldn't just be useful for changing actual costumes, like when Guybrush puts on a dress in LeChuck's Revenge, but it could also be used for changing player emotions, like requested in #30.

To test this, just start the test scene and "USE" the can to put on the hat. Then "USE" the ice cream to take off the hat again.

What doesn't work so well right now is that the costume change isn't immediate. I'm guessing that it's because the player is standing still when triggering the action, which presumably halts processing somewhere, but I haven't started digging into that yet.


fleskesvor:
Note: I don't think this should be merged as-is, but I put it up to get a few more eyes on it.

fleskesvor:
Costume change is now immediate. The problem was pretty obvious really: I needed to play an idle animation from the new set after the player had changed their costume.

punto-:
I like this approach, why not put 2 parameters on set_costume, to reference
the object that changes costume? This should work on items too.. It would
also remain consistent with almost all the other esc commands, which take
the object as first parameter.

set_costume player hat

we would also need a way to save this in the savegame when the costume is
not the default

fleskesvor:
That's a good idea to have a first parameter reference the object. It would also be compatible with a future multiple playable characters scenario.

The item script doesn't keep a reference to the animation node, but instead fetches it by node name every time it's used, so it would be a lot of (trivial) changes. There's also the `set_state()` function in item, which changes which animation is currently played. Which use cases are there for items where a "costume change" would be better?

Note, I'm not against doing this for items too, I'm just trying to understand how it should be implemented. There's also some animation stuff in interactive.gd, which item.gd extends (and maybe player.gd should do too), so maybe this is generic enough that it should go there.

Yeah, good point; the costume needs to be saved too. If it's just the one player, we could just find that  and check which animation is set, and if not the default, write the proper instructions to file. If items too, we'd need a more generic solution.

PR #39, opened 2017-12-05
Fix 3.0

### LINKED ISSUE #36: Missing files and directories after 3.0 update
Since the 3.0 update, the project only contains the files which were previously under the device/ directory. This would be a good thing, in my opinion, since the previous structure was a bit confusing to use as a base for a new project, except some useful files are missing now too. Compare the 2.1 and 3.0 directory structure:

2.1:

    ├── device
    │   ├── game
    │   ├── globals
    │   ├── scenes
    │   │   └── test
    │   │       └── player
    │   ├── templates
    │   └── ui
    ├── doc 
    ├── tools
    └── translations

3.0:

    ├── game
    ├── globals
    ├── scenes
    │   └── test
    │       └── player
    ├── templates
    └── ui

Most notably, the README.md and LICENSE.md files and the doc/ and translations/ directories are missing. The tools/ directory is also missing, but I never actually used anything from there, so I'm not sure what they were used for.

fleskesvor:
This PR is intended to fix #36.

I've restored the original directory structure, but put the translations into `device/` to be able to add them from the editor. I also added the missing lightmap to the test scene and made a change to `achievements.gd`, since `has_singleton()` is no longer a valid function, though I'm not sure if `has_setting()` is equivalent.

I have not touched the scripts in `tools/`, and I'm not sure if they have been tested with Godot 3.0 yet.

**EDIT:** I will squash the commits before it's considered for merge. Not 100% sure what happened there.

**EDIT 2:** I'm actually not sure how to solve this. My local commit log looks fine. I tried to rebase this branch onto the last commit of master, but git is putting these old commits in before that one, so it just explodes.


fleskesvor:
After looking into it, I'm pretty sure this happened because I merged the changes from upstream into my fork, instead of using force pull/push to get it in sync.

I think I can fix this by cherry picking the last two commits from this branch onto master (after fixing its commit history) and then just squash them into one and then force push the branch to my origin/fix-3.0.

StraToN:
Great job !

Issue #46, opened 2018-02-08
Screen / camera limit issues


mjtorn:
Hi!

Started looking into remaking my Unity/AC prototype with Escoria and some things have already arisen, some may be accompanied with a pull request if/when I get that far.

Here's one I don't currently feel comfortable pull-requesting.

1)
`global_vm.gd` and `game.gd` make a lot of assumptions about the screen size. My proto was originally drawn for 1280x720 and Escoria is making it somewhat non-trivial to use this.

I was puzzled because I did configure this in `ProjectSettings` _as the documentation tells to_, but there seems to be a strong tendency to assume full-hd resolution. This is made all the more awkward by the fact I have a 4k setup at the office, and I certainly don't want a test window that big!

I'm pretty sure the correct solution would be to consult `ProjectSettings` first and only then the OS!

2)
The documentation could also point out how camera limits work. It may be obvious - at least in retrospect - but upper-left is (0, 0) and as my first scene (background) is 2060x720, the settings should probably be

```
x = 780
y = 0
w = 1280
h = 720
```

Currently the camera-limit part is vague enough to easily miss altogether!

I'm not entirely done with this yet, the stuff I get in my hard-coded 1280x720 window is not correct, maybe partly because...

3)
Godot3 renamed `pos` to `position`. You should be able to fix this faster than I can make a PR, so let's see who gets around to it first ;) But I changed some of these here and let's see where it goes.

(But I'll submit this now so I won't forget to do so, will update later if required)

Thanks!

mjtorn:
I won't be getting the camera limits going today, so I'll ask: should the camera limits essentially follow the limits of the background or should they be like the viewport to the scene or both of the above?

This is made a bit more problematic because the camera does not currently follow the player. Does this need the ESC script in place? Because I didn't get that far yet.

Also noted that `camera.set_limit(...)` is not documented for Godot 3 anymore, but the editor doesn't complain either, so I suppose a fix for this would include setting all the `limit_` attributes one by one.

mjtorn:
Somewhat related to this, what does the ProjectSettings Use Custom Camera do?

I get slightly better results if I turn it on, but this is not documented (afaict) anywhere and I feel very lost right now, guessing about the code and settings.

mjtorn:
I'm attaching screenshots of what it looks like when the camera is off.

The only changes:

  * `global_vm.gd` has the size hard-coded to 1280x720 because ProjectSettings is not honored
[...]

PR #47, opened 2018-02-09
Seems maybe Globals should be ProjectSettings here

mjtorn:
I saw a complaint here that `Globals` is not defined. Despite my lack of experience here, I think `ProjectSettings` is what should be used here.

Thoughts?

### DIFF
--- a/device/globals/action_menu.gd
+++ b/device/globals/action_menu.gd
@@ -14,7 +14,7 @@ func start(p_target):
 		target = p_target
 		target.connect("visibility_changed", self, "target_visibility_changed")
 
-	var scale = Globals.get("platform/action_menu_scale")
+	var scale = ProjectSettings.get("platform/action_menu_scale")
 	set_scale(Vector2(scale, scale))
 
 func stop():
--- a/device/globals/action_menu.gd
+++ b/device/globals/action_menu.gd
@@ -14,7 +14,7 @@ func start(p_target):
 		target = p_target
 		target.connect("visibility_changed", self, "target_visibility_changed")
 
-	var scale = ProjectSettings.get("platform/action_menu_scale")
+	var scale = ProjectSettings.get_setting("platform/action_menu_scale")
 	set_scale(Vector2(scale, scale))
 
 func stop():
--- a/device/globals/achievements.gd
+++ b/device/globals/achievements.gd
@@ -24,7 +24,7 @@ func award(aid):
 		return
 
 	printt("showing rate screen")
-	var url = iOS.get_rate_url(ProjectSettings.get("ios/app_id"))
+	var url = iOS.get_rate_url(ProjectSettings.get_setting("ios/app_id"))
 	vm.show_rate(url)
 
 	vm.settings.rate_shown = true
--- a/device/globals/dd_player.gd
+++ b/device/globals/dd_player.gd
@@ -7,7 +7,7 @@ func start(params, level):
 	else:
 		type = params[1]
 
-	type = type + ProjectSettings.get("platform/dialog_type_suffix")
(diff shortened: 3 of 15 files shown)


device/globals/action_menu.gd
```diff
@@ -14,7 +14,7 @@ func start(p_target):
 		target = p_target
 		target.connect("visibility_changed", self, "target_visibility_changed")
 
-	var scale = Globals.get("platform/action_menu_scale")
+	var scale = ProjectSettings.get("platform/action_menu_scale")
```
fleskesvor: Does it work?

According to the documentation the correct function seems to be `get_setting()`, but there's inconsistency between the example and the listed functions (`set()` vs `set_setting()`), so I'm not sure what's right.

mjtorn:
I actually don't have the action menu going yet, though this did suppress the complaints.

Now I pushed a commit that gets that setting as well, but because the documentation is quite lacking on these and there's no obvious substitute for `Globals.get_singleton(...)` I just guessed maybe `.get(...)`  would still work.

Off-topic for this PR but I'm still stuck trying to get the camera to behave at all, let's see if it goes anywhere before I give up for good.

device/globals/action_menu.gd
mjtorn: If you want, I can squash/fixup `get_setting`, but I went with `get` in the second commit for consistency. Because the Godot editor still autocompletes `get`, it can't be all wrong.

fleskesvor: Ok, that's probably still good then. I see that `get()` is also used pretty much everywhere in Escoria, so that sounds like a good call to me.

fleskesvor: Yeah, I think it would be good to squash the commits, but might as well wait and see if one of the maintainers have any suggested changes.

mjtorn: If `get(...)` is most commonly used, I usually prefer to stick to the fewest surprises and most consistency in my usual programming. Then it's fine to have a separate branch to change them all to `get_setting(...)` later if that becomes relevant. Not my project and I'm fine either way with this, just my two cents for whomever wants to comment.

fleskesvor:
I think `Globals.get_singleton()` might be a relic that is no longer useful, since you can just reference a singleton by name, but I'm not sure. I'm also pretty sure `get()` does still work, 

I'll take a look at your camera issue, but it's not something I'm very familiar with either.

device/globals/action_menu.gd
fleskesvor: I agree.

mjtorn:
@fleskesvor I added some screenshots to #46 which may be worth a thousand words of uneducated explanation, hope they help!

device/globals/action_menu.gd
StraToN: `get()` is actually a `Node` method. `ProjectSettings::get_setting()` actually directly calls `get()` as you can see here: https://github.com/godotengine/godot/blob/9f479f096cb50fa8d1215e68c262f110116114ef/core/project_settings.cpp#L854
`get()` won't disappear so there's no problem using `get()` for now. However, if at some point `get_setting()` evolves and performs other actions, then it would be cleaner to change every `ProjectSettings.get()` in Escoria in favor of `ProjectSettings.get_setting()`.

@mjtorn If you are ok, you can perform the change on the whole Escoria directory with this command (under Linux):
`find ./ -type f -exec sed -i -e 's/ProjectSettings.get\(/ProjectSettings.get_setting\(/g' {} \;`
That should do the trick :)

mjtorn: Pushed such a commit, thanks for the feedback! :)

fleskesvor:
If you rebase on master and squash your commits, I think this one should be good to go.

PR #48, opened 2018-02-09
Suppress complaints when terrain is null

mjtorn:
I don't know when or why `terrain` is unset, but it is.

Is there a severe underlying issue or is it ok just to suppress it?

Apparently a whitespace fix crept in for good measure ;)

### DIFF
--- a/device/globals/player.gd
+++ b/device/globals/player.gd
@@ -1,4 +1,4 @@
-tool 
+tool
 
 extends Node2D
 
@@ -221,6 +221,8 @@ func _check_bounds():
 		get_node("terrain_icon").show()
 
 func _update_terrain():
+	if !terrain:
+		return
 	var pos = get_position()
 	set_z_index(pos.y)
 	var color = terrain.get_terrain(pos)


fleskesvor:
It would be unset if you have a game scene with a player but not a terrain (like the first room of Daïza) or if the player and terrain don't share a parent.

In the first case, the player should be an item, visible or not, since none of the features of the player script are needed. This isn't intuitive at first, but a player is pretty much just an item with the ability to walk in Escoria.

mjtorn:
Thank you sincerely for your reply, but I don't understand :(

I have my `scene`, it has the children `background`, `terrain`, `player`, `action_menu`, `desk` (my attempt at having an item from another scene) and finally `game`.

All of these except `desk` have an appropriate script attached to them.

Did I miss something crucial in the documentation on why the terrain would not be there? FWIW the Godot Editor shows me a crossed-out tree, which maybe is the symbol for missing terrain, but I do have it, and it has the NavigationPolygonInstance child.

mjtorn:
Maybe I should have been more explicit also that it _does get set_ in `_ready`, if I add `print(...)` calls I see that, but at some point it gets unset. That's what I was pointing to, so I don't know when or why (or where, for that matter) it gets unset.

fleskesvor:
That is odd indeed. The crossed-out tree is used to indicate that the player is outside the bounds of a terrain, and should also appear in the editor when you move the player, since the player script is a tool script, and also runs in editor mode. That doesn't seem to work for me in the demo scene though, so maybe something broke in the migration from 2.1. The scene itself runs and appears to be working correctly.

If it's not too much trouble, could you attach an image of the editor with your scene, including the player and navigation polygon visible, as well as the expanded scene tree?

mjtorn:
@fleskesvor I think you can incidentally see this as well in the shots on #46

mjtorn:
I looked into why this might not have been merged, and noticed I don't get the error anymore.

I'll close this and throw the branch away, fix if it becomes a real problem.

fleskesvor:
I haven't looked into this, but I hope we at some point figure out what causes this. This solution works, but it feels more like masking than fixing problem. If it's a problem with the way the scene is set up, we need to offer better documentation and templates.

mjtorn:
Yeah, but I closed this PR for the reason that I can't reproduce it anymore.

It feels unfair to speculate that it's the complexity of the code that caused an accidental proper fix, but I have no better theory :/

Let's see if it pops up again!

fleskesvor:
Yeah, I agree. As long as it works for you now, I see no reason to chase after it until it pops up again.

PR #49, opened 2018-02-09
Update camera calls for Godot 3

mjtorn:
Though this does not fix the issues in #46 it does update using the `pos` attribute to `position` and also uses the new style of setting the limits.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -277,11 +277,11 @@ func set_camera_limits():
 			area = area.expand(pos)
 			area = area.expand(pos + size)
 
-		camera.set_limit(MARGIN_LEFT, area.position.x)
-		camera.set_limit(MARGIN_RIGHT, area.position.x + area.size.x)
+		camera.limit_left = area.position.x
+		camera.limit_right = area.position.x + area.size.x
 		var cam_top = area.position.y # - get_node("/root/main").screen_ofs.y
-		camera.set_limit(MARGIN_TOP, cam_top)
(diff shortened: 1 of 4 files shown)


device/globals/game.gd
```diff
@@ -290,10 +290,10 @@ func set_camera_limits():
...
 		cam_limit = area
 	else:
-		camera.set_limit(MARGIN_LEFT, camera_limits.pos.x)
-		camera.set_limit(MARGIN_RIGHT, camera_limits.pos.x + camera_limits.size.x)
-		camera.set_limit(MARGIN_TOP, camera_limits.pos.y)
-		camera.set_limit(MARGIN_BOTTOM, camera_limits.pos.y + camera_limits.size.y + get_node("/root/main").screen_ofs.y * 2)
+		camera.limit_left = camera_limits.position.x
+		camera.limit_right = camera_limits.position.x + camera_limits.size.x
+		camera.limit_top = camera_limits.position.y
+		camera.limit_bottom = camera_limits.position.y + camera_limits.size.y + get_node("/root/main").screen_ofs.y * 2
```
fleskesvor: Same as above.

mjtorn:
I replaced all of them to not introduce surprises. Not sure which of them I hit in testing, but seems to work!

fleskesvor:
Nice. 🙂 I think @StraToN replaced the calls to `get_tree().get_root().get_node("main")` with the simpler syntax during the migration to 3.0, but I guess this syntax changed a lot over time, and some variations were missed.

PR #51, opened 2018-02-12
[RFC] Set `current = true` on camera to follow player

mjtorn:
This does not fix all the problems in #46 but at least it places the camera on the player and respects camera limits, so that's something.

The camera is still zoomed wrong, but I'm almost sure this is a step in the right direction.

### DIFF
--- a/device/globals/game.tscn
+++ b/device/globals/game.tscn
@@ -32,7 +32,7 @@ autostart = false
 
 anchor_mode = 1
 rotating = false
-current = false
+current = true
 zoom = Vector2( 1, 1 )
 limit_left = -10000000
 limit_top = -10000000


StraToN:
I have no problem with this change. You simply set the default value of `current` so that the camera is directly active by default, which seems logicial for a newcomer. WDYT @fleskesvor ?

fleskesvor:
As long as it respects the bounds of the current scene, this sounds like a reasonable default to me. Does it work OK with the current demo scene too? If I remember correctly, the chainsaw has been outside of the viewport in my testing, so there has been no way to interact with it, since the camera doesn't scroll, so that might be fixed by this PR too.

In the future we could implement some kind of elasticity like Adventure Creator has, as described by @mjtorn in #46, but I agree that this PR sounds like an improvement over the current default.

mjtorn:
It seems to work a lot better with the camera limits/bounds, yes. If you change the zooms or such, it'll fail but I doubt it's this change's fault.

mjtorn:
As for the test scene, I had to set the Screen Size Override setting to true to have a proper size for the window, but the background on it is so small it doesn't really have an effect. I think the test scene may be broken somehow :/

StraToN:
Ok, I think it's ok to merge then. 
However I think the next important task is to clean out all Escoria's custom ProjectSettings entries and re-create them in a proper Escoria category. I'll create an issue for this.

PR #54, opened 2018-02-14
[RFC] ProjectSettings magic removal

mjtorn:
Clean up some of the settings

### DIFF
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -87,39 +87,10 @@ func settings_loaded(p_settings):
 	AudioServer.set_bus_volume_db(0, settings.sfx_volume)
 	TranslationServer.set_locale(settings.text_lang)
 	music_volume_changed()
-	update_window_fullscreen(true)
 	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "ui", "language_changed")
-
-func update_window_fullscreen(p_force = false):
-	if ProjectSettings.get_setting("escoria/debug/screen_size_override"):
-		return
-	if !p_force && (settings.fullscreen == OS.is_window_fullscreen()):
-		return
-	if settings.fullscreen:
-		OS.set_window_resizable(true)
-		OS.set_window_fullscreen(settings.fullscreen)
-		pass
-	else:
-		var title = ProjectSettings.get_setting("escoria/platform/window_title_height")
-		var sc = OS.get_current_screen()
-		var ratio = 1080 / 1920.0
-		var size = OS.get_screen_size(sc)
-		printt("***** got screen size ", size, title)
-		var h = size.y - title
-		if h / ratio > size.x:
-			size.y = size.x * ratio
-		else:
-			size.y = size.y - title
-			size.x = h / ratio
-		printt("setting window to size", size)
-		OS.set_window_fullscreen(settings.fullscreen)
-		OS.set_window_size(size)
-		#OS.set_window_position(Vector2(0, 0))
-		OS.set_window_resizable(ProjectSettings.get_setting("escoria/platform/screen_resizable"))
 
 func music_volume_changed():
 	emit_signal("music_volume_changed")
-
 
 func drag_begin(obj_id):
 	drag_object = obj_id
@@ -728,9 +699,7 @@ func _ready():
 	compiler = preload("res://globals/esc_compile.gd").new()
 	level = preload("res://globals/vm_level.gd").new()
 	level.set_vm(self)
-	game_size = Vector2()
-	game_size.x = ProjectSettings.get_setting("escoria/display/game_width")
-	game_size.y = ProjectSettings.get_setting("escoria/display/game_height")
+	game_size = get_viewport().size
 
 	scenes_cache_list.push_back(ProjectSettings.get_setting("escoria/platform/telon"))
 	scenes_cache_list.push_back(get_hud_scene())
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -91,7 +91,7 @@ func settings_loaded(p_settings):
 	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "ui", "language_changed")
 
 func update_window_fullscreen(p_force = false):
-	if ProjectSettings.get_setting("debug/screen_size_override"):
+	if ProjectSettings.get_setting("escoria/debug/screen_size_override"):
 		return
 	if !p_force && (settings.fullscreen == OS.is_window_fullscreen()):
 		return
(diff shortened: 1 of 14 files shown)


fleskesvor:
Yeah, maybe it's just as well to scrap that code for now, and rather remake other camera modes when/if needed. It will make the rewrite much easier if we eliminate some dead/non-functional code first.

What do you think, @StraToN?

device/globals/global_vm.gd
```diff
@@ -87,40 +87,11 @@ func settings_loaded(p_settings):
...
-			size.y = size.x * ratio
-		else:
-			size.y = size.y - title
-			size.x = h / ratio
-		printt("setting window to size", size)
-		OS.set_window_fullscreen(settings.fullscreen)
-		OS.set_window_size(size)
-		#OS.set_window_position(Vector2(0, 0))
-		OS.set_window_resizable(ProjectSettings.get_setting("platform/screen_resizable"))
-
```
fleskesvor: Did this not work at all?

mjtorn: If it did, I don't know what it was supposed to do 😱 

Godot resizes the window just fine, there's no need to go out to `OS` for it that I can see.

A valid use case for resizable would be when shrinking the window, but this would involve setting the viewport and I have no idea how to do it. I tried and gave up :( There should also be a sanity check that the window isn't larger than the camera limits or you get the gray background outside.

The least surprising code would be to deny resizing on the Escoria level and figure out a way to change resolutions/scaling through menus or something.

Is there a case for the opposite?

But this code doesn't seem to serve any purpose, though maybe it did in 2.1. @StraToN may be the leading expert in this matter. plz halp!

mjtorn:
I added two commits that help a bit with customization.

I tried asking around on irc, and looking online, but couldn't find any good information on how to deal with references within the scene tree and gave up.

To make working on Escoria easier, like not having to pick commits and worry about your game messing with the Escoria tree, this should be the general strategy.

Anyway, this branch should be mergable IMO. Comments?

device/globals/global_vm.gd
fleskesvor: It might be a remnant from an older version of Godot then. Yeah, I think @StraToN should take a look at this before it's merged, just in case.

fleskesvor:
What kind of references are you thinking of? If you are thinking of how to navigate the scene tree, you should find the information you need in 

http://docs.godotengine.org/en/3.0/getting_started/step_by_step/scene_tree.html
and 
http://docs.godotengine.org/en/3.0/classes/class_node.html

You can always find a node in the current scene by navigating from `get_node("/root")` or the equivalent `$"/root"`.

mjtorn:
`device/ui/hud.tscn` contains lines like these

```
[ext_resource path="res://globals/hud.gd" type="Script" id=1]
[ext_resource path="res://ui/inventory.tscn" type="PackedScene" id=2]
```

So basically even if the path to `hud.tscn` could be a setting in `ProjectSettings` (I don't know if it can, but these are very recursive anyway so my point is the same) can `res://ui/inventory.tscn` be a setting and `hud.tscn` look it up?

I don't think it can, which means copies need to be made and altered accordingly, or something, but it would be nice if this could be decoupled.

Did make it more understandable?

fleskesvor:
Ah, I see what you mean now. No, I don't think that's possible, but I'm thinking that `hud.tscn` isn't something a developer should need to touch, but the inventory and verb/action menu, etc. should be instanced from code and have their location defined in project settings, since those are likely to be modified by a developer.

mjtorn:
Yeah, maybe `hud.tscn` doesn't require that, but can we then have it so that `hud.gd` looks up the setting and injects it into `hud.tscn` ?

fleskesvor:
If we don't hear from @StraToN this weekend and you're eager to get this PR merged (I am), what I would be most comfortable with is:

- squash your commits from 14th February (d723a1c through 5d708f1) into a single "settings revamp" commit
- keep 9700ebe as it is. I'm not 100% sure about it, but I think it's a good idea, and that way it will be easy to change back if needed.
- change 1600726 to still check for `escoria/platform/use_custom_camera`, but set the default to `true` in the project settings (I think that requires a rebase with the previous commit hash and `edit` on this one, but I'm not 100% sure).
- squash the last three commits into another settings related commit

I'm sorry for bugging you with this, but I think it would give a cleaner and more meaningful graph, and it hopefully isn't too much work for you.

fleskesvor:
I did the proposed changes in my own branch, which you may force into your own, if you agree with the changes: [fleskesvor/escoria/commits/projectsettings-magic](https://github.com/fleskesvor/escoria/commits/projectsettings-magic)

This is my rebase:

`git rebase -i 6f583e6a6a4b8a71f7df8a554402b88e506a90db`

    pick d723a1c Collect escoria settings in one place
    squash 20996da Update get_settings for old application
    squash fdedf0d `print_clicked_control` was not referenced anywhere
    squash 3a55c9e Update get_setting for old debug
    squash 4868a41 Update get_settings for old display
    squash 6935d00 Update get_settings for old platform
    squash 17c5b2e Remove bunch of unreferenced settings
    squash 5d708f1 Update get_settings for old ui
    pick 9700ebe Prefer Godot's window handling code over old custom
    edit 1600726 Use regular camera instead of magic custom camera
    squash d3258f9 Implement unreferenced force_text_ids
    squash bf652bc Do not hard-code the path to in_game_menu
    squash 01b0043 confirm_popup does not have to be hard-coded either

mjtorn:
@fleskesvor ok, commits d723a1c through 5d708f1 are easily squashed and having had them separate has served its purpose. I may do that now because I don't know if it matters to @StraToN but the old commits are still available for reference and can be reverted and so on.

c0f9b5d has replaced what once was 9700ebe, because of amending the variable removal.

I dislike the idea of leaving "custom camera" in there. It's undocumented magic code that breaks everything. Naturally if you wish, I will comply, and setting it to `true` is a huge win. 

I will not diverge my fork from being all about PRs into something I primarily use, but because I feel strongly about this (for whatever reason :D) I feel I have to open an issue about deprecating it.

I get a notification about your rebase while typing things here, I'll investigate and see.

mjtorn:
`git reset gh-flesk/projectsettings-magic` ftw. I also re-edited in the removal of `game_size`, as stated in the new commit message.

fleskesvor:
> I dislike the idea of leaving "custom camera" in there. It's undocumented magic code that breaks everything. Naturally if you wish, I will comply, and setting it to `true` is a huge win.

It's not that I have strong feelings about it, but I find it difficult to tell if there are use cases where removing it might break things, and I'd prefer to let someone else make that call.

> I will not diverge my fork from being all about PRs into something I primarily use, but because I feel strongly about this (for whatever reason :D) I feel I have to open an issue about deprecating it.

This is a good idea.

fleskesvor:
The PR has some regressions for the test scene compared to previous commits, since it now runs in windowed mode with the scene positioned in the bottom right corner, but the current head of master does not work at all, so overall it's an improvement.

Thanks for the contributions, @mjtorn. 🙂

mjtorn:
@fleskesvor thank you for merging :)

As for regressions and such, my prototype repo is so messed up wrt the main Escoria development, I will have to recreate the prototype game as a submodule when I have time at the office.

This may give me insight into how to fix the test scene, but I think it's best someone else look into the test scene and such, because that's a second opinion.

PR #55, opened 2018-02-15
Fix where the action menu gets opened

mjtorn:


device/globals/action_menu.gd
```diff
@@ -31,20 +27,13 @@ func _input(event):
...
-			action_pressed(a.get_name())
-			break
-
 func _ready():
 
-	actions = []
 	var acts = get_node("actions")
 	for i in range(acts.get_child_count()):
 		var c = acts.get_child(i)
 		if !c.is_type("Button"):
```
fleskesvor: I think this needs to be `if !c is Button:` in Godot 3.0. Does the editor not complain about this?

device/globals/action_menu.gd
```diff
@@ -1,15 +1,11 @@
 var target
```
fleskesvor: Don't you get complaints about the script not extending a node? Could add `extends Node2D` at the top.

mjtorn: Thank you! Yes! I had a commit for that in my master branch originally.

Problem is that this bunch of PRs and issues and remotes and branches are gnawing away at my brain cells bit by bit.

I cherry-picked it into this now.

device/globals/game.gd
```diff
@@ -115,10 +115,7 @@ func spawn_action_menu(obj):
 		return
 	action_menu.show()
 	var pos
-	if obj.has_node("action_menu_pos"):
```
fleskesvor: I think the purpose of this was to fix the action menu to a specific position on screen, using a `Position2D` node. I don't remember ever playing a game that used this feature though, so I don't know how useful it was.

fleskesvor:
Good job; thank you! :smile:

PR #56, opened 2018-02-15
Allow setting esc script on scene.gd node and run load commands

fleskesvor:
I haven't tested this properly with dialoge, background music and other things a developer might want to trigger on loading a scene, but setting flags, eg. to add items to inventory, works fine.

### DIFF
--- a/device/globals/main.gd
+++ b/device/globals/main.gd
@@ -73,11 +73,14 @@ func menu_collapse():
 		i -= 1
 		menu_stack[i].menu_collapsed()
 
-func set_current_scene(p_scene):
+func set_current_scene(p_scene, events_path=null):
 	#print_stack()
 	current = p_scene
 	var root = get_tree().get_root()
 	root.move_child(p_scene, 0)
+
+	if events_path:
+		vm.load_file(events_path)
 
 func wait_finished():
 	vm.finished(wait_level)
--- a/device/globals/scene_base.gd
+++ b/device/globals/scene_base.gd
@@ -1,6 +1,7 @@
 extends Node
 
-func _ready():
-	main.call_deferred("set_current_scene", self)
+export(String, FILE) var events_path = ""
 
+func _ready():
+	main.call_deferred("set_current_scene", self, events_path)
 


mjtorn:
As it stands, this does not work. I want the player to say something when the scene is loaded, to set up the plot of the game.

The player is not found when `main.esc` contains

```
:load
say player "I better go deliver my speech."
```

It doesn't find the player node, so there's definitely some deferral issue going on.

I think there's a deeper problem, which I tried to address in my approach before I gave up.
The call stack from `vm.load_file(...)` goes to `vm.clear()` which clears all the global state and then goes to `vm.run_event(...)`.

So if you added items to your inventory, it would logically work because it's effectively the same as loading a saved game, which afaict are serialized esc scripts for the game state that get "evaled" onto a cleared state.

I wanted a `:ready` event which would not clear the state but only execute the commands in the event. I could not find a simple code path for executing commands.

mjtorn:
```diff
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -490,10 +490,13 @@ func load_file(p_game):
                return

        var game = compile(p_game)
-       clear()
-       loading_game = true
-       run_event(game["load"])
-       main.menu_collapse()
+       if "load" in game:
+               clear()
+               loading_game = true
+               run_event(game["load"])
+               main.menu_collapse()
+       elif "ready" in game:
+               run_event(game["ready"])

 func load_slot(p_game):
        var cb = [self, "game_str_loaded"]
```

The above diff onto your HEAD is maybe a bit hackish, but it has the benefit of working. I was probably over-engineering this because I didn't want to (think of) overloading the `load_file(...)` function, though maybe it is for the best this way.

This also solves the deferral problem of not having the `player` node instantiated. I think it may have gone out with the `clear()` call or maybe something. Someone who knows about concurrency in Godot/Escoria can inform me why. Now the `elif:` clause has `player` there.

And I have to give love to esc scripts, this was a lot of messing about with UI elements in Unity/AC but I am confident this will work exactly as expected once I get more rooms implemented:

```
:ready
> [!said_deliver_speech]
	say player "I better go deliver my speech"
	set_global said_deliver_speech true
```

fleskesvor:
Thanks for sorting this stuff out. :slightly_smiling_face: For many things, this is a much easier way to introduce a room than having to create a cutscene.

PR #57, opened 2018-02-16
Allow setting esc script on scene.gd node and run load commands

mjtorn:
This replaces #56 as discussed on IRC/#escoria.

It introduces a `ready` event that can be used when attaching an esc script to a scene.


fleskesvor:
I think this is a good way to solve this. Any objections, @StraToN ?

fleskesvor:
It works as advertised, and retains the old behaviour of the `:load` event, so I see no reason not to merge this.

PR #58, opened 2018-02-16
Fix dialog speech

mjtorn:
This is heavily dependent on #54 and should be kept based on it until it is merged.

Sorry about the hassle, but lingering PRs are what they are :/


fleskesvor:
Works perfectly in my testing. 🙂Thank you, @mjtorn!

PR #61, opened 2018-02-19
Experimental project structure

fleskesvor:
This PR is a bit experimental, but what it does is move game specific scenes into `game/`, where `scene_cache.gd` and `inventory_items.tscn` used to reside, while those are moved to `globals/` and `ui/` respectively.

Additionally, the two scenes with resource loaders for adding custom scenes, `dd_player.tscn` and `dialog_player.tscn`, are moved into `ui/`, which contains the framework scenes that a developer must tweak to customize the appearance of their game, etc. This is still a bit awkward, since these are scenes that might need to be updated upstream, so it doesn't offer the clean separation of framework and game that would be ideal for a game developer wanting to keep the framework up to date.

Another dilemma is that the `game/` directory contains the demo/template that a developer might want to base their game off of, but it should ideally be thought of as a sub module with its own version control, managed by the game developer. We could rename this directory to `demo/` and instruct developers to copy the files or start from scratch in a different directory, eg. `game/`, but then they'd also need to update any project settings pointing to `demo/`, though that might not be an issue.

I think it's going to be difficult to avoid merge conflicts between Escoria and the game altogether, but it might be possible to reduce the problem by placing more of the customisation in project settings, and less in the scenes in `ui/`. An example of this would be to toggle the verb menu in project settings, instead of having to having to hide/remove it from the hud.


mjtorn:
I'm all for toggling the menu style in settings!

And also IMO the demo should be placed in `demo/` so that `game/` can be a submodule.

Maybe the first iteration won't bring total separation, but the more we can contain things that the user customizes in one place, even if it requires copying files over and tweaking paths, the better for PRs and probably making this an asset later on.

The big thing here is transitioning and documentation, because this most definitely breaks some kind of APIs or at least expectations.

There is the benefit, though, that if the demo is in `demo/`, no one has to do create a git submodule and still benefit from that `game/` directory, so the documentation can just say to do that copy, or use a submodule, or just mess with the code as-is, with the caveat that contributing to Escoria will be somewhat painful.

fleskesvor:
> And also IMO the demo should be placed in `demo/` so that `game/` can be a submodule.

Yes, the more I think about it, the more convinced I am that it's the best and most intuitive solution. We should also offer clear instructions on how to best set up a good project structure in the README.

> The big thing here is transitioning and documentation, because this most definitely breaks some kind of APIs or at least expectations.

I don't think it breaks any APIs, but it will make the Floss Manuals tutorial outdated on file locations, since it refers to scenes by their current resource paths. Then again, it's already a bit incorrect, since it refers to binary scene files, with the .scn file suffix, of which there are none left.

PR #62, opened 2018-02-20
Branch for removing dead code

mjtorn:
We may want short-lived branches for these, or what do you think?

I don't know of any other dead code right now, so IMO this can go in.


fleskesvor:
Doesn't this also make the whole `_input()` function redundant, as well as the call to `set_process_input()` in ´_ready()`? I mean, the `pressed` event is connected to the action buttons when the scene is ready, and it doesn't look like I would be necessary for the action_menu node to perform any input handling itself after that.

What I don't understand though, is why the current implementation doesn't trigger duplicate events when an action button is pressed. Or does it?

mjtorn:
So far I can confirm it does trigger both `_input()` and `action_pressed()`. I understood (misunderstood?) this so that godot bails out handling the input if `_input` returns. Not sure how to verify it, but I think the checks can be moved to `action_pressed()`.

mjtorn:
Now the name of the branch is a bit misleading, but at least the action menu code should be more acceptable.

fleskesvor:
It bails out of `_input()` but `action_pressed()` is a custom function triggered by connecting the `pressed`, so it won't have any effect on that. I don't think the first and third checks are needed in `action_pressed()` at least, because I don't see how that can be triggered unless the action is visible and pressed. The second is to prevent input while other events are taking place, I think, like a dialogue being played.

fleskesvor:
Did you check if `set_process_input(true)` in `_ready()` is needed, now that there's no `_input()` function?

mjtorn:
Removing `set_process_input(true)` had no effect that I could tell, force-pushing the removal now.

fleskesvor:
Thanks! :slightly_smiling_face:

PR #63, opened 2018-02-20
Have a setting so the right mouse button can open the action menu

mjtorn:
This makes for a more latter-day SCUMM experience.

### DIFF
--- a/device/globals/background.gd
+++ b/device/globals/background.gd
@@ -4,9 +4,9 @@ export var action = "walk"
 
 func input(event):
 	if event is InputEventMouseButton && event.pressed:
-		if (event.button_index == 1):
-			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, get_position() + Vector2(event.position.x, event.position.y))
-		elif (event.button_index == 2):
+		if (event.button_index == BUTTON_LEFT):
+			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, get_position() + event.position, event)
+		elif (event.button_index == BUTTON_RIGHT):
 			emit_right_click()
 
 func get_action():
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -67,7 +67,7 @@ func set_current_action(p_act):
 func set_current_tool(p_tool):
 	current_tool = p_tool
 
-func clicked(obj, pos):
+func clicked(obj, pos, input_event = null):
 	# If multiple areas are clicked at once, an item_background "wins"
 	if obj is Area2D:
 		for area in obj.get_overlapping_areas():
@@ -80,7 +80,7 @@ func clicked(obj, pos):
 		player = self
 	if mode == "default":
 		var action = obj.get_action()
-		# Hide the action menu when performing actions, so it's not eg. open while walking
+		# Hide the action menu (where available) when performing actions, so it's not eg. open while walking
 		if action_menu:
 			action_menu.stop()
 		if action == "walk":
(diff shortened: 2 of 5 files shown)


device/globals/global_vm.gd
```diff
@@ -101,7 +101,7 @@ func drag_end():
 		printt("********** dragging ends")
 		if hover_object != null && !hover_object.inventory:
 			printt("calling clicked")
-			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", hover_object, hover_object.get_position())
+			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", hover_object, hover_object.get_position(), null)  # null for the `input_event` argument
```
fleskesvor: Make `clicked()` default to `null` instead if the argument isn't present.

fleskesvor: For the `input_event` argument that is.

device/globals/background.gd
```diff
@@ -5,7 +5,7 @@ export var action = "walk"
 func input(event):
 	if event is InputEventMouseButton && event.pressed:
 		if (event.button_index == 1):
-			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, get_position() + Vector2(event.position.x, event.position.y))
+			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, get_position() + Vector2(event.position.x, event.position.y), event)
```
fleskesvor: `Vector2(event.position.x, event.position.y)` is better written as `event.position`

fleskesvor: You should also pass the event object from the equivalent function call in background_area.gd:

https://github.com/godotengine/escoria/blob/master/device/globals/background_area.gd#L10

device/globals/game.gd
```diff
@@ -108,7 +108,19 @@ func clicked(obj, pos):
 			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", "")
 
 		elif obj.use_action_menu && action_menu != null:
-			spawn_action_menu(obj)
+			if ProjectSettings.get_setting("escoria/ui/right_mouse_button_action_menu"):
+				if input_event.button_index == 2:
```
fleskesvor: Better to use `BUTTON_RIGHT` than `2` for clarity.

device/globals/background.gd
mjtorn: In my defense I offer the court that the original `Vector2` code wasn't mine, though I plead guilty of not having fixed it ;)

Of course it should be clearer.

device/globals/game.gd
mjtorn: There were other instances of referring to the buttons by numbers; I'll take a moment to see if I can hunt them down.

mjtorn:
I believe I addressed the issues now

device/globals/background.gd
fleskesvor: I believe you. :slightly_smiling_face: This script has remained unchanged since Escoria was open sourced in August 2016, so my guess is, at some point the x and y coordinates were separate properties of the event object, making it necessary to wrap them in a `Vector2D` type.

device/globals/game.gd
fleskesvor: Thanks, that would be awesome. :smile: Not strictly related to this branch of course, but I appreciate the extra effort.

fleskesvor:
Remember to check that this doesn't cause any regressions with no action menu and the verb menu enabled instead.

fleskesvor:
This PR has a conflict in a comment now.

mjtorn:
Last time I use GitHub to fix a conflict, the tree became a mess :/ the problem is fixed and hopefull Git will figure out the history so it doesn't suck too hard.

Meaning I hope I don't need to put effort into cleaning up GH's mess and that this can be merged ;)

fleskesvor:
Works well in my testing, and some of this code is needed for a two-click-interface as well. Thanks for the contribution, @mjtorn. 🙂

PR #64, opened 2018-02-20
[RFC] Tooltip follows mouse

mjtorn:
This is work-in-progress for having the tooltip follow the mouse.

Currently hiding the tooltip when opening the action menu is broken. Currently I have no idea how to fix it, please help, unless I figure something out soon!

### DIFF
--- a/device/globals/action_menu.gd
+++ b/device/globals/action_menu.gd
@@ -16,6 +16,10 @@ func start(p_target):
 		target = p_target
 		target.connect("visibility_changed", self, "target_visibility_changed")
 
+		# Do not display the tooltip alongside the menu
+		if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
+			get_tree().call_group("hud", "hide")
+
 	var scale = ProjectSettings.get_setting("escoria/platform/action_menu_scale")
 	set_scale(Vector2(scale, scale))
 
@@ -24,6 +28,8 @@ func stop():
 		target.disconnect("visibility_changed", self, "target_visibility_changed")
 	target = null
 	hide()
+	if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
+		get_tree().call_group("hud", "show")
 
 func _input(event):
 	if !is_visible():
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -21,6 +21,8 @@ var click_anim
 var camera
 export var camera_limits = Rect2()
 
+var tooltip
+
 func set_mode(p_mode):
 	mode = p_mode
 
@@ -233,7 +235,7 @@ func _process(time):
 					mobj = objs[key]
 		if typeof(mobj) != typeof(null):
 			if mdist < min_interact_dist:
-				if true || mobj != last_obj:
+				if mobj != last_obj:
 					spawn_action_menu(mobj)
 					mouse_enter(mobj)
 					last_obj = mobj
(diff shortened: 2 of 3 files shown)


device/globals/action_menu.gd
```diff
@@ -16,6 +16,13 @@ func start(p_target):
 		target = p_target
 		target.connect("visibility_changed", self, "target_visibility_changed")
 
+		# Do not display the tooltip alongside the menu
+		if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
+			# FIXME: Hiding does not work :(
+			var hud = get_tree().get_root().get_node("/root/hud")
```
fleskesvor: `get_tree().get_root()` is equivalent to `get_node("/root")`

Isn't the path to the hud in the scene tree `get_node("/root/game/hud")` though?

mjtorn: I tried `print_tree()` and got only the current tree, and I couldn't find it by trying, but maybe I didn't try that one.

What do you think of the addition to `ProjectSettings` for the hud? Is that valuable in restructuring the project?

Because if I can avoid exporting/exposing it, I'm more than happy to get rid of it. It doesn't feel right.

fleskesvor: A better solution would be to use `get_tree().call_group("hud", ...)` though:

http://docs.godotengine.org/en/3.0/classes/class_scenetree.html#class-scenetree-call-group

mjtorn: I'll look into that after I get the mouse button stuff cleared, thanks!

It would be nice if this magically fixed the problem as well ;)

fleskesvor: I'm not sure it makes sense to make a singleton of hud, or at least it's not a quick fix, because there's still the hud that is instanced into the current scene and the singleton isn't a part of that scene, so I think there's a chance of things blowing up in unexpected ways.

mjtorn: Either way, your `call_group()` solution was a good solution!

Squashed the two commits and force-pushing now.

device/globals/game.gd
```diff
@@ -251,6 +251,18 @@ func _process(time):
...
+	if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
+		# Must verify `position` is there, key inputs do not have it
+		if vm.hover_object and "position" in ev:
+			var tooltip = get_node("hud_layer/hud/tooltip")
+			var pos = ev.position
+			var pos_x = ev.position.x - (tooltip.get_size().x / 2)
+			var pos_y = ev.position.y - (tooltip.get_size().y)
+			pos.x = pos_x
+			pos.y = pos_y
+			tooltip.set_position(pos)
```
fleskesvor: My vector math is a bit rusty, but I think this can be written as `var pos = ev.position - tooltip.get_size() / Vector2(2, 1)`

mjtorn: Something new every day! That works and looks like the preferred way. Force-push incoming soon.

fleskesvor:
> Currently hiding the tooltip when opening the action menu is broken. Currently I have no idea how to fix it, please help, unless I figure something out soon!

This works now, right?

It might be less "expensive" to expose a function in the hud to adjust the tooltip position via a group call, but this is looks good to me, and I'm happy to merge it when you're done with it.

fleskesvor:
Oops, I accidentally hit the wrong button. :slightly_frowning_face:

mjtorn:
The tooltip is now hidden by the menu, yes.

I also figured out how to have the tooltip node looked up only once, this is nice.

As for the idea discussed on IRC about maybe call_group being more efficient, I couldn't get it to work and if you were willing to accept this as it was, please accept this performance increase of setting the node only once :)

fleskesvor:
Oh, there's a `load_hud()` function. This looks like the proper way to do this. Thanks for fixing this. :slightly_smiling_face:

Issue #45, opened 2018-02-05
Documentation idea(s)


mjtorn:
I'm seriously considering Godot and Escoria for a point 'n click adventure game, because Unity assets are sometimes quite cumbersome and I have a hunch fighting with black-box assets could be replaced with doing things myself. Hoping that this is a valid assessment, I need to investigate Escoria properly.

It would be valuable to have a section of the Flossmanuals documentation, very early on, about best practices in the file system, related to installation. Like that `device` is the root of the project.

Parts of the documentation feel like everything's laid out really well and other parts feel like the reader should really figure things out. I got something going with the documentation but the consistency feels a bit off...

For example

https://fr.flossmanuals.net/creating-point-and-click-games-with-escoria/how-to-start-your-escoria-project/

says "You want to copy the Escoria folder and rename that copy to whatever suits you, like the codename of your project. Once it’s done, open Godot by clicking the file you downloaded earlier, this will open the Godot project manager window."

Apparently the "Escoria folder" is the Git checkout or Zip file, and it should be copied in its entirety. Referring to a file downloaded earlier sounds like maybe it would be the Zip file or something, when "launch Godot" would actually be clearer, though it's fewer words.

One "best practice" to point out would be deleting the `.git` path if it's there. It's kind of obvious, but because you're trying to be very explicit (a good thing!) maybe pointing this out is worth it. This will allow developers to init this as their own Git repository.

A developer may wonder if it's possible to start with an empty project and just copy/create certain files, and if so which ones?

Anyway, using the Git repository brings its own stuff with it, and it looks like the correct assets, but not everything is there. What if they were removed from the repository and the documentation updated to say "Download the entire flossmanuals fork as a Zip file or `git clone https://github.com/flossmanualsfr/escoria.git flossmanuals-escoria`" or something?

https://fr.flossmanuals.net/creating-point-and-click-games-with-escoria/creat-your-scene/

There it might be valuable to say the template scene should be copied under `device/` for clarity.

It might also be clearer if the `scenes` directory were called `resources` or `assets` because that's what it's apparently about.

Issue #5 would of course be helpful in this as well :)

fleskesvor:
Yes, I 100% agree that there needs to be better documentation and examples on best practices and a better project structure, because I think a developer is much more likely to have a good experience with Escoria if they don't unintentionally make a huge mess of their project from the get-go.

[...]

PR #67, opened 2018-02-21
Fixing the inventory

mjtorn:
This is not complete yet, but opened to gather preliminary feedback. There is even a non- yet near-zero possibility I'll be committing to it today.

The inventory is very broken as it stands in master, and something like the in-game buttons are not documented in the book (#45 flossmanualsfr#15).

I can only guess at what the intentions originally were here and try to fix accordingly. For what it's worth, with these changes in, I can hit the tab key to open an inventory and complaints are benign instead of crashing everything.

I have not created the in-game buttons and would hope that the branch goes in as is now, or with minimal changes, and someone else fix the buttons without me. This is because I do not currently see a need to have the buttons in anything I'm working on and it's a huge distraction. If I change my mind, this would still be on IRC/#escoria or a new PR or wherever separate from this.

@fleskesvor I am happy to squash these if they pass inspection ;)

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -29,20 +29,37 @@ func set_mode(p_mode):
 func mouse_enter(obj):
 	var text
 	var tt = obj.get_tooltip()
-	if current_action != "" && current_tool != null:
-		text = tr(current_action + ".combine_id")
-		text = text.replace("%2", tr(tt))
-		text = text.replace("%1", tr(current_tool.get_tooltip()))
-	elif obj.inventory:
-		var action = inventory.get_action()
-		if action == "":
-			action = current_action
-		text = tr(action + ".id")
-		text = text.replace("%1", tr(tt))
+	# We must hide all non-inventory tooltips and interactions when the inventory is open
+	if action_menu and inventory.is_visible():
+		if obj.inventory:
+			if !current_action:
+				text = tr(tt)
+			else:
+				var action = inventory.get_action()
+				if action == "":
+					action = current_action
+				text = tr(action + ".id")
+				text = text.replace("%1", tr(tt))
+			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", text)
+			vm.hover_begin(obj)
 	else:
-		text = tt
-	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", text)
-	vm.hover_begin(obj)
+		if current_action != "" && current_tool != null:
+			text = tr(current_action + ".combine_id")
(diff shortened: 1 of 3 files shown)


device/globals/game.gd
```diff
@@ -91,6 +97,21 @@ func clicked(obj, pos):
 			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", "")
 
 		elif obj.inventory:
+			# Use and look are the only valid choices with an action menu
+			if action_menu:
+				# Do not set `look` as permanent action
+				if input_event.button_index == BUTTON_RIGHT:
```
fleskesvor: I don't think `input_event` is declared anywhere.

device/globals/game.gd
```diff
@@ -55,6 +58,9 @@ func mouse_exit(obj):
 
 func clear_action():
 	current_tool = null
+	# XXX: Should current_action be cleared also for verb menus? Until now it has not been.
```
fleskesvor: I don't think it should be. With an action menu, that's OK, since the actions are always next to your cursor, but with a verb menu, you'd have to go all the way back to the verb menu to pick the verb again every time it's cleared. Maniac Mansion and Zak McKracken did that, but I always found that more than a little annoying when playing those games. Later LucasArts games did not clear the action.

device/globals/inventory.gd
```diff
@@ -22,19 +20,26 @@ func open():
...
+
+	var animation = get_node("animation")
+	if animation:
+		if animation.is_playing():
+			return
+		animation.play("hide")
+
+	# XXX: What is this `look` node? A verb menu thing?
+	if has_node("look"):
+		get_node("look").set_pressed(false)
```
fleskesvor: I'm guessing that if the inventory actions thing was working, it would have actions defined by adding buttons with their named actions under the actions node, similar to the action menu, and look would be the pre-selected action.

device/globals/game.gd
mjtorn: It is `null` by default, but I was not able to cause a problem here. It's as if whenever that code path is traveled, the `input_event` has been passed.

mjtorn:
The latest commit turned into a bit of a hackfest, but AFAICT for now, everything appears to work as expected! Finally!

mjtorn:
@fleskesvor please merge if it looks at all good to you!

fleskesvor:
I'm unable to get the inventory to appear by pressing Tab in your test branch, but everything works as expected in the demo scene, so this looks good to me. Thanks for the contribution! 🙂

mjtorn:
I'll leave a note for posterity here, to mirror what I said on IRC/#escoria, my test branch was probably un-pushed at the time.

PR #69, opened 2018-02-23
Documentation on how to use a git submodule

mjtorn:
I documented how I got https://github.com/mjtorn/escoria-proto-submodule embedded this way.

Note that it is embedded in the https://github.com/mjtorn/escoria/tree/devel branch.

@fleskesvor is this in line with your thoughts as well?

### DIFF
--- a/doc/git_submodule.md
+++ b/doc/git_submodule.md
@@ -1,0 +1,68 @@
+# Using git submodules to develop a game
+
+Caveat: This is not necessarily the one true way. The framework and its practices
+are constantly being developed, but this document contains some pointers.
+
+Submodules are a bit advanced and specific to your project, and therefore the commands
+you may use are as well. You can get into Git [here](https://git-scm.com/book/en/v1/Getting-Started).
+
+This is the [chapter on submodules](https://git-scm.com/book/en/v1/Git-Tools-Submodules).
+
+## Why?
+
+Git's strength is in how well it handles remotes and branches. This may go beyond
+Git 101, but it's a very efficient way to work once you get into it.
+
+Having the framework separate from your game also helps to identify changes to the
+code you may have to make. The goal of a framework is to be completely separate from
+whatever it's used for, so please open issues on GitHub if you find yourself making
+changes to the framework that aren't bugfixes or new features.
+
+If possible, please make a PR that alters Escoria's behavior, eg. through `ProjectSettings`,
+so that the change doesn't happen in Escoria's upstream code.
+
+You may read more about this in the ../CONTRIBUTING.md document.
+
+## How?
+
+Create a git repository and add the Escoria GitHub URL as a remote, eg. `gh-escoria`.
+This document will refer to your repository as `origin`.
+
+Reset your `master` to match whatever `gh-escoria/master` is at the time.
+
+Escoria is not (at the time of writing this) at the stage where you never have to touch
+the framework code. Whether you contribute or not, or whatever workflow you want to use,
+this document will refer to your work branch as `devel`. This is checked off of `master`.
+
+The project root is `device/`. It contains your project settings. When submitting branches
+for merging, please do not include your game's settings. Any new features should be disabled
+by default, to not mess with the tutorial [*Creating Point and Click Games with Escoria*](https://fr.flossmanuals.net/creating-point-and-click-games-with-escoria/).
+
+## Actual work
+
+Create a new repository for the submodule.
+
+Add it as `device/game/`.
+
+You can also create subdirectories like `rooms`, `actors` and `items`. The best practices are
+not clearly defined yet, but such a naming scheme correlates with tradition. As your game comes
+along, you will logically create a subdirectory for every item or actor and so on.
+
+Now you're ready to launch Godot!
+
+## Misc notes
+
+Whatever is parallel to `game/` can be copied under `game/` and the project structure changed
+accordingly. This is your last resort if you can't expose a setting.
+
+As another use case, if you want to use the `demo/ui/action_menu.tscn` for a template, you could
+copy it under `game/ui/` - your submodule - and maintain it there.
+
+In case you want to edit the "heads-up display", `ui/hud.tscn`, you will have to do it in-tree
+at least for the time being. Take care not to include it in any PRs upstream.
+
+If you want to contribute to Escoria upstream, please do not submit a branch that is based on
+`devel`. It may feel like branch acrobatics to have one taken off `master` and rebased onto
+`devel` for live work, maybe even cherry-pick-squashing changes from the `devel` version onto
+the `master` version, but it is necessary to keep the Escoria code base as clean as possible.
+


doc/git_submodule.md
```diff
@@ -0,0 +1,70 @@
...
+
+The project root is `device/`. It contains your project settings. When submitting branches
+for merging, please do not include your game's settings. Any new features should be disabled
+by default, to not mess with the tutorial [*Creating Point and Click Games with Escoria*](https://fr.flossmanuals.net/creating-point-and-click-games-with-escoria/).
+
+## Actual work
+
+Create a new repository for the submodule.
+
+Add it as `device/game/`.
```
fleskesvor: Add your new directory to a `.gitignore` file at the root of your directory and initialize a new git repository under `device/game` with the command `git init`.

doc/git_submodule.md
```diff
@@ -0,0 +1,70 @@
...
+
+## How?
+
+Create a git repository and add the Escoria GitHub URL as a remote, eg. `gh-escoria`.
+This document will refer to your repository as `origin`.
+
+Reset your `master` to match whatever `gh-escoria/master` is at the time.
+
+Generally messing with `master` is discouraged. You may adapt whatever workflow you want
+to this, but this document will use `devel`.
```
fleskesvor: I'm thinking that maybe we should just keep it simple here and assume the reader isn't interested in contributing to upstream, but rather put information about that in a contribution guide later on.

doc/git_submodule.md
```diff
@@ -0,0 +1,70 @@
+# Using git submodules to develop a game
+
+Caveat: This is not necessarily the one true way. The framework and its practices
+are constantly being developed, but this document contains some pointers.
+
+If someone wants to add specific git commands here, please do. The current author
+will not bother, because formatting and anonymizing the repository URLs is too
+much work. This is the downside of open source, you do not always get everything
+for free.
```
fleskesvor: This paragraph could be replaced with a few words about why it's a good idea to use version control (eg. backup, pinpoint the introduction of a bug, easily switch between different states of a project, sync changes between multiple developers, etc.) and just point to the official getting started document: https://git-scm.com/book/en/v1/Getting-Started

fleskesvor:
Thanks, @mjtorn. 🙂Yeah, it's in line with my thoughts. I think some of the information should be split out into separate [contribution guidelines](https://help.github.com/articles/setting-guidelines-for-repository-contributors/) though, in order to keep this one brief and more focused on the benefits of using a sub module just for ease of updating Escoria while developing a game, and less on contributing upstream.

doc/git_submodule.md
mjtorn: Is this how you would write it? I don't really know why it'd be in `.gitignore` (whose most useful contents seem to be `*.import`), but I never tried it.

It's valuable for the branches to be able to track the submodule. Like in the trying-to-sort-out-inventory work. I have a devel branch in Escoria for the work and an "inventory" branch in my submodule, and the devel branch has committed the submodule to the correct ("inventory") branch and everything works as expected.

Maybe because I never tried ignoring a submodule, I'm missing out on something, so would you please elaborate?

doc/git_submodule.md
mjtorn: Maybe you're right. I have been approaching this very tightly from a contributor POV all the time, and while I stand behind my words here, I'm also ok with splitting the contribution stuff. This could say something more vague, "Reset your master and adopt whatever workflow". I'll look into it soon.

doc/git_submodule.md
mjtorn: I apologize for not being in the best of moods when writing this. Some kind of fatigue seems to be hitting me, repeatedly. Anyway, yeah, a pointer to some kind of guide could be there.

doc/git_submodule.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Why?
+
+Git's strength is in how well it handles remotes and branches. This may go beyond
+Git 101, but it's a very efficient way to work once you get into it.
```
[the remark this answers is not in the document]
mjtorn: Here I disagree with you. I'm quite certain people who are interested in submodules know things beyond Git 101. Doubly so if there's a link to a Git guide before this.

Having said that, if I split the contributor stuff out, the remotes and branches are more valuable in that context. I will contemplate on this.

doc/git_submodule.md
fleskesvor: The reason for the confusion is that I didn't know sub modules was an actual feature of Git until just now, so I've always just used ad-hoc sub modules, where I work with the two repositories independently. I think this is likely to be too difficult for your average game developer to follow, so I'm leaning towards just providing advice on separating game from framework, since that would be useful, even if they are just using zip-archives for "version control", as some do.

doc/git_submodule.md
fleskesvor: Ok, I see where you're coming from with this now. I think this entire document would be a great start on a contributor guideline, which we don't have at all at the moment, so maybe it would be a good idea rename this file to `CONTRIBUTING.md` and place it at the project root?

doc/git_submodule.md
mjtorn: Ah! That does explain it! Ok, we're addressing different audiences. I'm addressing the audience I represent myself, the one who's familiar and somewhat comfortable with this submodule-style stuff, but doesn't know how to apply it to Escoria.

IMO the non-submodule approach is by its very nature - literally it could be anything but a submodule - something that would be outside the scope of this document. The average game developer can read that other document instead.

doc/git_submodule.md
mjtorn: From the comment above, I have come to believe three (3) documents are required:

  * CONTRIBUTING.md for general contributor guidelines for developing Escoria
  * Best practices for how `game/` works
  * Using Git submodules for the best of both above worlds

The third one is essential for me as long as Escoria is somewhat broken and laden with surprises. I know I can have it set up and keep my notes somewhere in a not-document format, and maybe no one else cares, but submodules are a very effective and efficient tool, invented for good reason, so I think sharing is caring here.

doc/git_submodule.md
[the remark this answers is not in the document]
fleskesvor: Yes, we were looking at this from different perspectives, so I agree that in the context of a potential Escoria contributor, you are right.

doc/git_submodule.md
fleskesvor: No worries. :slightly_smiling_face: If we do split this up into two or more documents, it is probably not needed anyway, since we can assume that the reader has the skills required to understand what they're reading, or else knows where to look for help.

doc/git_submodule.md
fleskesvor: This sounds like a great idea! I can write the first two, and merge this PR as the third. But if you want to write either of the first two, I appreciate any contributions, of course. :wink:

mjtorn: Please write them! As I have no ownership in the project, and barely time for code (which I don't always like doing alone in my office instead of the game etc), I think it's best someone with a bigger stake takes care of such documents.

I will then massage this PR to fit the first two documents once you're done.

Thank you!

PR #70, opened 2018-02-23
Get rid of unused and broken action code in inventory

### LINKED ISSUE #29: Need of "actions" node in inventory, missing part in Escoria's documentation
**Hello once again!**

   Today, I was going through inventory system, I followed everything as it is written [here in Inventory chapter](https://fr.flossmanuals.net/creating-point-and-click-games-with-escoria/inventory/), but it seems, that it is missing a word about node of Control type called "actions".

   Before adding this node into inventory scene, I was experiencing unsuccesful attempt to call function get_child_count in inventory.gd, then I checked basic inventory setup, which comes with _Escoria_, and there it was. After adding, no problem. Please, for another new users, update documentation. There is written, you have to include those arrows for going through items, which I don't need, but it is there, so I added them and it works. This should be included too.

   By the way, what is that "actions" node used for? Is it used while using items from your inventory? Thank you for your response, sir!

mjtorn:
Should close #29

### DIFF
--- a/device/demo/ui/inventory.tscn
+++ b/device/demo/ui/inventory.tscn
@@ -3,7 +3,7 @@
 [ext_resource path="res://globals/inventory.gd" type="Script" id=1]
 [ext_resource path="res://demo/ui/inventory_items.tscn" type="PackedScene" id=2]
 
-[node name="inventory" type="Control"]
+[node name="inventory" type="Control" index="0"]
 
 anchor_left = 0.0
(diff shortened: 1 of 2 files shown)


fleskesvor:
As speculated on IRC, this might have been intended on inventory action buttons like in Gemini Rue, but currently the functionality is broken and hasn't been requested, so let's remove it for now. Thanks, @mjtorn.

PR #71, opened 2018-02-23
Show in-game buttons is broken, remove it so Escoria does not crash

mjtorn:
IMO this code should all be removed and replaced with something under `doc/` on how to create in-game buttons if someone wants it.

@StraToN @fleskesvor what do you think? Should this just be replaced with documentation? Would you write it?

### DIFF
--- a/device/globals/hud.gd
+++ b/device/globals/hud.gd
@@ -43,25 +43,4 @@ func menu_closed():
 func _ready():
 	add_to_group("hud")
 	add_to_group("game")
-	#get_node("inv_toggle").connect("pressed", self, "inv_toggle")
-	#get_node("inv_toggle").set_focus_mode(Control.FOCUS_NONE)
-
-	#get_node("buttons").hide()
-	if ProjectSettings.get_setting("escoria/platform/show_ingame_buttons"):
-		if (not get_node("inv_toggle").is_hidden()):
-			get_node("buttons").show()
-
-		var p = get_parent().get_parent().get_parent()
-		for i in range(0, p.get_child_count()):
-			var c = p.get_child(i)
-			if (c is preload("res://globals/background.gd")):
-				background = c
-				break
-
-		get_node("inv_toggle").connect("visibility_changed",self,"_on_inv_toggle_vis_chaged")
-		get_node("buttons/hints").connect("pressed",self,"_on_hint_pressed")
-		get_node("buttons/menu").connect("pressed",self,"_on_menu_pressed")
-
-
-	set_tooltip("")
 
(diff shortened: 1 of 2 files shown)


fleskesvor:
I'm not sure. It doesn't feel so Escoria-esque to lump all this together in the hud node. We should support the keyboard shortcuts that are expected of an adventure game framework, toggle inventory, show hotspots, skip dialogue, etc., but if it's broken now, I think we should just gut it and make an enhancement issue for future implementation.

We should add the shortcuts we want to support to the input map and this might be the right place to connect the signals, but implementation should be coded in the corresponding nodes, like hotspot highlight in item scenes, show menu in the menu scene, etc.

Do you need this feature now for the game you're making?

mjtorn:
I do not need this. If I change my mind, I'd implement it as an actual game feature, without involving `ProjectSettings` at all. This is why I suggested gutting it and documenting instead how to make buttons.

> We should support the keyboard shortcuts that are expected of an adventure game framework, toggle inventory, show hotspots, skip dialogue, etc., but if it's broken now, I think we should just gut it and make an enhancement issue for future implementation.

My inventory branch (when I have time for it later or tomorrow or soon) has the keyboard shortcut to toggle the inventory, but I don't really know how to set these up in the editor or code. They don't appear to be in the project settings. This should be addressed, but this issue #71 is about the ability to enable undocumented game-GUI nodes that aren't necessary for most games.

So do we concur that I will replace this PR or its contents with a removal?

Thanks for replying!

fleskesvor:
Yes, I think this feature should be supported by using custom input events, since that gives the developer the ability to map these events to one or more keyboard or controller buttons in the input map, as they see fit. I agree that ProjectSettings should not be needed. If they have added the key mappings, it can be assumed that they intend to use them.

If you don't use this feature, I don't know of anyone who does, so a removal seems like the best solution to me. I can make a proof-of-concept for how I see this working with the inventory button, if you want. I also find it odd that the inventory doesn't have an AnimationPlayer for showing/hiding the inventory. Was that something you implemented yourself in your branch?

mjtorn:
I can make a removal branch because I think I can have it done before you're off work, but if you have any proofs-of-concept, I'm more than interested in seeing them!

Also no idea about the animation stuff. I did not implement anything related to animations, as I'm not convinced I'll be wanting animations for my inventory, so we must assume this code is also broken.

Let's discuss the animation stuff further once I have a replacement PR for #67 .

Thanks!

fleskesvor:
No, I don't have a POC readily available, but I think it's OK to just merge your removal first and do a re-implementation in a later commit.

Yeah, animations might not always be needed, but if we have them, it will be easier to make the kinds of inventories that eg. scroll up/down from the bottom when moving the mouse close to a screen edge. That isn't supported yet, but it could be implemented with a developer defined Area2D.

mjtorn:
This day has turned out worse than anticipated scheduling-wise. I will deliver the removal ASAP. Cheers!

PR #72, opened 2018-02-25
Fix automatic background area code

fleskesvor:
This updates the automatic generation of a shape for the background when using a Sprite node to work with Godot 3, and also uses a Sprite instead of an Area2D to make it more similar to using a TextureRect node for the background. To make the process more straight-forward, the Area2D node is also generated from code, so these are the only steps required to use a background_area

1. create background Sprite and attach image
2. disable centering of Sprite node
3. attach background_area.gd script

Step 2 should probably be done automatically, but it will not be visible to the developer unless the script is made into a tool script, so maybe it's best to just document this.

### DIFF
--- a/device/globals/background_area.gd
+++ b/device/globals/background_area.gd
@@ -1,11 +1,10 @@
-extends Area2D
+extends Sprite
 
 export var action = "walk"
-var collisionshape_click
-var rectangleshape
+var area
 
 func input(viewport, event, shape_idx):
-	if event.type == InputEvent.MOUSE_BUTTON && event.pressed:
+	if event is InputEventMouseButton and event.pressed:
 		if (event.button_index == BUTTON_LEFT):
 			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, get_position() + event.position, event)
 		elif (event.button_index == BUTTON_RIGHT):
@@ -19,19 +18,26 @@ func _init():
 
 func _enter_tree():
 	# Use size of background texture to calculate collision shape
-	var background = get_parent().get_node("background")
-	if background:
-		var size = background.get_size()
-		var extents = Vector2(size.x / 2, size.y / 2)
-		var transform = Transform2D(Vector2(1, 0), Vector2(0, 1), extents)
+	var size = get_texture().get_size()
 
-		var shape = RectangleShape2D.new()
-		shape.set_extents(extents)
-		add_shape(shape)
-		set_shape_transform(0, transform)
+	area = Area2D.new()
+	var shape = RectangleShape2D.new()
+
+	var sid = area.create_shape_owner(area)
+
+	# Move origin of Area2D to center of Sprite
+	var transform = area.shape_owner_get_transform(sid)
+	transform.origin = size / 2
+	area.shape_owner_set_transform(sid, transform)
+
+	# Set extents of RectangleShape2D to cover entire Sprite
+	shape.set_extents(size / 2)
+	area.shape_owner_add_shape(sid, shape)
+
+	add_child(area)
 
 func _ready():
-	connect("gui_input", self, "input")
+	area.connect("input_event", self, "input")
 	add_to_group("background")
 
 func emit_right_click():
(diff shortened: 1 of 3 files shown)


fleskesvor:
With this change, the two ways to add a background are so similar that I think they should use the same script, so I will add that functionality before merging this PR.

mjtorn:
You mean `background_area.gd` would replace `background.gd` altogether? I couldn't find any documentation or reference to this, so it sounds like a good idea. At least definitely maybe ;)

fleskesvor:
I meant to implement combine them to have `background.gd` work with both `Control` and `Sprite` nodes, similar to how `item.gd` works, so that it's backwards compatible. Unlike items though, there may not be good cases for using the old method, since the new one is more flexible, and as easy to use with the automatic shape logic, so maybe we should just replace it.

I don't think anything breaks by using `Sprite` backgrounds with `Control` items, but I will do some proper testing to verify.

StraToN:
Good idea. Users will probably tend to use a `Sprite` in first place since their backgrounds usually are a simple texture, with some animated sprites over it. Sounds clearer like this.
Is it really required to make it work the same on a `Control` node (ie. `TextureRect` I suppose) since the `Area2D` will already get the inputs?

mjtorn:
On Mon, Mar 05, 2018 at 03:42:31AM -0800, Julian Murgia wrote:
>Good idea. Users will probably tend to use a `Sprite` in first place since their backgrounds usually are a simple texture, with some animated sprites over it. Sounds clearer like this.
>Is it really required to make it work the same on a `Control` node (ie. `TextureRect` I suppose) since the `Area2D` will already get the inputs?

The thought also occurred to me, that I tried setting up a parallax
background, and it failed.

I made `set_camera_limits` call a recursive `find_background` function so as
to set the limits. This involved a bit more work and I'll probably have to
take it up elsewhere, like why it had to be broken out at all.

Anyway, whatever happens with the background-handling code, it would be nice
if parallaxes would work as well.

fleskesvor:
@StraToN: Do you mean that we should deprecate the old method and just use this one? Maybe that's a good idea, since this method should be much easier to adapt. I haven't used parallax backgrounds before, but I will try to set up a test scene tomorrow and see which changes need to be made to have it work out of the box with that as well.

One problem with this script as it is, is that it makes the assumption that the sprite is uncentered (which is not the default of the Sprite node) and adjust the click shape accordingly. This could be fixed by either overriding the `_get()` method to always return false for the `centered` property or to monitor the property, and position as well, come to think of it, and adjust accordingly. Or, I guess, just document the correct configurations.

I also noticed a regression with interaction with area items, where the player takes turns standing at the interact_pos and the event position when interacting with items. This was an issue that was fixed in 2.1, but I guess recent changes in Godot or Escoria have introduced a regression. I will look into the problem.

mjtorn:
@fleskesvor your experience is much valued with parallaxes! They are not a _super_ priority for me, but having a sense of depth makes a game feel much more modern.

I looked at the 2d platformer demo to maybe get some ideas. There were problems with camera positioning and how the sprites are scaled. This probably has to be documented separately, as it's not present in the test scene or the Daiza game. Also because it most likely affects things when playing on a tablet, if sprite sizes and such are a huge deal.

It would be great if I could keep that part on a personal back-burner and focus on other things, but will of course happily test and work on something, if you get a demo or something started :)

StraToN:
@fleskesvor 
> Do you mean that we should deprecate the old method and just use this one? Maybe that's a good idea, since this method should be much easier to adapt.

I think so.
It now appears with our experience that the background is not always just a still image. There is no real gain using a `TextureRect`. We want both advantages of `Sprite `and `Area2D `in the same node (which is why I do believe that a custom node would even be easier for newcomers). I think most people will naturally choose a `Sprite `for backgrounds anyway, which enables him to apply transformations on it.

Concerning the centering, I also think people naturally un-center it because they will easily see it doesn't fit the screen, so it is no big deal.

fleskesvor:
I've put off the decision of removing the old method for now, since it might still be useful until we have good examples of the benefits of the new method. I instead added some documentation to this PR, which should hopefully be enough to teach developers how to use Sprites for backgrounds instead. I also just fixed the issue with every other click being captured by the background, which became a problem again as a result of my other changes. This all works now, so I'm going to merge it.

PR #73, opened 2018-02-28
Setting to force disabling typewriter speech

mjtorn:
Because the actual variable is sort of convolutedly hidden, this makes configuration easier.


fleskesvor:
Eventually, I think we might want to expose dialogue scenes more to the developer by instancing them as a child of a player/item, but for now, this seems like a good workaround. :+1:

PR #75, opened 2018-03-01
Variables to affect player speed

mjtorn:
Because sometimes a scaled-down player sprite can take annoyingly long to cross the screen. By default the speed is not affected.

Also make double-clicking move the player faster, as is customary in modern adventure games. The default is a multiplier of 1.5.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -91,6 +91,10 @@ func set_current_tool(p_tool):
 	current_tool = p_tool
 
 func clicked(obj, pos, input_event = null):
+	var walk_context = null
+
+	if input_event:
+		walk_context = {"fast": input_event.doubleclick}
 	# If multiple areas are clicked at once, an item_background "wins"
 	if obj is Area2D:
 		for area in obj.get_overlapping_areas():
@@ -112,7 +116,7 @@ func clicked(obj, pos, input_event = null):
 			if player == self:
 				return
 			if (action_menu and !inventory.is_visible()) or !action_menu:
-				player.walk_to(pos)
+				player.walk_to(pos, walk_context)
 				# Leave the tooltip if the player is in eg. a "use key with" state
 				if !current_action:
 					get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", "")
@@ -150,7 +154,7 @@ func clicked(obj, pos, input_event = null):
 			if player == self:
 				return
 			if (action_menu and !inventory.is_visible()) or !action_menu:
-				player.walk_to(pos)
+				player.walk_to(pos, walk_context)
 				# Leave the tooltip if the player is in eg. a "use key with" state
 				if !current_action:
 					get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", "")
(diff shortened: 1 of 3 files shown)


device/globals/game.gd
```diff
@@ -15,6 +15,8 @@ var last_obj = null
 var current_action = ""
 var current_tool = null
 
+var walk_context = null
```
fleskesvor: Isn't this variable only used in `clicked()`?

device/globals/player.gd
```diff
@@ -244,7 +244,9 @@ func _process(time):
 		else:
 			next = walk_path[path_ofs]
 
-		var dist = speed * time * last_scale.x * last_scale.x
+		var dist = speed * time * last_scale.x * last_scale.x * terrain.player_speed_multiplier
```
fleskesvor: Can `last_scale.x * last_scale.x` be written as `last_scale.x ** 2` or `last_scale.x ^ 2`? Should take precendence over multiplication and is more concise.

device/globals/game.gd
mjtorn: Yes. However, I force-pushed this change, because I got a complaint somewhere `clicked()` that the variable wasn't found! I declared it separately in both the `if` and the `else` which may have been a problem, but this is a sure-fire way of making it work.

device/globals/player.gd
mjtorn: Suppose it could be. Not my original code but I can push that in.

mjtorn: Actually it can not. I either get a syntax error on `last_scale.x ** 2` or unsupported operand for `last_scale.x ^2` so we're stuck with what we got.

device/globals/terrain.gd
```diff
@@ -9,6 +9,8 @@ export(int, "None", "Scales", "Lightmap") var debug_mode = 1 setget debug_mode_u
 export var lightmap_modulate = Color(1, 1, 1, 1)
 export var scale_min = 0.3
 export var scale_max = 1.0
+export var player_speed_multiplier = 1.0  # Compensate if you want a small player to move fast
```
fleskesvor: Could change the comment to something like "override player speed in current scene", since there might be use cases where one would want to increase speed not just because the player is smaller.

device/globals/terrain.gd
```diff
@@ -9,6 +9,8 @@ export(int, "None", "Scales", "Lightmap") var debug_mode = 1 setget debug_mode_u
...
+export var player_doubleclick_speed_multiplier = 1.5  # Make the player move faster when doubleclicked
```
fleskesvor: I've been thinking about this for ages, but never got around to implementing it, so thanks! :slightly_smiling_face:

device/globals/player.gd
fleskesvor: Yeah, it's `pow(last_scale, 2)`:

http://docs.godotengine.org/en/3.0/classes/class_@gdscript.html?highlight=pow#class-gdscript-pow

device/globals/terrain.gd
mjtorn: Done, force-pushed.

device/globals/game.gd
fleskesvor: Yeah, but couldn't you declare it at the top of the `clicked()` function, before the if-else-statement?

mjtorn: Yes. Pushed it that way now.

device/globals/terrain.gd
mjtorn: My pleasure!

PR #76, opened 2018-03-02
[RFC] Camera zoom WIP

mjtorn:
It would be useful to have a feature where the camera can zoom in. Say there's a conversation going, the camera could be set to zoom in on the characters on `:talk`.

Another use case is that the character walks somewhere and the camera pushes in on a close-up, and when the character walks back out, the camera pulls out.

I think there could technically be other cameras, but with only one visible at any time and limited support in ESC scripts, surely it'd be more pain to deal with. And those cameras need to zoom as well.

After some struggling, I sort of got zooming to work, but didn't realize in advance that scaling the viewport causes the action menu (and all that, tooltips etc) to get zoomed as well! There must be a better way!

Is there a way to filter out the HUD and dialog stuff and scale them down when scaling the viewport up? Does that question make sense? It should be easier than tagging all the TextureRects and sub-scene nodes with `affected_by_zoom` or whatever.

I've been asking around on IRC and searching online for a while, but no evident answer has popped up.

For some reason `camera.zoom = ...` has had no impact, does this work for anyone?

This branch is off my devel branch, which has apparently deviated a lot from master. Sorry. It is not meant to be merged, but to be discussed, though I did hope the commit history would be cleaner. Pay no attention to the commits older than 2018-03-02 ;) (I may sort my devel branch later and force-push it)

You can check out the `gh-mjtorn-devel-zoom` branch and update the submodule, I set it up so it points to a corresponding branch in my prototype submodule. It includes buttons for changing the zoom level to play around with. Unfortunately the TextureRects I used are white, as the background, so they are a bit awkward to find, but `look`ing at them changes the zoom. That means the camera zoom functionality is properly exported in the ESC scripts, so that's at least something.

The relevant HEAD in my prototype submodule is 7e27cdafe176e6405780680c4bda3832b9e952f3 (branch `room.002.zoom`.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "device/game"]
+	path = device/game
+	url = git+ssh://git@github.com/mjtorn/escoria-proto-submodule
--- a/README.md
+++ b/README.md
@@ -22,7 +22,19 @@ to the backers of the Dog Mendonça Kickstarter campaign.
 
 ## Usage
 
-Here be dragons.
+Here be dragons. For https://github.com/mjtorn/escoria/tree/devel you want either
+
+    $ git clone https://github.com/mjtorn/escoria/ --recurse-submodules
+
+to see the code, but you have to `git checkout -b devel origin/devel` to see where the money at.
+
+Alternatively you can use
+
+    $ git clone https://github.com/mjtorn/escoria/ --recurse-submodules -b devel
+
+to get there by default.
+
+(These instructions shall never be merged to the upstream master!)
 
 ## Documentation
 
--- a/device/globals/esc_compile.gd
+++ b/device/globals/esc_compile.gd
@@ -31,6 +31,7 @@ var commands = {
 	"camera_set_target": { "min_args": 1, "types": [TYPE_REAL] },
 	"camera_set_pos": { "min_args": 3, "types": [TYPE_REAL, TYPE_INT, TYPE_INT] },
 	"camera_set_zoom_height": { "min_args": 1, "types": [TYPE_INT] },
+	"camera_unset_zoom": { "min_args": 0 },
 	"autosave": { "min_args": 0 },
 	"queue_resource": { "min_args": 1, "types": [TYPE_STRING, TYPE_BOOL] },
 	"queue_animation": { "min_args": 2, "types": [TYPE_STRING, TYPE_STRING, TYPE_BOOL] },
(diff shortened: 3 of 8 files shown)


fleskesvor:
As discussed on IRC, I don't think this can be achieved* without also scaling the hud, including inventory and verb/action menu, as long as it's all part of the same viewport. I think the only good solution for this is to have the game scene be a part of a second viewport, but that would require a major rewrite, so it would be helpful if @punto- and @StraToN would comment before we start work on that.

This rewrite would give a scene tree looking something like this:

    game
    ├── hud
    │   ├── verb_menu
    │   ├── inventory
    │   └── action_menu
    └── viewport
        └── scene
            ├── background
            ├── terrain
            ├── player
            ├── item1
            └── item2

Which would mean that we can't load a room directly, but instead instanciate it into the game scene.

* That is achieved without hacky workarounds like scaling the hud elements the other way to counteract the effect.

mjtorn:
AFAIU this means the camera is moved out of `game` so that it can use `viewport`, right?

The problem may be that the children of `hud` are still scaled up, and would require that hacky workaround. I conducted a little experiment with it and couldn't get them scaled back.

Would this have performance problems, with checking if things have been scaled?

I also posted [a question on the Q&A](https://godotengine.org/qa/25100/scale-or-zoom-only-background-not-hud-pop-up-menu-etc), and we can wait a while to see if anyone replies, but judging by the (in)activity in some of the questions, we shouldn't hold our breaths.

fleskesvor:
I meant that the game and hud would use the default viewport and camera to not be affected by scaling, since we'd only scale the second viewport. And, yeah, that second viewport will also need a camera. I don't think the second viewport node will need to worry about whether or not it's scaled at all, but I will make a proof-of-concept as soon as possible to check that this doesn't cause problems with input events being offset from where they should occur.

mjtorn:
I dropped some links on IRC for discussion,

http://docs.godotengine.org/en/3.0/tutorials/2d/canvas_layers.html

http://docs.godotengine.org/en/3.0/tutorials/2d/2d_transforms.html

A friend of mine, who admittedly doesn't know Godot but also hasn't been broken by this, suggested that maybe viewports aren't the way to go but some kind of canvas layering.

Thoughts?

fleskesvor:
After making a simple demo project to test this, I agree that canvas layering is definitely the way to go:

https://github.com/fleskesvor/canvas-test

The project zooms in/out on mouse click, and as you can see, the scene node and its children are scaled together, while the sprite in the separate canvas layer remains unchanged.

This is very similar to how Escoria is structured today, so there likely shouldn't be too many changes needed to get this to work. The action menu must be placed within the canvas layer of the hud though, and there are probably some other issues as well. This means that your friend is right though, as I'm now also convinced this is the way to go.

fleskesvor:
Fixed by #80.

PR #77, opened 2018-03-05
Remove use_custom_camera

### LINKED ISSUE #60: What about the use_custom_camera setting?
It was discussed in #54 that this setting should remain in the code for now. It works while it's set to `true`, but it's not documented behavior anywhere.

@punto- or @StraToN should be summoned to state if this is a thing that needs to exist, and if so, where to document what it does exactly.

Of course, being a discussion issue for something that currently works, anyone can pitch in :)

My two cents say it has to go.

mjtorn:
Would close #60

### DIFF
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -155,13 +155,12 @@ func update_camera(time):
 			camera.set_position(cpos + dif.normalized() * dist)
 			pos = cpos + dif.normalized() * dist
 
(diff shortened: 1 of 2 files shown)


fleskesvor:
As agreed upon in #60, `escoria/platform/use_custom_camera=false` does not work, so it doesn't make sense to have this toggle until it's requested and fixed to work properly. Thanks, @mjtorn. I'm merging this now.

PR #78, opened 2018-03-06
Make cache system explicit

mjtorn:
This is the most likely cause for Heisenbugs and waste of time while developing something else. It was not even widely used.

Godot should take care of loading well enough without reinventing any wheels. Modern computers are fast. Kernels have file system caches. There are probably even more reasons why not to overcomplicate things.

Therefore, to simplify the code and to make caching-related bugs impossible, throw it out.

No perceptible difference in performance and the demo test scene works as expected.

The commits are split into smaller pieces if that makes discussion easier, and although I hope this history can be merged as-is, I can - of course - also squash them.

### DIFF
--- a/device/project.godot
+++ b/device/project.godot
@@ -32,6 +32,7 @@ application/save_data="res://globals/save_data.gd"
 application/speech_suffix=".spx"
 application/tooltip_lang_default="en"
 application/speech_path="res://demo/audio/speech/"
+platform/skip_cache=false
 platform/action_menu_scale=2
 platform/credits="res://demo/ui/credits/credits.txt"
 platform/dialog_option_height=1
--- a/device/globals/esc_compile.gd
+++ b/device/globals/esc_compile.gd
@@ -25,7 +25,6 @@ var commands = {
 	"%": { "alias": "label", "min_args": 1},
 	"jump": { "min_args": 1 },
 	"dialog_config": { "min_args": 3, "types": [TYPE_STRING, TYPE_BOOL, TYPE_BOOL] },
-	"queue_scene": { "min_args": 1 },
 	"sched_event": { "min_args": 3, "types": [TYPE_REAL, TYPE_STRING, TYPE_STRING] },
 	"custom": { "min_args": 2 },
 	"camera_set_target": { "min_args": 1, "types": [TYPE_REAL] },
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -402,6 +402,8 @@ func finished(context):
 	context.waiting = false
 
 func change_scene(params, context):
+	# It might be tempting to use `get_tree().change_scene(params[0])`,
+	# but this custom solution is safer around your scene structure
 	printt("change scene to ", params[0])
 	#var res = ResourceLoader.load(params[0])
 	check_cache()
@@ -469,6 +471,10 @@ func autosave_done(err):
 	last_autosave = OS.get_ticks_msec()
 
 func check_cache():
+	# Warm the cache from the hard-coded list, unless configured to skip
+	if ProjectSettings.get_setting("escoria/platform/skip_cache"):
+		return
+
 	for s in scenes_cache_list:
 		if s in scenes_cache:
 			continue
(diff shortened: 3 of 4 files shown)


StraToN:
Resourcequeue was probably some old stuff that was added to Escoria early, when there was no ResourceLoader. Which ended up using both. Great work !

mjtorn:
Thanks :) I'm still confused as to what would have been pulled off the cache there, because I'm sure it was used only for animations.

If caching is really required, it shouldn't be hard to resurrect the resource queue, or preferably see if Godot's mechanisms can be used.

In that case make it so it caches only adjacent scenes. Maybe the esc compiler would notify of possible `change_scene` calls or something, and the scene loader would take the scene and its resources from a cache. Making it like that, resolving things automatically, would also make it easier on game designers.

fleskesvor:
For the sake of posterity, I'll try to summarize @punto-'s explanaition of this feature from the Escoria channel on IRC.

The usage is explained in the [Floss Manuals tutorial on optimizing](https://fr.flossmanuals.net/creating-point-and-click-games-with-escoria/optimization/). The reason why a `ResourceLoader` for holding packed scenes alone isn't enough is that it would block the current thread while packing/loading, which might not have a noticeable impact on a small game with only a few rooms, but makes a big difference on a bigger game, and especially on mobile devices and consoles.

We should therefore keep this functionality, but if anything is broken, it needs to be fixed, and I think it we should see if some of this can be put in a separate singleton to limit the interaction points in `global_vm.gd` to isolate this functionality and make the code easier to understand.

mjtorn:
If needed, maybe the internals could be written using ResourcePreloaders or something more Godot-ish, but I won't have time to do anything beyond fixing this PR, I'm afraid :/

These are the action points off IRC:

1. Drop the undocumented `queue_scene` and `swap_scene`
1. Bring back the `skip_cache` setting but in a proper place and so it caches either everything or nothing, no the current "some are implicit" approach
1. Comment on why scene loading is not done the Godot way, like "Godot can't know your scene layout so it's possible to break it by simply loading a new scene. This solution also handles ESC scripts."

I'm not sure the last point will be eternally valid, but if there's a risk of things breaking now, leave it. Maybe someone will look into it in the future, because TBH the code looks a bit messy to me.

mjtorn:
I believe the three points are now addressed in the recent force-push.

Up for merging?

device/globals/scenes_cache.gd
```diff
@@ -1,4 +0,0 @@
-const scenes = [
```
fleskesvor: This file is still referenced in `global_vm.gd` on line 61:

https://github.com/mjtorn/escoria/blob/826c06f5a6e2170bde04f7b654cf0bebd037e868/device/globals/global_vm.gd#L61

device/project.godot
```diff
@@ -32,6 +32,8 @@ application/save_data="res://globals/save_data.gd"
 application/speech_suffix=".spx"
 application/tooltip_lang_default="en"
 application/speech_path="res://demo/audio/speech/"
+platform/skip_cache=false
+platform/path_to_cache_list="res://game/scenes_cache.gd"
```
fleskesvor: Shouldn't the default path be `res://game/scenes_cache.gd`, and a placeholder file exist in that location?

device/globals/scenes_cache.gd
fleskesvor: The declaration can remain, but the assignment should only happen if a scenes_cache has actually been set in the project settings. I think one should be set by default though, and I think it makes sense that the file is in `demo/`.

device/project.godot
mjtorn: As said on IRC, the `path_to_cache_list` was a failed experiment that ended up in the PR by accident. I was trying to make it configurable, so the game developer wouldn't have to commit into the framework code to change what gets cached.

I think this is related to bug godotengine/godot#6798 - not sure I interpret it correctly - but anyway the setting cannot be used for `preload()` for whatever reason.

I'll remove the setting and postpone worrying about separation of concerns into the future.

mjtorn:
@fleskesvor is this more satisfactory now that the non-working setting is gone?

It's a lower-priority thing to fix, so I'll open an issue for it so someone like @punto- or @StraToN who actually knows Godot (and Escoria) can chime in.

fleskesvor:
Yes, I agree that it's better to leave that out of this commit until we can make its location fully configurable from the project settings. Thanks for sorting this all out. :slightly_smiling_face:

PR #79, opened 2018-03-07
Camera zoom + Action menu from project settings

fleskesvor:
Enable zooming of game scene with esc commands by changing scale. An array is used to save previous scales so that one can easily revert to previous scales.

To make action menu unaffected by zoom, it needs to be added to the hud layer instead of as a direct child of the game node, so a project settings property is exposed which is used to instance the action menu from code, and also hide the verb menu if an action menu is found.


fleskesvor:
This branch is/will be fixed by #80.

PR #82, opened 2018-03-09
Zoom height

mjtorn:
This PR introduces an ESC command to set the zoom in height pixels.

It also fixes the bug when the player is standing on a sizemap and a zoom happens, now the sprite is scaled as well.

I threw in a fix for the `set_camera_pos` command for good measure, it must have been broken by the big regexp change way back.

@fleskesvor wanna confirm this is fine?

### DIFF
--- a/device/globals/esc_compile.gd
+++ b/device/globals/esc_compile.gd
@@ -29,6 +29,7 @@ var commands = {
 	"custom": { "min_args": 2 },
 	"camera_set_target": { "min_args": 1, "types": [TYPE_REAL] },
 	"camera_set_pos": { "min_args": 3, "types": [TYPE_REAL, TYPE_INT, TYPE_INT] },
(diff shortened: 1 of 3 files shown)


fleskesvor:
Zooming the player isn't necessary in my testing, since it's a child of the scene which is scaled. In fact, also scaling the player makes it much larger compared to the other elements in the scene.

Here's a comparison from the demo scene before and after scaling to 500:

https://i.imgur.com/UNCJU7A.png

https://i.imgur.com/aKylixy.png

mjtorn:
@fleskesvor I dropped the last commit. I may have been fooled by the fact the player was on a sizemap and walking around.

Wanna merge the two commits as they are?


Cheers!

fleskesvor:
Yes, this looks good. :slightly_smiling_face: Thanks, @mjtorn!

PR #88, opened 2018-03-14
Fix dealing with a clamped camera that tries to move

mjtorn:
I noticed that moving the camera from a corner to the center took a long time to start moving. After some digging around, it's because the camera would actually show they gray area outside the screen, which was properly fixed. Thus it started moving only after the camera though it was far enough from the bounds.

The solution is to check if the camera has been clamped down, and if so, reset its position. Without this fix it could take roughly a second for the camera to start moving, now it should take something like a delta-time between frames, like 30ms or whatever imperceptible.

This PR updates some pre-Godot3 code I found and renames (and comments on) some other stuff, so the functionality is hopefully clearer to someone who's never seen this before.

### DIFF
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -146,22 +146,31 @@ func update_camera(time):
 
 	var cpos = camera.get_position()
 
+	# The camera position is set to target when it's about to overstep it,
+	# or when it's moved there instantly.
+	# Compare the camera and target position until then
 	if cpos != pos:
-		#return
+		var v = pos - cpos  # vector to move along
+		var step = cam_speed * time  # pixel size of step to move
 
-		var dif = pos - cpos
-		var dist = cam_speed * time
-		if dist > dif.length() || cam_speed == 0:
+		# This is where we may overstep or move instantly
+		if step > v.length() || cam_speed == 0:
 			camera.set_position(pos)
-			#return
 		else:
-			camera.set_position(cpos + dif.normalized() * dist)
-			pos = cpos + dif.normalized() * dist
+			pos = cpos + v.normalized() * step
+			camera.set_position(pos)
 
 	var half = game_size / 2
-	pos = _adjust_camera(pos)
+
+	var clamp_data = _clamp_camera(pos, half)
+	var camera_clamped = clamp_data[0]
+	pos = clamp_data[1]
+
 	var t = Transform2D()
-	t.origin = (-(pos - half))
+	t.origin = half - pos
+
+	if camera_clamped:
+		camera.set_position(pos)
 
 	get_node("/root").canvas_transform = t
 
@@ -182,8 +191,13 @@ func camera_zoom_out():
 	if current_scene and current_scene is preload("res://globals/scene.gd") and prev_state:
 		current_scene.scale = prev_state["scale"]
 
-func _adjust_camera(pos):
-	var half = game_size / 2
+func _clamp_camera(pos, half):
+	# Helper function clamps camera within the given camera limits; if we try
+	# to show anything outside the limits, we return a position that doesn't break the limits.
+	# `half` is passed in because there's a good chance it gets altered when zoomed in
+	# Returns whether or not the camera was clamped and what the new position is regardless
+
+	var orig_pos = pos
 
 	if pos.x + half.x > camera_limits.position.x + camera_limits.size.x:
 		pos.x = (camera_limits.position.x + camera_limits.size.x) - half.x
@@ -195,7 +209,7 @@ func _adjust_camera(pos):
 	if pos.y - half.y < camera_limits.position.y:
 		pos.y = camera_limits.position.y + half.y
 
-	return pos
+	return [orig_pos != pos, pos]
 
 func set_cam_limits(limits):
 	camera_limits = limits
@@ -738,7 +752,7 @@ func _ready():
 			print("s is ", s)
 			res_cache.queue_resource(s, false, true)
 
-	printt("********** vm calling get scene", get_tree(), get_tree().get_root())
+	printt("********** vm calling get scene", get_tree(), get_node("/root"))
 
 	achievements = preload("res://globals/achievements.gd").new()
 	achievements.start()
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -161,9 +161,9 @@ func update_camera(time):
 	var half = game_size / 2
 	pos = _adjust_camera(pos)
 	var t = Transform2D()
-	t[2] = (-(pos - half))
+	t.origin = (-(pos - half))
 
-	get_node("/root").set_canvas_transform(t)
+	get_node("/root").canvas_transform = t
 
 func camera_set_zoom_height(zoom_height):
 	var scale = game_size.y / zoom_height
--- a/device/globals/player.gd
+++ b/device/globals/player.gd
@@ -5,7 +5,6 @@ extends Node2D
 var task
 var walk_destination
 var animation
-var vm
 var terrain
 var walk_path
 var walk_context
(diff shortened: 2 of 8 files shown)


device/globals/global_vm.gd
```diff
@@ -28,6 +28,9 @@ var cam_target = null
 var cam_speed = 0
 var camera
 var camera_limits
+var clamp_data = []  # [was the camera clamped, camera position]
+var camera_clamped  # clamp_data[0]
+var pos  # clamp_data[1] and other target camera positions
```
fleskesvor: It looks like these are only used in `update_camera()`. Should be local.

device/globals/global_vm.gd
```diff
@@ -146,24 +148,33 @@ func update_camera(time):
...
 	var half = game_size / 2
-	pos = _adjust_camera(pos)
+
+	clamp_data = _clamp_camera(pos, half)
+	camera_clamped = clamp_data[0]
+	pos = clamp_data[1]
+
 	var t = Transform2D()
-	t[2] = (-(pos - half))
+	t.origin = (-(pos - half))
```
fleskesvor: Could `(-(pos - half))` be replaced with `-(pos - half)`?

device/globals/player.gd
```diff
@@ -317,7 +317,7 @@ func _ready():
 		return
 
 	animation = get_node("animation")
-	vm = get_tree().get_root().get_node("vm")
+	vm = get_node("/root/vm")
```
fleskesvor: I think this line can be removed. `vm` is a singleton and can be accessed by just its name.

device/ui/main_menu.gd
```diff
@@ -102,7 +102,7 @@ func _ready():
 	get_node("exit").connect("pressed", self, "_on_exit_pressed")
 	#get_node("settings").connect("pressed", self, "settings_pressed")
 	#get_node("credits").connect("pressed",self,"credits_pressed")
-	vm = get_tree().get_root().get_node("vm")
+	vm = get_node("/root/vm")
```
fleskesvor: Same here.

device/globals/global_vm.gd
mjtorn: Hold on, isn't this the same as `t.origin = half - pos`? Would that be cleaner?

This is most likely the line that will require knowledge when I make a plz halp pull request, so I'll defer judgement to you entirely.

device/globals/global_vm.gd
mjtorn: `pos` is used elsewhere, but my actual reason for defining these up here is that I don't know how Godot does variables and how expensive they are. Defining these variables on every update cycle felt like extra overhead.

Does this make any sense? Do you still want me to move them? Because in general I just want this in ;)

fleskesvor:
This looks like a good and useful improvement. :slightly_smiling_face: 

One note on code style in general though, since this PR also modernizes old code: I think we should decide on whether to use the old `get_node()` function to get nodes or the new `$` syntax. I prefer the latter, since it makes for more compact (and more readable, in my opinion) syntax. Usually you can omit quotation marks with the new syntax, but I'm not sure if that works with leading slashes, so `get_node("/root")` might become `$"/root"` instead of just `$/root`.

mjtorn:
Let's get this merged and then do a new PR for updating the syntax. I'm ok with the `$` one and it's already started creeping in.

`get_node("/root")` becomes `$"/root"` which does look a bit ugly, but should we do `$"foo"` just to make our eyes get used to the quotes?

device/globals/global_vm.gd
fleskesvor: Yes, unless it's a one-element tuple, which I don't think GDScript has, so I think you're right. Come to think of it, I'm pretty sure it's a Vector2, so that should be correct.

device/globals/global_vm.gd
fleskesvor: I think we should prioritize readability over performance until performance becomes a problem, so  I prefer locally scoped variables.

device/globals/global_vm.gd
fleskesvor: Also, wouldn't it be better to either use an array/dictionary or two scalar instead of both?

fleskesvor:
Yes, I agree that it's a good idea to settle on a single unified style, so I'm happy with always using `$"" `, even when the quotes aren't needed.

device/globals/global_vm.gd
mjtorn: Fair enough, I'll remove the comment from `pos` and move the other two down. Expect a force-push hopefully soon.

device/globals/global_vm.gd
mjtorn: The pleasures of weak typing ;)

`pos` is a Vector2 and so's `half` by virtue of coming from `Viewport.size`.

I don't understand "Also, wouldn't it be better to either use an array/dictionary or two scalar instead of both?" but we can take that up on IRC after I force-push `half - pos`.

mjtorn:
There's a lot more of modernization waiting to happen so I recommend those get handled separately and keep what's now in the PR here. No need to bloat the PR and potentially introduce risk while changing multiple things at once.

fleskesvor:
What I meant is that camera_clamped and pos is also in clamp_data, but I understand that clamp_data is only used for unpacking return values, I think. Can you unpack multiple return values directly, like you can in Python? Also, is there a reason for pos not being a local variable?

fleskesvor:
Other than that this looks great now. :+1:

mjtorn:
On Thu, Mar 15, 2018 at 07:51:14AM -0700, Eivind wrote:
>What I meant is that camera_clamped and pos is also in clamp_data, but I understand that clamp_data is only used for unpacking return values, I think. Can you unpack multiple return values directly, like you can in Python?

I tried as hard as I could but the Python unpacking did not work. IIRC it did
not set the values at best and died of syntax error at worst.

> Also, is there a reason for pos not being a local variable?

Pretty sure I saw it in the file somewhere else. Can't check now, can you see
it? I will fix asap if it's a mistake I made.

fleskesvor:
> I tried as hard as I could but the Python unpacking did not work. IIRC it did
not set the values at best and died of syntax error at worst.

Yeah, doesn't look like there's a shorthand for it, and in that case I agree that it's a good idea to reassign to two scalars, since it clarifies what they do by explicit naming.

> Pretty sure I saw it in the file somewhere else. Can't check now, can you see
it? I will fix asap if it's a mistake I made.

In current master, two separate variables `pos` are declared on lines 131 of `update_camera()` and 578 of `save()`. One is calculated and the other is copied from a node position, so I think it's better to change those back to the original declarations to avoid confusion.

device/globals/global_vm.gd
```diff
@@ -146,24 +146,36 @@ func update_camera(time):
...
 
 	var half = game_size / 2
-	pos = _adjust_camera(pos)
+
+	var clamp_data = []
+	var camera_clamped
+
+	clamp_data = _clamp_camera(pos, half)
+	camera_clamped = clamp_data[0]
+	pos = clamp_data[1]
```
fleskesvor: I think declaration and assignment should be on the same line in this case, since it's immediately clear that `clamp_data` is an array, since values are fetched by index on the next line, ie.:

    var clamp_data = _clamp_camera(pos, half)
    var camera_clamped = clamp_data[0]
    var pos = clamp_data[1]

fleskesvor:
Other than that, and the inline comment, this looks good to me. 🙂

device/globals/global_vm.gd
mjtorn: Actually `var pos` was not on that line, I restored it to where it originally was, along with the other `var pos`, so that should be all fine now.

fleskesvor: Thanks, @mjtorn. 🙂 Merging this now.

PR #89, opened 2018-03-14
esc commands: use_action_menu and set_speed, with documentation

mjtorn:
`use_action_menu` "soft-disables" an item.

`set_speed` is useful before `walk` to make an NPC walk faster than the player can catch up with. It could also be used in things like dream sequences or whatever to mess with the player.


device/globals/esc_compile.gd
```diff
@@ -13,9 +13,11 @@ var commands = {
 	"inventory_remove": { "min_args": 1 },
 	"inventory_open": { "min_args": 1, "types": [TYPE_BOOL] },
 	"set_active": { "min_args": 2, "types": [TYPE_STRING, TYPE_BOOL] },
+	"use_action_menu": { "min_args": 2, "types": [TYPE_STRING, TYPE_BOOL] },
 	"stop": true,
 	"repeat": true,
 	"wait": true,
+	"set_speed": { "min_args": 2, "types": [TYPE_STRING, TYPE_INT] },
```
fleskesvor: Might be better with a more general purpose `set_property` which takes a global_id, the property to set and the new value?

fleskesvor: That could replace `use_action_menu` as well.

fleskesvor: You'd probably have to use [Object.get()](http://docs.godotengine.org/en/3.0/classes/class_object.html#id1) to check the type of the property, cast it to the right type and call [Object.set()](http://docs.godotengine.org/en/3.0/classes/class_object.html#id4) to set the new value.

mjtorn: Naturally I did consider this, but complicated checking and a lack of need made me go the other way.

I tried coming up with use cases and most of them, like changing `use_custom_z` just don't make sense, so having a general `set_property` opens up a lot of options a game developer may not want to bother with. IMO the ESC scripts should be simple.

If you want, `set_property` could be there, but I'd advocate _strongly_ for having short-hands for the most common cases _as well_ to keep it simple.

If you don't feel very strongly about this, I advocate not implementing `set_property`.

fleskesvor:
Yes, you're right. If there's a future demand for more setters, we refactor this then. One thing though, could you prefix the esc command use_action_camera with set_ so that it has the same name as the GDScript function? And also place it next to the other setter?

mjtorn:
On Thu, Mar 15, 2018 at 07:57:07AM -0700, Eivind wrote:
>Yes, you're right. If there's a future demand for more setters, we refactor this then. One thing though, could you prefix the esc command use_action_camera with set_ so that it has the same name as the GDScript function? And also place it next to the other setter?

Sure, thanks, I'll address that asap too :)

fleskesvor:
Thanks. 🙂I'll merge as soon as it's fixed.

fleskesvor:
I didn't realise you'd updated this. Merging now, and thanks for the contribution. 🙂

Issue #44, opened 2018-01-10
Rewrite Escoria framework


StraToN:
We've come to a point where maintaining Escoria framework gets harder and harder. Escoria was written early in the history of Godot and has not evolved at the same pace. 

Godot Engine 3.0 stable release is very close, and we are facing many problems and bugs which are hard to solve (#43 https://github.com/flossmanualsfr/escoria/issues/12 #41 #35 ...)

To be able to clean the code from its useless parts and fix the most problematic bugs, I propose a complete rewrite of Escoria.

## Pros
- We will be able to provide a cleaner code thanks to Godot 3.0 evolutions, in an easier way than simply mutate the current code. Would fix #10 
- Reorganize elements so users know what included scripts/scenes they may need to modify. As much as possible, users should not modify base scripts and scenes.
- I'd personally like to "open" the global_vm (currently accessed through .esc files) to other ways, such as graph tools (for dialogs, game logic, items use, etc). 
- Keep .esc files as well. They're also a good choice for game scripting. Simply, adding more possibilities to achieve game scripting is, IMHO, a good thing for many users, especially if all solutions are compatible one with the other.
- We can integrate Escoria closer to the editor by creating a plugin (with custom nodes) and editor tools directly integrated in the AssetLib. Would fix #5 

## Cons
- As a new version, we do not need it to be fully compatible with previous versions. Thus projects based on current Escoria will have some hard time moving to it if they want to.

Share your ideas here.

# TODOs
- .esc files edition in Godot Script Editor (https://github.com/godotengine/godot/issues/1390)
- Define 1 unique script to manage items and NPC (name it "interactive" for example) and allow:
  - Wandering around by .esc script and/or Position2D/3D definition in the editor
  - ?
- Custom nodes:
  - RoomExit: define an Area2D>CollisionPolygon2D with a target scene.
  - ?
- Manage different types of terrains
  - 2D:
    - Background image
    - TileMap
  - 3D

# Target workflow:
![escoria1](https://user-images.githubusercontent.com/1289207/35041347-b62f03e4-fb84-11e7-84c9-e59c6061b0c6.png)

fleskesvor:
I think this is a good idea, even if it's going to be a massive job. Are you aiming to keep the framework in a functional state during the process, and do you have ideas on how to best break the job up into actionable tasks?

> Reorganize elements so users know what included scripts/scenes they may need to modify. As much as possible, users should not modify base scripts and scenes.

[...]

PR #90, opened 2018-03-14
Exit node

mjtorn:
Steals some ideas from #44 but uses `TextureButton` because nothing else will work at all.

I plea that this gets merged anyway, because it's not horribly broken, as I'm using it successfully in my proto. Thus it should serve as a solid base for when `Area2D -> CollisionPolygon2D` becomes reality.

### DIFF
--- a/device/globals/exit.gd
+++ b/device/globals/exit.gd
@@ -1,0 +1,44 @@
+extends "res://globals/interactive.gd"
+
+export(String, FILE, ".esc") var esc_script  # must contain :dblclick
+export var global_id = ""
+export var tooltip = ""
+
+var event_table = {}
+
+func get_tooltip():
+	if TranslationServer.get_locale() == ProjectSettings.get_setting("escoria/application/tooltip_lang_default"):
+		return tooltip
+	else:
+		if tr(tooltip) == tooltip:
+			return global_id+".tooltip"
+		else:
+			return tooltip
+
+func mouse_enter():
+	vm.hover_begin(self)
+	var tt = get_tooltip()
+	var text = tr(tt)
+	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", text)
+
+func mouse_clicked(event):
+	if !(event is InputEventMouseButton):
+		return
+
+	if event.doubleclick:
+		vm.run_event(event_table["dblclick"])
+
+func mouse_exit():
+	vm.hover_end()
+	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", "")
+
+func _ready():
+	if esc_script != "":
+		event_table = vm.compile(esc_script)
+		if !("dblclick" in event_table):
+			vm.report_errors(esc_script, ["Missing :dblclick"])
+
+	self.connect("mouse_entered", self, "mouse_enter")
+	self.connect("gui_input", self, "mouse_clicked")
+	self.connect("mouse_exited", self, "mouse_exit")
+
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -424,7 +424,6 @@ func finished(context):
 func change_scene(params, context):
 	# It might be tempting to use `get_tree().change_scene(params[0])`,
 	# but this custom solution is safer around your scene structure
-	printt("change scene to ", params[0])
 	#var res = ResourceLoader.load(params[0])
 	check_cache()
 	main.clear_scene()
@@ -434,7 +433,12 @@ func change_scene(params, context):
 	res_cache.clear()
 	var scene = res.instance()
 	if scene:
-		main.set_scene(scene)
+		if params.size() == 1:
+			printt("change scene to ", params[0])
+			main.set_scene(scene)
+		else:
+			printt("change scene to ", params[0], " position ", params[1])
+			main.set_scene(scene, params[1])
 	else:
 		report_errors("", ["Failed loading scene "+params[0]+" for change_scene"])
 
--- a/device/globals/main.gd
+++ b/device/globals/main.gd
@@ -22,15 +22,14 @@ func clear_scene():
 	current.free()
 	current = null
 
-func set_scene(p_scene):
+func set_scene(p_scene, p_start_pos_node):
 	if current != null:
 		clear_scene()
 
 	if !p_scene:
 		return
-	var root = get_tree().get_root()
-	root.add_child(p_scene)
-	set_current_scene(p_scene)
+	get_node("/root").add_child(p_scene)
+	set_current_scene(p_scene, p_start_pos_node)
 
 func get_current_scene():
 	return current
@@ -73,11 +72,15 @@ func menu_collapse():
 		i -= 1
 		menu_stack[i].menu_collapsed()
 
-func set_current_scene(p_scene, events_path=null):
+func set_current_scene(p_scene, start_pos_node=null, events_path=null):
 	#print_stack()
 	current = p_scene
 	var root = get_tree().get_root()
 	root.move_child(p_scene, 0)
+
+	if start_pos_node:
+		var start_pos = current.get_node(start_pos_node).position
+		vm.get_object("player").teleport_pos(start_pos.x, start_pos.y)
 
 	if events_path:
 		vm.load_file(events_path)
--- a/device/globals/scene_base.gd
+++ b/device/globals/scene_base.gd
@@ -3,5 +3,6 @@ extends Node
 export(String, FILE) var events_path = ""
 
 func _ready():
-	main.call_deferred("set_current_scene", self, events_path)
+	var start_pos_node = null
+	main.call_deferred("set_current_scene", self, start_pos_node, events_path)
 
--- a/device/globals/vm_level.gd
+++ b/device/globals/vm_level.gd
@@ -142,11 +142,15 @@ func walk_block(params):
 	return _walk(params, true)
 
 func change_scene(params):
+	var start_pos_node
+
 	# looking for localized string format in scene. this should be somewhere else
 	var sep = params[0].find(":\"")
+	if params.size() == 2:
+		start_pos_node = params[1]
 	if sep >= 0:
 		var path = params[0].substr(sep + 2, params[0].length() - (sep + 2))
-		vm.call_deferred("change_scene", [path], current_context)
+		vm.call_deferred("change_scene", [path, start_pos_node], current_context)
 	else:
 		vm.call_deferred("change_scene", params, current_context)
 
--- a/doc/esc_reference.md
+++ b/doc/esc_reference.md
@@ -112,8 +112,9 @@ set_globals i/* false
 - wait seconds
   Blocks execution of the current script for a number of seconds specified by the "seconds" parameter.
 
-- change_scene path
-  Loads a new scene, specified by "path"
+- change_scene path position
+  Loads a new scene, specified by "path". Optional parameters:
+  - position is the name of a Position2D node where the player will be put.
 
 - teleport object1 object2
   Sets the position of object1 to the position of object2


device/globals/exit.gd
```diff
@@ -0,0 +1,44 @@
+extends TextureButton  # XXX: Use TextureButton until background is fixed so CollisionPolygon2D will work
```
fleskesvor: As discussed on IRC, it might be better if this script shares a common base with `item.gd`, so that less code is duplicated. Eg. a `hotspot.gd` which extends `interactive.gd` and is extended by `item.gd` and this script. Is that in sync with your thoughts as well?

device/globals/exit.gd
```diff
@@ -0,0 +1,44 @@
...
+
+export(String, FILE, ".esc") var esc_script  # must contain :dblclick
```
fleskesvor: What is the purpose of `:dblclick`? Wouldn't it make more sense to have a project setting for making double-click on exit nodes teleport the player to the exit and have a single `:exit` action or something trigger other events, whether the player fast-travelled to the exit or just walked over?

device/globals/global_vm.gd
```diff
@@ -425,6 +425,9 @@ func change_scene(params, context):
 	# It might be tempting to use `get_tree().change_scene(params[0])`,
 	# but this custom solution is safer around your scene structure
 	printt("change scene to ", params[0])
+	var start_pos_node = null
+	if params.size() == 2:
+		start_pos_node = params[1]
```
fleskesvor: Does `start_pos_node` need to be declared? I feel like a comment over the call to `main.set_scene()` explaining what `params[1]` might be would be better.

fleskesvor:
I'm not 100% convinced that it's necessary to set the position in the next scene from the current scene, since that could be solved by using globals and `:ready`. My main concern is that it leads to more code complexity. What are your thoughts on this?

device/globals/exit.gd
mjtorn: Exactly this. Have smaller explicit scripts for these things.

device/globals/exit.gd
mjtorn: Especially if exit were replaced with hotspot, having scriptable actions makes sense to me.

I may have stated this somewhere, but if there was a single teleport, your game couldn't implement the scenario where the player tries to leave a screen and instead says "I'm not finished here yet." or something. At least not easily.

device/globals/global_vm.gd
mjtorn: I actually like your approach better now, it makes the prints also clearer. I'll force-push that.

mjtorn:
Using examples from my prototype, the exit node would be like

```
:dblclick
set_global previous_exit_from_bridge true
```
or something, because ”set_global name value Changes the value of the global flag "name" with the value. Value can be "true" or "false"”. These kinds of boolean globals clobber the namespace and force the developer to mentally track a lot of variables.

Then the office `main.esc` would be like
```
:ready
teleport player from_bridge [previous_exit_from_bridge]
```
which feels a bit counter-intuitive as well.

Therefore from a developer point of view, it's less cognitive load to say "change the scene and put the player in this position by the door", but in ESC script instead of English.

Consider my alternative exit node at the bridge
```
:dblclick
change_scene res://game/rooms/001.office/main.tscn from_bridge
```

Which would you rather use?

... I do wonder how actual games made with Escoria have dealt with this, if there's code that didn't make the open-source cut or do the players just appear wherever or what?

device/globals/exit.gd
mjtorn: I'm force-pushing a version that extends `interactive.gd`, it's clearly the cleaner approach and it works.

It does introduce `Speed`, `Scale on map` and `Light on map` which don't necessarily make sense, unless you want to alter the lighting of an exit under some conditions. Still it's better to expose this stuff, because it draws attention to things that may be relevant to refactor later.

For example, do items require `Speed` if there's going to be an `npc.gd` script?

fleskesvor:
So if we implement globals of different types and support eq, gt, lt, etc. in esc scripts this would be easier, right, because then you could check the value of eg. a last_exit global against a few different types and set the right position on :ready?

mjtorn:
I suppose we could but then we need a comparison operator that's separate from integers or there'll be pain with whether or not esc should parse `1` as `1` or `"1"` and that's a very special hell to be in. My idea with counters was to be exactly only integers, not complete comparisons.

But yeah, if there was a comparison operator for strings, we could do that. Technically. Still my gut tells me that having a position node and optionally naming it on `change_scene` isn't that bad or that the code is overtly complex due to it.

device/globals/exit.gd
fleskesvor: Wouldn't you have state variables that you set when each of the goals of the scene are achieved anyway though, like if you have a certain item in your inventory (which might again have other dependencies?

If `:dblclick` is a separate event, you need to do the same checks in both `:exit` and `:dblclick`

fleskesvor:
There are only a few built-in basic types though, and if we only support (and document) use of booleans, integers and strings, I don't think it would be a big deal to safe-guard with type checks.

No, it's not so bad, but I feel that it's not so good for readability to have one function do two things at once, and kind of acts on two scenes at once. Also, to me it feels more natural that you can find the esc script that places the character in a scene in the same directory as the scene itself. Otherwise, if you rearrange things in your scene and place obstacles where there once were entrances, you have to hunt around in other places to find where you set those positions.

mjtorn:
On Thu, Mar 15, 2018 at 11:04:06PM +0000, Eivind wrote:
>
>> @@ -0,0 +1,44 @@
>+extends TextureButton  # XXX: Use TextureButton until background is fixed so CollisionPolygon2D will work
>+
>+export(String, FILE, ".esc") var esc_script  # must contain :dblclick
>
>Wouldn't you have state variables that you set when each of the goals of the scene are achieved anyway though, like if you have a certain item in your inventory (which might again have other dependencies?

Yeah, but the event would check for those and branch accordingly, yes?

>If `:dblclick` is a separate event, you need to do the same checks in both `:exit` and `:dblclick`

I never thought of an `:exit` event. Would that be configures, like "on
doubleclick to this node, run :exit"? That sounds like extra indirection.

Or do you mean that different types of hotspots have different hard-coded
event names for doubleclicks? This may be cumbersome if you want to change a
node's script, though I don't know if that's a relevant concern. IMO a generic
event handler type like doubleclick is straight-forwad.

When I started tearing interactive.gd and item.gd apart for fun, I had to drop
this requirement, because obviously not everything has to be doubleclickable.

mjtorn:
On Thu, Mar 15, 2018 at 11:11:16PM +0000, Eivind wrote:
>There are only a few built-in basic types though, and if we only support (and document) use of booleans, integers and strings, I don't think it would be a big deal to safe-guard with type checks.

I detest writing parsers with a profundity that's hard to put in words, but
maybe it could work. `var foo = "1" < 2` for example dies on a type check,
so maybe looking for double quotes is fairly easy and won't lead to
javascript-esque side effects.

>No, it's not so bad, but I feel that it's not so good for readability to have one function do two things at once, and kind of acts on two scenes at once. Also, to me it feels more natural that you can find the esc script that places the character in a scene in the same directory as the scene itself. Otherwise, if you rearrange things in your scene and place obstacles where there once were entrances, you have to hunt around in other places to find where you set those positions.

This is a case of people thinking differently. I read my version as "Change
scene (to) res://foo.tscn [at optional location]" and you have a maybe more
declarative view on things. I usually like declarative programming, but
haven't thought of that pattern in ESC scripts.

Having obstacles is one thing, but when editing the scene, you would take care
of the entry Position2D nodes as well.

Still, having them contained also makes sense and I don't feel deeply enough
about "being more right", which isn't productive when both alternatives work,
so let's see what kind of pain it is to support double quotes and go with
that. Not my project etc.

A best practice that says

```
:ready

# Set things up
teleport player from_lake [last_exit eq "lake"]
teleport player from_sauna [last_exit eq "sauna"]

# Do things
```

is quite readable and my concern is mainly about avoiding touching the parser.

I do wonder how [!i/foo, !i/bar] would work with comparison operators.
Will we end up in a quagmire doing `say player "Ah, lake and sauna <3"
[last_exit ne "lake", last_exit ne "sauna"]` (did not validate the logic,
that is only a syntactical example)

I will touch the parser in the interest of putting this to bed with high
hopes it's a quick thing to fix.

-- 
mjt

fleskesvor:
> Yeah, but the event would check for those and branch accordingly, yes?

> Or do you mean that different types of hotspots have different hard-coded
event names for doubleclicks? This may be cumbersome if you want to change a
node's script, though I don't know if that's a relevant concern. IMO a generic
event handler type like doubleclick is straight-forwad.

> When I started tearing interactive.gd and item.gd apart for fun, I had to drop
this requirement, because obviously not everything has to be doubleclickable.

No, I mean that `:dblclick` isn't game logic, and shouldn't be handled by an esc script. The esc script should only care that a player reached the exit, whether they walked or ran. I think it's better to check for  `event.doubleclick` in the input function and teleport the player to the exit, if double-click to teleport is configured in project settings and/or on the exit node.

fleskesvor:
Ok, I see what you mean regarding teleport position now. In both cases, that would be configured in the destination scene, either with esc scripts or positioning of nodes, so I think it's mostly a matter of taste.

mjtorn:
On Fri, Mar 16, 2018 at 01:50:31PM +0000, Eivind wrote:
>No, I mean that `:dblclick` isn't game logic, and shouldn't be handled by an esc script. The esc script should only care that a player reached the exit, whether they walked or ran. I think it's better to check for  `event.doubleclick` in the input function and teleport the player to the exit, if double-click to teleport is configured in project settings and/or on the exit node.

Hmh, ok. I can sort of understand having the double-click be Godot-level and
not Escoria-level logic, so it has to be "aliased".

Of course then the `:exit` has to be triggered on collision or proximity or
something, and double-click-to-teleport is a node-level configuration because
it's (probably) relevant only for exit nodes.

This is too kämpemykket for me to do any time soon, but hopefully I'll get back
to the code at some point. I'm willing to accept PRs for my fork and if you or
someone wants to usurp this PR and base something on it and get it merged,
that's fine by me too.

-- 
mjt

fleskesvor:
Ok. I've checked out your branch and will take a closer look at it. I'll also see if I can rewrite the parser to make it easier to extend.

mjtorn:
On Sat, Mar 17, 2018 at 01:58:53AM -0700, Eivind wrote:
>Ok. I've checked out your branch and will take a closer look at it. I'll also see if I can rewrite the parser to make it easier to extend. 

That'd be awesome, thanks!

I haven't had the opportunity to look at the parser _at all_, but I don't
think it needs a complete rewrite. Sometimes Escoria has given us lucky
breaks and maybe just altering the `set_global` logic so it doesn't enforce
`TYPE_BOOL` is enough. Not sure what it would be instead, `TYPE_STRING` and
some parsing might work, maybe? And the corresponding check.

If you get this going in a separate PR, it'd be real helpful for rewriting
the exit node/hotspot code :)

mjtorn:
Just taking a moment to dump the ideas here that I think we may otherwise forget.

Beside extending the parser, I think there was talk about making the player a `KinematicBody2D` to support collisions with polygons, wasn't it, @fleskesvor? So then doubleclicks on polygons could be quick exits, collisions could trigger all sorts of things.

Also NPCs could be kinematic so there'd be an npc.gd script.

Was there something else beside supporting strings in variables and those type changes that are blocking the exit (no pun intended ;))?

fleskesvor:
Yes, I think it might be a good idea to make the player a KinematicBody2D, because then I think it should be possible to use built-in functionality in Godot to check if it enters an exit/sensor. To make that work, the background needs to be a Sprite with an Area2D for input, so we need to document this and/or start thinking about deprecating TextureRect backgrounds.

The other thing I'm not sure how to best support is moving the player past a sensor/exit node on their way to the exit, which should block them from moving, whether they walk normally or double-click to teleport to the exit. An example of this feature is when you try to cross the bridge outside Woodtick in MI2 and are shaken down by Largo LeGrande. This isn't something that is supported in Escoria currently though, so that can be implemented later.

fleskesvor:
I've updated [my exit-node branch](https://github.com/fleskesvor/escoria/tree/exit-node) to include a functional Area2D exit node using a KinematicBody2D player, and I think there are some clear advantages of using Area2Ds over Controls:

- It's easy to make the node trigger every time the player enters, whether it's their final destination or it's just something along the way, like a puddle, which should trigger some splash animation or the actual exit
- It's easy to make irregularly shaped exits using in-editor tools
- Area2Ds don't block input, so walking works as normal using the terrain, without workarounds

The main con is that it's a bit more work to configure a KinematicBody2D to trigger exits, since they must have a collision shape, but using default position and size at the bottom of a player's feet should work just fine in most cases, and the feet will already be the base of the player scene to make the player move around the terrain correctly, so maybe this isn't an issue.

fleskesvor:
Closed by #124

PR #94, opened 2018-03-18
Quit by windows X button 3-17-18, and the project name is now "Prototype-Escoria"

lavaduderDev:
Broken down from my previous Pull request. This gives power to close Escoria by simply hitting the X button on a window.

I also renamed the project from 'prototype' to prototype Escoria, because I thought that sounded too generic.

### DIFF
--- a/device/project.godot
+++ b/device/project.godot
@@ -10,7 +10,7 @@ config_version=3
 
 [application]
 
-config/name="Prototype_Escoria"
+config/name="Escoria Prototype"
 run/main_scene="res://globals/scene_main.tscn"
 
 [autoload]
--- a/device/project.godot
+++ b/device/project.godot
@@ -10,7 +10,7 @@ config_version=3
 
 [application]
 
-config/name="Prototype"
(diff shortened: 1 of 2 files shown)


device/project.godot
```diff
@@ -10,7 +10,7 @@ config_version=3
 
 [application]
 
-config/name="Prototype"
+config/name="Prototype_Escoria"
```
fleskesvor: Could you change this to "Escoria Prototype" instead, without the underscore, since that reads better`

fleskesvor:
I think it's a good idea to make this configurable. Eventually, we might want to make it possible to open the menu on force quit too, but we can worry about that later.

If you make a new commit with the suggested name change, I can walk you through squashing those commits into one commit.

fleskesvor:
Thanks, @lavaduderDev. 🙂Now we could remove the change to the project name on the first commit using rebase and `e`/`edit`, but I don't think we'll bother with that. I'm merging this now.

PR #97, opened 2018-03-23
Clamp action menu within viewport

### LINKED ISSUE #91: Action menu and maybe others outside the camera limits
If you open the action menu close to the edge, it's partly outside the camera limits or viewport or whatever.

All HUD elements should also be clamped within the view.

Does someone know an easy fix for this to put in a PR?

mjtorn:
This alters the action menu so it has to inherit from a Container class.

The reason behind it is that Node2D boundary calculation may or may not
be impossible, but Containers do it naturally.

It has also been said that you should always use Control nodes for UI,
so this feels like a more idiomatic approach.

Partially fixes #91

### DIFF
--- a/device/demo/ui/action_menu.tscn
+++ b/device/demo/ui/action_menu.tscn
@@ -2,7 +2,7 @@
 
 [ext_resource path="res://globals/action_menu.gd" type="Script" id=1]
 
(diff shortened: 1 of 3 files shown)


fleskesvor:
Seems to work nicely for me. Thanks, @mjtorn. 🙂

PR #99, opened 2018-03-26
Fix broken dialog choices.

StraToN:


fleskesvor:
Great job, @StraToN! 😄I just cherry picked this commit into my Daïza branch and tested, and it works like a charm.

I can also confirm that the `*.translation` files are auto-generated, so they should be removed from the commit and added to `.gitignore`.

device/demo/ui/dd_default.tscn
```diff
@@ -149,13 +151,17 @@ anchor_left = 0.0
...
+rect_clip_content = false
 mouse_filter = 1
 mouse_default_cursor_shape = 0
-size_flags_horizontal = 2
-size_flags_vertical = 2
+size_flags_horizontal = 3
+size_flags_vertical = 3
 custom_constants/separation = 10
 alignment = 0
+_sections_unfolded = [ "Size Flags" ]
```
fleskesvor: I think that maybe we should make a habit of closing sections before staging scene files for commits, and maybe add this to the contributing guidelines as well, since it adds noise to the diff. Not really a problem in this case though, since there are already the other diffs, like the node indices that Godot 3.0 insists on adding.

device/demo/ui/dd_default.tscn
```diff
@@ -198,6 +206,7 @@ toggle_mode = false
 enabled_focus_mode = 2
 shortcut = null
 group = null
+_sections_unfolded = [ "Anchor" ]
```
fleskesvor: This one should be collapsed too.

device/demo/ui/dd_default.tscn
```diff
@@ -211,17 +220,20 @@ margin_right = 1605.0
...
 custom_colors/font_color = Color( 1, 1, 1, 1 )
-custom_colors/font_color_shadow = Color( 1, 1, 1, 1 )
+custom_colors/font_color_shadow = Color( 0, 0, 0, 1 )
 text = "I should pick option one"
+valign = 1
 autowrap = true
 percent_visible = 1.0
 lines_skipped = 0
 max_lines_visible = -1
+_sections_unfolded = [ "Mouse", "Rect", "custom_colors", "custom_fonts" ]
```
fleskesvor: Here's another one.

StraToN:
@fleskesvor All your remarks were fixed :)

Issue #66, opened 2018-02-21
Support for counters in esc scripts?


mjtorn:
It would be useful to have counters in esc scripts. The feature I'm thinking of is when the player tries to do something many times, eg. open a locked door, after a couple of tries he could say "Maybe there's a key somewhere."

I'm not sure, but I think Thimbleweed Park had this kind of a mechanic.

The implications for code could be pretty huge and I'm not sure how to go about implementing it, but here are some ideas.

`incr_global tried_door` would bump the value by one, starting with 1 if `tried_door` is unset.

`decr_global tried_door` would decrement the value. What do you feel about going to 0 or below?

`gt tried_door 3` style RPN notation would evaluate. `gt` for `greater than`. `eq` for `equals` (in case the door is kicked by `use` and takes three times to break in). Characters used in case there are problems using symbols like `>=` which may even confuse the parser into thinking there's a group coming.

The syntax could be like this
```
:use
say player "It's locked."
incr_global tried_door
say player "Maybe there's a key somewhere." [ge tried_door 3]
```

or

```
:use
> [eq kicked_door 3, !door_opened]
    say player "Finally!"
    anim door player_kicks_door_open
    set_global door_opened

> [!door_opened]
    say player "It's stuck."
    anim player player_kicks_door
    incr_global kicked_door
```

Writing these examples also served as an exercise for me, because I haven't done animations and am not sure if I got the syntax in general correct, so please let me know ;)

But the point is that counters have their uses - what do you think?

mjtorn:
Some other things came to mind.

Can inventory items be incremented, like finding up to 8 pieces of eight?

Should the counters be tied to a character instead of being global, in case Escoria starts supporting multiple player characters?

This could potentially cause interesting interactions where an NPC has 8 pieces of eight and the player does things to get all of them.

fleskesvor:
I think this is a good idea, and it's functionality we also talked about for The Farrier. For that game, we ended up just using multiple flags for now. One thing to keep in mind is that the save system currently only supports booleans, so that will also need to be improved, because otherwise you'd lose the current state when you quit the game.

With this functionality in place, incrementing inventory items should be trivial, since they can use an AnimatedSprite and have their state changed when you perform the action which increases the counter.

StraToN:
This is a very good idea. SCUMM language does manage counter like this : http://www.pagetable.com/?p=614

lavaduderDev:
[...]

PR #102, opened 2018-03-29
Support integer globals, including integer comparisons

fleskesvor:
The goal of this PR is to properly support non-boolean globals so that it will eventually be possible to add counters (#66) and write more advanced comparisons in esc scripts.

This will require thorough testing, since there are many parts of the framework which rely on the return value from `vm.get_global()` being a global.

I have altered the autosave functionality to write the actual value of globals to the save script and also added comparisons using `eq [global] [value]` and `ne [global] [value]`. The latter does not perform any sanity checks yet, and should at the very least attempt to collect a quoted string containing comma(s).

I also haven't looked into regular save games yet, to make sure these changes don't cause any regressions.

### DIFF
--- a/device/globals/esc_compile.gd
+++ b/device/globals/esc_compile.gd
@@ -1,6 +1,8 @@
 
 var commands = {
 	"set_global": { "min_args": 2, "types": [TYPE_STRING, TYPE_STRING] },
+	"dec_global": { "min_args": 2, "types": [TYPE_STRING, TYPE_INT] },
+	"inc_global": { "min_args": 2, "types": [TYPE_STRING, TYPE_INT] },
 	"set_globals": { "min_args": 2, "types": [TYPE_STRING, TYPE_BOOL] },
 	"debug": { "min_args": 1 },
 	"anim": { "min_args": 2, "types": [TYPE_STRING, TYPE_STRING, TYPE_BOOL, TYPE_BOOL, TYPE_BOOL] },
@@ -39,7 +41,6 @@ var commands = {
 	"queue_animation": { "min_args": 2, "types": [TYPE_STRING, TYPE_STRING, TYPE_BOOL] },
 	"game_over": { "min_args": 1, "types": [TYPE_BOOL] },
 }
-
 
 func check_command(cmd, state, errors):
 	if !(cmd.name in commands):
@@ -190,7 +191,7 @@ func trim(p_str):
 
 
 
-func parse_flags(p_flags, flags_list, if_true, if_false, if_inv, if_not_inv):
+func parse_flags(p_flags, flags_list, ifs):
 	var from = 1
 	while true:
 		var next = p_flags.find(",", from)
@@ -201,19 +202,27 @@ func parse_flags(p_flags, flags_list, if_true, if_false, if_inv, if_not_inv):
 			flag = p_flags.substr(from, next - from)
 		flag = trim(flag)
 		var list = []
+
 		if flag[0] == "!":
 			list.push_back(true)
-			flag = flag.substr(1, flag.length()-1)
-			if flag.find("inv-") == -1:
-				if_false.push_back(trim(flag))
+			flag = trim(flag.substr(1, flag.length()-1))
+			if flag.find("inv-") == 0:
+				ifs["if_not_inv"].push_back(trim(flag).substr(4, flag.length()-1))
+			elif flag.substr(0, 3) in ["eq ", "gt ", "lt "]:
+				var elems = flag.split(" ", true, 2)
+				var comparison = "ne" if elems[0] == "eq" else "le" if elems[0] == "gt" else "ge"
+				ifs["if_" + comparison].push_back([elems[1], elems[2]])
 			else:
-				if_not_inv.push_back(trim(flag).substr(4, flag.length()-1))
+				ifs["if_false"].push_back(trim(flag))
 		else:
 			list.push_back(false)
 			if flag.find("inv-") == 0:
-				if_inv.push_back(trim(flag).substr(4, flag.length()-1))
+				ifs["if_inv"].push_back(trim(flag).substr(4, flag.length()-1))
+			elif flag.substr(0, 3) in ["eq ", "gt ", "lt "]:
+				var elems = flag.split(" ", true, 2)
+				ifs["if_" + elems[0]].push_back([elems[1], elems[2]])
 			else:
-				if_true.push_back(trim(flag))
+				ifs["if_true"].push_back(trim(flag))
 		if flag.find(":") >= 0:
 			var pos = flag.substr(0, flag.find(":"))
 			var inv = flag.substr(0, pos)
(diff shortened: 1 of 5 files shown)


device/globals/esc_compile.gd
```diff
@@ -256,21 +262,12 @@ func read_dialog_option(state, level, errors):
...
 		var flag_list = []
-		parse_flags(q_flags, flag_list, if_true, if_false, if_inv, if_not_inv)
-		if if_true.size():
-			cmd.if_true = if_true
-		if if_false.size():
-			cmd.if_false = if_false
-		if if_inv.size():
-			cmd.if_inv = if_inv
-		if if_not_inv.size():
-			cmd.if_not_inv = if_inv
```
fleskesvor: I think this must have been a bug, but since the `inv-inv_item` and `!inv-inv_item` is not documented anywhere, as far as I know, it was probably never encountered.

device/globals/esc_compile.gd
```diff
@@ -327,14 +321,9 @@ func read_cmd(state, level, errors):
...
 		read_line(state)
 
-	if if_true.size():
-		cmd.if_true = if_true
-	if if_false.size():
-		cmd.if_false = if_false
-	if if_inv.size():
-		cmd.if_inv = if_inv
-	if if_not_inv.size():
-		cmd.if_not_inv = if_inv
```
fleskesvor: Same presumed bug.

fleskesvor:
This PR should also fix a presumed bug where if one used the `!inv-inv_item` syntax instead of `!i/inv_item`, to check if `inv_item` is not in the inventory, the check for `inv-inv_item` would be run instead.

device/globals/global_vm.gd
```diff
@@ -306,7 +316,12 @@ func sched_event(time, obj, event):
 	event_queue.push_back([time, obj, event])
 
 func get_global(name):
-	return (name in globals) && globals[name]
+	# If no value or boolean, return boolean for backwards compatibility
+	if not name in globals:
+		return false
+	if typeof(globals[name]) == TYPE_BOOL:
+		return !!globals[name]
```
fleskesvor: It's probably safer to treat all globals as strings going forward and compare the string to "true"+"false".

fleskesvor:
Added support for arithmetic operations in esc scripts using eg. `set_global my_counter +=1`, `set_global foo /=2`, etc. The second argument must match the regular expression `([-+*\/])=(\d+)`, so no spaces between `=` and the addend/multiplier/divisor/subtrahend.

mjtorn:
This looks pretty good!

One outstanding question: how does this discern between `1` and `"1"`?

I was thinking strings would need to have the double-quotes explicitly, so only integers and booleans would work without them, and trying to set something as `foo` would be an error.

Not sure we ever have to care about `"1"` anyway, but I couldn't find the code for integers so is that still underway?

Thanks anyway, good stuff :)

fleskesvor:
> One outstanding question: how does this discern between `1` and `"1"`?

It doesn't, all globals are strings, but cast to a boolean or number when convenient. I don't think this is an issue though, as long as we only use integers for numbers, since integers aren't subject to rounding errors, and equality checks will hold true whether they're actual integers or represented as strings. The number strings will only temporarily be cast to integers when context demand that they be numbers, like when using addition and subtraction, and doing number comparisons with lt, gt, le and ge.

fleskesvor:
This is pretty much done now. The only thing I can think of that might cause surprises is that comparison to a string containing a comma won't work, eg.:

`say player "Hello" [eq hello "Hello, World!"]`

I could fix this by extending the matched "flag" to include everything up until the second quotation mark, but I'm not sure if there are use cases for doing string comparisons anyway, and if so, should I also account for escaped quotation marks? Eg:

`say player "Hello" [eq hello "Hello, \"World\"!"]`

Anyway, I think this is ready for review now. I've tidied up my commits in a way that I think makes sense, but let me know if I should squash and/or rearrange anything.

device/globals/vm_level.gd
```diff
@@ -29,7 +32,11 @@ func _walk(params, block):
 ### commands
 
 func set_global(params):
-	vm.set_global(params[0], params[1])
+	var result = _diff_pattern.search(params[1])
```
mjtorn: Is this maybe a bit generic with `_diff_pattern` and `result`?

```
var calc_operation = _calc_pattern.search(params[1])
if calc_operation:
    vm.set_global_calculate(params[0], calc_operation.get_string(1), calc_operation.get_string(2))
```

or somesuch variable names would be more obvious.

fleskesvor: Yes, you're right. I will change the names to your suggestions.

mjtorn:
@fleskesvor I think beside the equality/comparison checking, the documentation should describe how to add or subtract.

Would you be open to a syntax like `incr_global i/pieces_of_eight 4` where the default to increment by would be `1` in the case of `incr_global i/pieces_of_eight` or just forcing it to always have an argument?

Not sure I got this right, but now it's `set_global i/pieces_of_eight +=4` without the space, right? That feels a bit different from the rest of ESC, and forbidding the space after the operator in ` += 4` feels a bit exotic in general.

Still loving the work, though!

fleskesvor:
> I think beside the equality/comparison checking, the documentation should describe how to add or subtract.

Yes, definitely.

> Would you be open to a syntax like `incr_global i/pieces_of_eight 4` where the default to increment by would be 1 in the case of `incr_global i/pieces_of_eight` or just forcing it to always have an argument?

Yes, I'm open to that, and I think it might actually be a better idea to use descriptive esc commands instead of adding "side effects" to `set_global`, even though it makes the language more verbose. Adding `incr_global` and `decr_global` (or similarly named commands) to the language would also remove the need for regular expressions or splitting strings to get at the value to add/subtract, so that's an added benefit. I think it would make sense to force the developer to use an integer as the second argument in that case though.

Thanks for the feedback, @mjtorn. I think I've also pretty much decided that putting in the work to improve string comparisons is not worth it for now, but maybe I'll write up what could be done in a feature request and leave that one for later.

mjtorn:
@fleskesvor sounds like a good plan from start to finish :) Hopefully moving to `inc_global` and `dec_global` isn't too big a hassle, and good thing to drop the r from the names!

Looking forward to testing this, there's a use case for me in my prototype game for counters :)

fleskesvor:
I've implemented the requested changes and documented the feature in `doc/esc_reference.md`.

device/globals/esc_compile.gd
```diff
@@ -205,15 +207,18 @@ func parse_flags(p_flags, flags_list, if_true, if_false, if_inv, if_not_inv):
 			list.push_back(true)
 			flag = flag.substr(1, flag.length()-1)
 			if flag.find("inv-") == -1:
-				if_false.push_back(trim(flag))
+				ifs["if_false"].push_back(trim(flag))
 			else:
-				if_not_inv.push_back(trim(flag).substr(4, flag.length()-1))
+				ifs["if_not_inv"].push_back(trim(flag).substr(4, flag.length()-1))
+		if flag.substr(0, 3) in ["eq ", "ne ", "gt ", "ge ", "lt ", "le "]:
```
fleskesvor: This should be an `elif`; otherwise negated conditions will be evaluated by the `else` block.

fleskesvor: Another thing: Should it be possible to negate these new expressions? Possibly, and if so I should probably create a helper function for populating the lists.

fleskesvor:
I fixed the bug I introduced yesterday and added support for negating the new comparison operators. With that, having six separate operators became redundant, so I removed `ne`, `le` and `ge`. I'm now also trimming the expression after removing the negation operator (`!`), so it's now safe to use spaces between negation and expression, and that allowed me to reorder the order of the conditions within the block which parses negations to be consistent with the block for un-negated expressions.

This will require thorough testing though, since it touches on a core feature of Escoria. I have tested the new feature extensively, but have not tested properly for possible regressions yet, and would appreciate help with testing. The easiest way to test is to make an autosave with an item already in the inventory and one or two integer globals set, which can be decremented/incremented by interacting with an item in the demo scene. Testing should also include expressions with multiple conditions.

fleskesvor:
I have rebased on master, fixed the formatting on my additions to `doc/esc_reference.md` and squashed the commits I felt it made sense to squash. I have also done pretty extensive testing with combinations of boolean globals, inventory items and integer comparisons with variations with negations of all of the above, and it seems to be working as expected for me now. I believe this PR is ready to be merged now, and will do so before the weekend if there are no objections.

fleskesvor:
@mjtorn I made some of the commits from my work Mac, so it snuck in. :confused: I've removed it from the commit and added it to .gitignore in a separate commit. Merging this now that you have also tested it.

PR #105, opened 2018-04-05
Add more markdown ticks to esc reference

mjtorn:
This makes the commands and arguments stick out better.


fleskesvor:
This is a nice readability improvement. :slightly_smiling_face: Thanks, @mjtorn!

PR #106, opened 2018-04-06
Terminate after reporting errors

mjtorn:
Escoria suffers from errors that can leave the game in an inconsistent
state and are easily missed in the output, which is sometimes very
verbose.

Without exceptions to throw, this hackish solution is (apparently) the
only way to make GDScript realize something is wrong and terminate.


fleskesvor:
This is very useful. :+1: Thanks!

PR #108, opened 2018-04-06
Refactoring, cleanups and fixing around translations and dialog

mjtorn:
This is probably a lot to take in, but I hope it gets merged. I tried a ton of different language configurations and typewriter text and everything and it never broke. Too bad there is no unit testing for this.

I tried very hard not to break any existing APIs, but still move Escoria more in the direction of a framework when it comes to settings.

This does not solve world hunger, so three things still come to mind that need work:

  1. Maybe dialog needs to be a container for clamping etc, but then the size problems need to be solved
  1. Implement something like `tooltip_lang_default` for dialog text
  1. Fix `lockit.gd`

Those issues are solvable in a separate branch and better done once this cleanup gets approved.


mjtorn:
If it's not a huge (t)ask, it would be very sweet if commits could be cherry-picked from this PR if some are more ok than others, it would make further cleanups easier thanks to fewer commits in squash-rebasing.

device/globals/dialog_instance.gd
```diff
@@ -207,17 +205,14 @@ func anim_finished(anim_name):
...
 
-	finish_with_speech = vm.settings.skip_dialog
-
-	#speech_language = TranslationServer.get_locale().substr(0, 2)
-	speech_language = vm.settings.voice_lang
-	if !(speech_language in speech_locales):
-		speech_language = default_speech_language
+	# Ensure a supported speech locale has been set
+	var speech_locales_path = load(ProjectSettings.get_setting("escoria/application/speech_locales_path")).new()
+	assert(vm.settings.voice_lang in speech_locales_path.speech_locales)
```
fleskesvor: Maybe this could use `vm.report_errors()` as well?

device/project.godot
```diff
@@ -31,7 +31,12 @@ application/max_voice_volume=2
 application/save_data="res://globals/save_data.gd"
 application/speech_suffix=".spx"
 application/tooltip_lang_default="en"
+application/text_lang="en"
```
fleskesvor: Why is `tooltip_lang_default` and `text_lang` different settings? Are there any real use cases where a developer would want different language settings for them, or could `text_lang` be used for tooltips as well?

fleskesvor: Nevermind; I just saw the explanation in the documentation below.

device/globals/dialog_instance.gd
mjtorn: It would generally cause the same effect, of the game crashing, but I figured that this is a case where the game/code would be in such a horribly inconsistent state that it's better to let the programmer be told explicitly.

The use case for disabling crashes on error reporting might be collecting errors in broad sweeps instead of fixing them one by one, or something similar.

I can't think of a strong use case for having an unsupported locale - something that should be a pivotal part of the game design - be a benign thing.

Having argued that case, it's a trivial fix so if you want that to happen, I'll make it happen :D

device/globals/dialog_instance.gd
```diff
@@ -79,17 +77,17 @@ func init(p_params, p_context, p_intro, p_outro):
...
-		var ptext = TranslationServer.translate(tid)
-		if ptext != tid:
+		var ptext = TranslationServer.translate(text_id)
+		if ptext != text_id:
 			text = ptext
-		elif force_ids:
-			text = tid + " (" + text + ")"
-		text_id = tid
+		else:
+			text = "(NOT TRANSLATED)\n\n" + text
```
fleskesvor: Shouldn't this also use `vm.report_errors()´, similar to if text_id is not provided on line 90? For the cowboy devs, I think the dialogue should appear similarly in the two cases, eg. `"(NO TRANSLATION)\n\n" + text` and `"(NO TEXT_ID)\n\n" + text` OR `"(NO TRANSLATION) " + text` and `"(NO TEXT_ID)" + text`

device/demo/speech_locales.gd
```diff
@@ -0,0 +1,6 @@
+# Define your game's speech/voice locales within the game
+# because ProjectSettings doesn't deal well with arrays.
+# These can be used in your game to change the spoken language.
```
fleskesvor: Indeed. I believe Godot supports `PoolStringArray`s internally, but they don't seem to be editable from the project settings menu, so this seems like a good solution.

fleskesvor:
Could you check if the locale settings under https://github.com/mjtorn/escoria/blob/a058642d17b3e9d2bc831f837ea79809d03a9079/device/project.godot#L96 are used for anything and, if not, remove them? Maybe as an edit to 
718882e, since it touches on translation settings.

    [locale]

    translations=PoolStringArray( "res://translations/general.en.translation", "res://translations/general.es.translation", "res://translations/general.fr.translation" )

It's been there for a while, and that directory doesn't even exist.

fleskesvor:
> Implement something like tooltip_lang_default for dialog text

This is solved by this PR, right?

> Fix lockit.gd

If this isn't something that is solved by this PR, could you create an issue explaining what the issue is?

device/globals/dialog_instance.gd
fleskesvor: Yeah, I think you're probably right. You should check if speech is enabled though, because if it's not, the developer might not have any speech_locales set, and it shouldn't cause an assertion error in that case.

device/globals/dialog_instance.gd
fleskesvor: Ok, I agree with your reasoning. 👍

fleskesvor:
Ok, those locale settings can be fixed in a separate commit then.

mjtorn:
I said this on IRC too, but when I force-pushed the change to asserting the speech locale, I included updating the translations path in the project settings. That enabled me to see some French in the demo scene :D

fleskesvor:
I just tested it again myself, and this is a major improvement. Thanks, @mjtorn! 😄

PR #110, opened 2018-04-08
Rename and expand tooltip_lang_default to development_lang

mjtorn:
A quick new config option I made. Tested with the demo scene. Works as expected.

I put this under `platform` and moved `tooltip_lang_default` there as well
for reasons explained in the commit messages.

Everything should be safe to fly because when `lockit.gd` is fixed, these
texts will be the base of translations anyway, so there shouldn't be a
discrepancy unless the game developer does strange things with the translation CSV.


mjtorn:
When this gets merged, #45 needs to be updated with a comment that The Book(tm) needs an update for the path to the tooltip lang setting.

fleskesvor:
Now that you've implemented the same "short circuit" for dialog text, would it make sense to use the same project setting for both and call it `development_lang` or something like that?

fleskesvor:
Yeah, I'm hoping to dedicate some time to updating the Floss Manuals tutorial before the Adventure Jam in case anyone wants to use Escoria for that.

Looks like this should be good to go once you've made that change (and updated your documentation). :slight_smile:

fleskesvor:
Awesome! 👍 Thanks for the contribution!

PR #114, opened 2018-04-09
Change of UI workflow

fleskesvor:
This is an attempted start at the refactorings discussed in #113.

I have moved most scenes from `device/ui/` to `device/demo/ui/` to ensure that they work with the demo scene. The dialogue `ResourcePreloader`s were left behind because they're not so tightly coupled to a project, but it might make sense to move them as well, to make sure they are not altered by updating the framework.

To support the discussed modularity, I removed the code used to set inventory and action menu based on project settings, and have instead created `action_hud.tscn` which is similar to `hud.tscn`, and instanced by `game_am.tscn`, which should then be used as a drop-in replacement for `game.tscn` in rooms where a an action menu should be used in place of the verb menu. The verb menu still relies on having the correct project setting for the hud, which isn't ideal, but I'm not sure how to best solve that yet.

When everything is working as intended, unused (but useable) UI scenes should be moved into `templates/`.


device/globals/game.gd
```diff
@@ -373,28 +373,16 @@ func set_camera_limits():
...
-	else:
-		inventory = get_node("hud_layer/hud/inventory")
-
-	# Add action menu to hud layer if found in project settings
-	if ProjectSettings.get_setting("escoria/ui/action_menu"):
-		action_menu = load(ProjectSettings.get_setting("escoria/ui/action_menu")).instance()
-		if action_menu and action_menu is preload("res://globals/action_menu.gd"):
-			$hud_layer.add_child(action_menu)
+	if get_node("hud_layer/hud") is InstancePlaceholder:
+		get_node("hud_layer/hud").replace_by_instance(hres)
```
fleskesvor: The hud has already been instanced in the `game_am.tscn`. I will check to see why that might be.

fleskesvor:
I've separated the UI into three sets of templates, which can be used by instancing the appropriate game scene into a room:

- `game_vm.tscn` contains `hud_vm.tscn`, which is the old `game.tscn` and `hud.tscn` for the verb menu + inventory used in the demo room
- `game_am.tscn` contains `hud_am.tscn`, which has an action menu and inventory
- `game.tscn` contains `hud.tscn`, which used to be `hud_minimal.tscn` and is appropriate for left-click interfaces where the left-click action is set on each item separately

The unused templates have been moved into `templates/ui`, so the easiest way to eg. use the action menu is to copy `game_am.tscn`, `hud_am.tscn` and `action_menu.tscn` into `device/demo/ui` and use the editor to move them from there to resolve resource paths. Additionally, `device/globals/scenes_cache.gd` must reference a valid game scene template and the project settings must reference a valid HUD scene.

This is rather awkward at the time being, but it's the sort of thing that should be easy to handle with an editor plugin. I'm not sure if there's a better solution at the moment.

The `device/ui` directory now only contains the dialog resource preloaders and fonts, which are all things that can be used for multiple projects if wanted.

I think this PR is at a point now that it would be helpful with some feedback from others.

fleskesvor:
Having thought about it, I think it would be much better to focus on making a proper editor plugin, so I'm closing this.

PR #115, opened 2018-04-10
Trivial UI fixes

mjtorn:
Prevent the tooltip text from showing up in the wrong place when mouse enters an item.

Close action menu and inventory (if it's implied to be supposed to be toggled by the presence of an action menu) when hitting escape to open the overlay menu.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -29,6 +29,12 @@ func set_mode(p_mode):
 func mouse_enter(obj):
 	var text
 	var tt = obj.get_tooltip()
+
+	# When following the mouse, prevent text from flashing for a moment in the wrong place
+	if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
+		var pos = get_viewport().get_mouse_position()
+		tooltip.set_position(pos)
+
 	# We must hide all non-inventory tooltips and interactions when the inventory is open
 	if action_menu and inventory.is_visible():
 		if obj.inventory:
@@ -58,6 +64,7 @@ func mouse_enter(obj):
 				text = text.replace("%1", tr(tt))
 		else:
 			text = tt
+
 		get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", text)
 		vm.hover_begin(obj)
 
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -264,6 +264,15 @@ func scene_input(event):
 
 
 	if event.is_action("menu_request") && event.is_pressed() && !event.is_echo():
+		# Do not display overlay menu with action menu or inventory, it looks silly and weird
+		if action_menu:
+			if action_menu.is_visible():
+				action_menu.stop()
+
+			# Hide inventory only when it's implicitly not always visible
+			if inventory and inventory.is_visible():
+				inventory.close()
+
 		if vm.can_save() && vm.can_interact() && vm.menu_enabled():
 			main.load_menu(ProjectSettings.get_setting("escoria/ui/main_menu"))
 		else:


device/globals/game.gd
```diff
@@ -29,6 +29,12 @@ func set_mode(p_mode):
 func mouse_enter(obj):
 	var text
 	var tt = obj.get_tooltip()
+
+	# When following the mouse, prevent text from flashing for a moment in the wrong place
+	if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
+		var pos = get_viewport().get_mouse_position()
+		tooltip.set_position(pos)
```
fleskesvor: Does this make ` func _input(ev)` on line 336 redundant?

mjtorn: No, because that's what actually tracks the mouse position and updates the tooltip location, as every movement of the mouse is an input event.

fleskesvor: Is this logic only triggered when the mouse enters the object then? If so, wouldn't it also be possible to remove the `vm.hover_object` condition from `func _input(ev)` to have the tooltip always track the mouse, even when it's not hovering over an object?

mjtorn: I tested this, and also a bug came up. Removing the `vm.hover_object` condition leads to a crash on startup, because `tooltip` is not set. This is something like a race condition of sorts that would be more painful to hunt down, to initialize `tooltip` elsewhere.

So I think it should be as it is.

device/globals/game.gd
```diff
@@ -257,6 +264,15 @@ func scene_input(event):
...
 
 	if event.is_action("menu_request") && event.is_pressed() && !event.is_echo():
+		# Do not display overlay menu with action menu or inventory, it looks silly and weird
+		if action_menu:
+			if action_menu.is_visible():
+				action_menu.stop()
+
+			# Hide inventory only when it's implicitly not always visible
+			if inventory and inventory.is_visible():
+				inventory.close()
```
fleskesvor: I'm not too happy about the implementation of the inventory being so closely tied to the action menu. What if a (to be implemented) two-click interface is being used, instead of a verb or action menu, and we want the inventory to be closed when the menu is opened or if an action menu is used and the developer doesn't want an auto-closeable inventory? Would it be better to have an exported `auto_closeable` variable in `inventory` instead?

I don't remember if we already make similar assumptions elsewhere. If so, this has to be revisited anyway, so I'm fine with it.

device/globals/game.gd
fleskesvor: Ok, I understand, and I think this is a good workaround for now. :slightly_smiling_face:

device/globals/game.gd
mjtorn: I don't really like this hack either, though https://github.com/godotengine/escoria/blob/master/device/globals/game.gd#L33 is a similar thing.

But I think I see where you're going with this now that I looked at the code.

Because `load_hud()` looks at the settings for an inventory already, and injects it into the scene tree if it's not there (like it would be in old SCUMM and Daiza), the default for `auto_closeable` would be `false` for Daiza and the demo scene but I'd set it to `true`.

I might opt for a more verbose name like `hidden_by_default`.

Did I understand correctly?

fleskesvor: > Because `load_hud()` looks at the settings for an inventory already, and injects it into the scene tree if it's not there (like it would be in old SCUMM and Daiza), the default for `auto_closeable` would be `false` for Daiza and the demo scene but I'd set it to `true`.

It does, but I removed that code in #114. Since there's already code like this present, I think it's fine for this PR though, and I can change that in my PR instead. Yes, I agree that `hidden_by_default` is a better name.

mjtorn: Nice, thanks!

mjtorn: But forgot to ask, do you want the name and stuff in this PR still?

fleskesvor: No, just leave it like this, and I'll merge it once I've tested it.

fleskesvor:
When using the demo scene with the verb menu and inventory, the entire hud is actually hidden when the menu is opened with the `menu_opened()` function in `hud.gd`. Maybe it would be best to figure out why this doesn't happen when the action menu is used, because I think this is the desired behaviour, regardless of input scheme.

mjtorn:
This may be a matter of preference, but I find it weird that the verb menu and inventory get hidden :D 

Still I think it might be best to focus on making games by some "ground rules", like "if you make a game with Escoria that uses a verb menu and an inventory, they will get hidden when you open the in-game menu unless the in-game menu covers the entire screen". This at least means stability and then we can focus better on the other issues.

And yeah, it is weird that we got the menu and the actin menu at the same time, I don't get it.

fleskesvor:
> This may be a matter of preference, but I find it weird that the verb menu and inventory get hidden :D

Maybe it should be configurable in the future then.

I will merge this as-is and ask you to check if my PR works with your game once I'm done with it. Thanks for the contribution! :slightly_smiling_face:

PR #117, opened 2018-04-13
Set camera limits properly when using background_area

mjtorn:
Walking around in this area is horribly broken because `background_area.gd` uses viewport coordinates instead of "proper" coordinates like `TextureRect` and `background.gd` uses.

But at least this fixes the camera placement.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -356,12 +356,16 @@ func set_camera_limits():
 		var area = Rect2()
 		for i in range(0, p.get_child_count()):
 			var c = p.get_child(i)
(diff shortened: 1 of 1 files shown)


fleskesvor:
> Walking around in this area is horribly broken because `background_area.gd` uses viewport coordinates instead of "proper" coordinates like `TextureRect` and `background.gd` uses.

Can you describe in what way it's broken? `background_area.gd` has been tested extensively with the demo scene, but I think the issue you're describing is present when the background is larger than the viewport, is that right?

PR #118, opened 2018-04-13
Restore `var vm` in `player.gd`

mjtorn:
Fix this problem that suddenly revealed itself.

```
SCRIPT ERROR: GDScript::reload: Compile Error: Identifier not found: vm
   At: res://globals/player.gd:57.
ERROR: reload: Method/Function Failed, returning: ERR_COMPILATION_FAILED
   At: modules/gdscript/gdscript.cpp:598.
```

It appears to have been introduced in 7336e54bd30ca1b696085e2dc5fdb7e78e6c4d19
but I haven't encountered this bug, so there may be a secondary cause.

### DIFF
--- a/device/globals/player.gd
+++ b/device/globals/player.gd
@@ -5,6 +5,7 @@ extends Node2D
 var task
 var walk_destination
 var animation
+var vm  # A tool script cannot refer to singletons in Godot
 var terrain
 var walk_path
 var walk_context
@@ -316,6 +317,7 @@ func _ready():
 		return
 
 	animation = get_node("animation")
+	vm = $"/root/vm"
 	vm.register_object("player", self)
 	#_update_terrain();
 	if has_node("animation"):


mjtorn:
https://github.com/godotengine/godot/issues/4236

mjtorn:
@fleskesvor you're absolutely right about the new accessor syntax, though unfortunately `onready var` complains that it can't find the node. I'm shocked about this, because I force-pushed broken code, so I'll get back to you ASAP.

mjtorn:
When reloading the script, I get this `SCRIPT ERROR: GDScript::reload: Parse Error: Use 'onready var vm = get_node(..)' instead`

So we're stuck with my most recent force-push, which is pretty much syntactical updating.

fleskesvor:
Does the old syntax work with onready in a tool script though? If it does, it might be better to use the old syntax (`get_node("/root/vm")`) and have it be a oneliner at the top, rather than using the new syntax and splitting declaration and assignment.

mjtorn:
No, it still doesn't find the node.

`onready var vm = get_node("/root/vm")`

```
ERROR: get_node: Node not found: /root/vm
   At: scene/main/node.cpp:1524.
```

fleskesvor:
Ok, then this is as good as it gets for now. 🙂 Thanks, @mjtorn!

Issue #98, opened 2018-03-24
Action menu prevents walking


fleskesvor:
When setting an action menu from the project settings, clicking the background to walk does not work. This functionality was introduced with #79 and adapted into #80, so a git bisect to pin down this issue should start from 895a5337a2586b21057d74c05bf6fece9ab7032b^.

mjtorn:
On Fri, Mar 23, 2018 at 11:42:23PM -0700, Eivind wrote:
>When setting an action menu from the project settings, cicking the background to walk does not work. This functionality was introduced with #79 and adapted into #80, so a git bisect to pin down this issue should start from 895a5337a2586b21057d74c05bf6fece9ab7032b^.

Is this a problem?

Mentioned on IRC that having messed about extensively (exclusively) with
action menus while prototyping, this feels like the natural way of things.

One problem that would have to be solved if this was altered is clamping.

Say the player opens the menu at the left edge of the room on `shrubbery`
and walks 9000 pixels to the right.

Would the action menu just float along at the edge of the screen in a long
room like that?

Would the action menu automatically close? But only when it would be
scrolled out?

My recommendation is stating this in documentation as a feature and having
a `ProjectSettings` toggle for the behavior if someone wants to write
a patch.

PS.

Been a while since I played the LucasArts games with these, but I think
they required you to push the RMB down and choose an action by releasing
over a button. I wouldn't seek to emulate that behavior because it offers
little extra value and is probably harder to code up.

fleskesvor:
> Is this a problem?

It is if you have a scrolling room with interactive elements outside of the viewport, because then there's no way for you to walk to get to them. I also don't think it's intuitive that clicking on the ground doesn't make the player move. In Full Throttle, which has an action menu/verb coin, you can click on the ground to move. I don't remember other specific examples, but I think this is the norm.

> Say the player opens the menu at the left edge of the room on `shrubbery`
and walks 9000 pixels to the right.

I don't think this will be a problem, because if the action menu is open somewhere, and you want to move in the opposite direction, you'd have to click on the ground where you want to walk, which would close the menu anyway, right? Unless walking was triggered from an esc script, in which case we'd have to write code for closing the action menu, if that isn't implemented already.

> Been a while since I played the LucasArts games with these, but I think
they required you to push the RMB down and choose an action by releasing
over a button. I wouldn't seek to emulate that behavior because it offers
little extra value and is probably harder to code up.

[...]

PR #120, opened 2018-04-16
Stop player from walking when opening action menu

mjtorn:
Investigating if #98 is still relevant (it is not, it should be closed), I noticed the player can walk while the action menu is opened.

This is problematic if the walking causes the background to scroll, and the fix is trivial.

Please merge!


fleskesvor:
Looks good; thanks @mjtorn! :slightly_smiling_face:

PR #122, opened 2018-04-16
Fix tooltip mouse position by centering it

mjtorn:
This should have been in fcca54c7c0a404d0d7e1f765c5270e3b51e9b569
but the test case happened to line the tooltip properly, when
another later test case showed that it does have to be set by center.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -33,6 +33,7 @@ func mouse_enter(obj):
 	# When following the mouse, prevent text from flashing for a moment in the wrong place
 	if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
 		var pos = get_viewport().get_mouse_position()
+		pos -= tooltip.get_size() / Vector2(2, 1)
 		tooltip.set_position(pos)
 
 	# We must hide all non-inventory tooltips and interactions when the inventory is open


device/globals/game.gd
```diff
@@ -33,7 +33,7 @@ func mouse_enter(obj):
 	# When following the mouse, prevent text from flashing for a moment in the wrong place
 	if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
 		var pos = get_viewport().get_mouse_position()
-		tooltip.set_position(pos)
+		tooltip.set_position(pos - (tooltip.get_size() / Vector2(2, 1)))
```
fleskesvor: Looks good, but could it be written as `tooltip.position pos - tooltip.get_size() / Vector2(2, 1)` to be a bit easier to parse mentally?

mjtorn: Afraid not. `"Invalid set index 'position'`

Sometimes I don't understand when the setter function needs to be called and when assignments work.

mjtorn: That actually made it harder for me to read, though I know it has precedence, so I put the subtraction on the line above and made it simply set the new position. Hope that's fine ;)

fleskesvor:
Looks good to me. Thanks, @mjtorn! 🙂

PR #123, opened 2018-04-17
Make collapsible inventory configurable

fleskesvor:
Exports a boolean `is_collapsible` on the inventory and uses `false` as default to make behaviour least surprising (even though many games will probably have this set to `true`).

Maybe a collapsible inventory should be opened when an item is picked up as well. It's possible to open the inventory from an esc script, but it is inconvenient to couple all pick_ups with explicit opening of the inventory.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -125,11 +125,12 @@ func clicked(obj, pos, input_event = null):
 			#click_anim.play("click")
 			if player == self:
 				return
-			if (action_menu and !inventory.is_visible()) or !action_menu:
-				player.walk_to(pos, walk_context)
-				# Leave the tooltip if the player is in eg. a "use key with" state
-				if !current_action:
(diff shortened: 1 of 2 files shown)


StraToN:
Is this PR supposed to fix #98 ?

fleskesvor:
Yes, in my opinion this is a reasonable fix for #98.

StraToN:
To illustrate easier, I'd integrate a button in the UI to show/hide the inventory since TAB is not straightforward.

fleskesvor:
Yeah, I think that's a good idea as well, and make it configurable whether it should trigger on click or mouse_entered. I think the most frequently used methods are either a small button in a corner or a mouse aware edge of the screen. This would allow both.

mjtorn:
I gave this a quick test run and couldn't find anything disturbing.

To echo what I said on IRC, tab has been used before in games, but currently there's no documentation for UI element interaction nor even that the shortcuts exist.

I opened #125 so they're not forgotten and this can be merged and maybe the UI examples and documentation could come in another PR to not keep this stuff waiting.

PR #124, opened 2018-04-17
Exit/sensor node [DON'T MERGE]

fleskesvor:
This branch is currently a mess of things and should not be merged.

Since exit/sensor nodes are very similar to items, I think they should inherit a common base, but my feeling is that they're different enough from player that it doesn't make sense to add everything to `interactive.gd`, so that's why the hierarchy looks like this:

    interactive
    ├── hotspot
    │   ├── exit
    │   └── item
    └── player

Hotspots are interactive elements that might eg. have tooltips and hotspot highlighting; attributes that might not make sense for the player character.

I've tried to support both `Control` and `Area2D` nodes for exits, but it turns out, not surprisingly, that `Control` nodes are not well suited for this task. The advantage of `Control` nodes is that they work with `TextureRect` backgrounds and don't require the player character to have a "physical presence", ie. by being a `KinematicPlayer2D` with a collision shape. The con is that hacks are required to make the player move to a `Control` exit node, and it's not possible to check if a player "passes through" without looking up the positions in the navigation polygon and checking if either is within the node.

`Area2D`s seem almost perfect for this task. They can be any shape, multiple shapes per exit and can easily trigger both on player enter and exit events. Their requirements are that the background must be a `Sprite` with a collision shape for catching input and the player character must be a `KinematicPlayer2D` with a shape. The exact shape is of little importance though, and since the player node will be at the feet of its `Sprite` anyway, one can just create a default shape and not change the transform or extents at all, and it will just work.

The way I see it, this might be another nail in the coffin for the `TextureRect` background, though we will need to fix some issues that have come to light before we can outright retire it.

**EDIT:** There are also a few outdated files in this branch and some test files that I will remove before it can be considered acceptable for merging.

### DIFF
--- a/device/demo/actors/player/player.tscn
+++ b/device/demo/actors/player/player.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=16 format=2]
+[gd_scene load_steps=17 format=2]
 
 [ext_resource path="res://globals/player.gd" type="Script" id=1]
 [ext_resource path="res://demo/actors/player/player_anims.gd" type="Script" id=2]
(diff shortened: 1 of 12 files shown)


fleskesvor:
Closed by #127

Issue #121, opened 2018-04-16
Walking broken for background_area.gd


mjtorn:
If the background is a `Sprite` with a texture of the size eg. 3100x1080 and the `background_area.gd` script attached, walking is broken.

Here are my forensics notes, gathered by adding print debugging to the relevant files. Looks like we need a mapping or other type of relationship between the viewport and the underlying sprite/texture/thing is required to move the player to the correct coordinates. Alternatively change the way coordinates work.

I have no idea about how to fix this in practice.

# Using background.gd

## First move a bit down
position	(0, 0)
event pos	(2009.369995, 1044)
var pos	(2009.369995, 1044)
viewport pos	(956, 1044)
get path 	(2013.369995, 895.833984)	(2009.369995, 1044)

## Then toward the right edge
position	(0, 0)
event pos	(2948.370117, 829)
var pos	(2948.370117, 829)
viewport pos	(1899, 829)
get path 	(2009.369995, 1044)	(2948.370117, 829)

# Using background_area.gd

  * Can only move once to the right because downward was moving to the left
  * Note that the event position matches the viewport position

position	(0, 0)
event pos	(1886, 803)
var pos	(1886, 803)
viewport pos	(1886, 803)
get path 	(2013.369995, 895.833984)	(1886, 803)

fleskesvor:
Thanks for the bug report, @mjtorn. I have some ideas on how to fix this, and I will investigate as soon as possible.

fleskesvor:
I made a pull request on your #127 branch (mjtorn/escoria#2), which should fix these issues, no advanced vector math required.

fleskesvor:
Closed by #127

PR #127, opened 2018-04-18
Hotspot and exit nodes

mjtorn:
Based on the splendid work by @fleskesvor on #124 and it's predecessor, I cooked this up.

I believe it ticks every box

  * [X] Areas that react when the player walks on them
  * [X] Exit to another room (Also `TextureRect` support)
  * [X] Double-click for quick move to enable a modern feel
  * [X] Showcase the scenarios in the test room
  * [X] `exit.gd` is so simple that it's a good foundation if anyone comes up with a new use case
  * [X] Documentation

It also enforces the player character to be a  `KinematicBody2D`, which was discussed on IRC to be an acceptable change, because certain external documentation is for 2.1 anyway.

Thoughts?

PS.
In the spirit of full disclosure, my https://github.com/mjtorn/escoria/tree/devel does not really work with this, but this may be due to overlapping `Control` (in this case `TextureRect`) nodes and these should be deprecated as soon as #121 is closed. When it is closed, I propose we deprecate support for non-`Area2D` hotspots and exits because only one _good_ way to do things is better than implicitly tons of ways.

### DIFF
--- a/device/demo/actors/player/player.tscn
+++ b/device/demo/actors/player/player.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=16 format=2]
+[gd_scene load_steps=17 format=2]
 
 [ext_resource path="res://globals/player.gd" type="Script" id=1]
 [ext_resource path="res://demo/actors/player/player_anims.gd" type="Script" id=2]
@@ -131,8 +131,17 @@ tracks/0/keys = {
 "values": [ 5 ]
 }
 
-[node name="player" type="Node2D"]
+[sub_resource type="RectangleShape2D" id=8]
 
+custom_solver_bias = 0.0
+extents = Vector2( 10, 10 )
+
+[node name="player" type="KinematicBody2D"]
+
+input_pickable = false
+collision_layer = 1
+collision_mask = 1
+collision/safe_margin = 0.08
 script = ExtResource( 1 )
 speed = 300
 v_speed_damp = 1.0
@@ -163,5 +172,9 @@ anims/walk_down = SubResource( 5 )
 anims/walk_right = SubResource( 6 )
 anims/walk_up = SubResource( 7 )
 blend_times = [  ]
+
+[node name="CollisionShape2D" type="CollisionShape2D" parent="." index="2"]
+
+shape = SubResource( 8 )
 
 
--- a/device/globals/exit.gd
+++ b/device/globals/exit.gd
@@ -1,0 +1,22 @@
+extends "res://globals/trigger.gd"
+
+func stopped_at(pos):
+	# TextureRect exits run this code
+	if self is Control and get_global_rect().has_point(pos):
+		run_event("exit_scene")
+
+func body_entered(body):
+	# Entering an exit node runs the `:exit_scene` event
+	if body is preload("res://globals/player.gd"):
+		run_event("exit_scene")
+
+func body_exited(body):
+	# Do nothing when exiting an exit node. Usually this code should not be hit.
+	return
+
+func _ready():
+	add_to_group("exit")
+
+	if not "exit_scene" in event_table:
+		vm.report_errors("exit", [":exit_scene missing for " + self.name])
+
--- a/device/demo/rooms/test/exit1.esc
+++ b/device/demo/rooms/test/exit1.esc
@@ -1,0 +1,2 @@
+:exit_scene
+say player exit_first_exit:"I'm at the first exit"
--- a/device/demo/rooms/test/exit2.esc
+++ b/device/demo/rooms/test/exit2.esc
@@ -1,0 +1,2 @@
+:exit_scene
+say player exit_second_exit:"I'm at the second exit"
--- a/device/demo/rooms/test/hotspot1.esc
+++ b/device/demo/rooms/test/hotspot1.esc
@@ -1,0 +1,5 @@
+:enter
+say player enter_first_trigger:"I entered the first trigger"
+
+:exit
+say player trigger_first_trigger:"I exited the first trigger"
--- a/device/demo/rooms/test/test.tscn
+++ b/device/demo/rooms/test/test.tscn
@@ -1,17 +1,19 @@
-[gd_scene load_steps=14 format=2]
+[gd_scene load_steps=19 format=2]
 
 [ext_resource path="res://globals/scene.gd" type="Script" id=1]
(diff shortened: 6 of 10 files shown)


device/globals/hotspot.gd
```diff
@@ -0,0 +1,86 @@
...
+	var player = vm.get_object("player")
+	# Get mouse position, since event.position only works with Area2D exits
+	var pos = get_viewport().get_mouse_position()
+	printt("exit pos", pos)
+
+	if player and event.doubleclick:
+		player.set_position(pos)
+	elif player:
+		# Control blocks input to background, so make player walk
+		player.walk_to(pos)
```
fleskesvor: As mentioned on IRC, this is probably wrong if the background is larger than the viewport, ie. the player will walk to the wrong position.

device/demo/actors/player/player.tscn
```diff
@@ -164,4 +173,8 @@ anims/walk_right = SubResource( 6 )
 anims/walk_up = SubResource( 7 )
 blend_times = [  ]
 
+[node name="CollisionShape2D" type="CollisionShape2D" parent="." index="2"]
+
+shape = SubResource( 8 )
+
```
fleskesvor: These changes will require some documentation as well.

device/globals/exit.gd
```diff
@@ -0,0 +1,22 @@
...
+
+func body_exited(body):
+	# Do nothing when exiting an exit node. Usually this code should not be hit.
+	return
+
+func _ready():
+	add_to_group("exit")
+
+	if not "exit_scene" in event_table:
+		vm.report_errors("exit", [":exit_scene missing for " + self.name])
```
fleskesvor: :+1:

device/globals/hotspot.gd
fleskesvor: That doesn't have to be solved by this PR though, as long as there's an issue for it.

fleskesvor:
This looks very good. :smile: The only thing I'm not sure about, which I'm not able to test now, is whether the new hotspots make sense within the demo scene. Maybe the hotspots offer enough context as it is or maybe they should have some simple textures. What do you think?

device/globals/hotspot.gd
mjtorn: Breaks with `event.position` too, #121 is close enough for this issue, or what do you think?

It's probably symptoms of the same problem.

device/demo/actors/player/player.tscn
mjtorn: Added it, will force-push soon

mjtorn:
@fleskesvor pretty much nothing makes sense in the demo scene :D So my opinion is that it shouldn't block the merge, because there's a lot to polish and do there, which should all be post-jam anyway.

The relevant parts are clearly visible in the editor and they have tooltips that work even with tighter settings for translation, so maybe that's enough for now.

device/globals/hotspot.gd
fleskesvor: Yes, I agree.

fleskesvor:
> @fleskesvor pretty much nothing makes sense in the demo scene :D

Haha, good point. :smile:

doc/exits-and-hotspots.md
```diff
@@ -0,0 +1,58 @@
+# Hotspots and exits
+
+Hotspots in Escoria are thought of as "things a character can walk on", either
+like opening a door in Prince of Persia by stepping on a plate or a more abstract
+"To street" exit hotspot.
```
fleskesvor: Maybe it would be better to call it "trigger" to avoid confusing it with established terminology?

fleskesvor: So, `trigger.gd` as well.

doc/exits-and-hotspots.md
```diff
@@ -0,0 +1,58 @@
...
+
+The working cases are described here.
+
+Exit is extended from hotspot. To make gameplay quicker and "more modern", hotspot
+implements double-clicking, which therefore works in exit as well.
+
+## Concerning the player
+
+Your player character must be a `KinematicBody2D` in order to work with hotspots
+and exits.
```
fleskesvor: `TextureRect` exits should work with `Node2D` player characters though.

fleskesvor: But maybe it's best not to mention it, since the `KinematicBody2D` player is easy to make and has this clear advantage.

doc/exits-and-hotspots.md
mjtorn: Myeah, I dunno what the established terminology is. In Unity/Adventure Creator iirc pretty much anything was a hotspot and additional scripts defined what kind, which is not how Godot does things.

I should have time to edit the commits in the afternoon.

Do you prefer I preserve the wording, like "Triggers in Escoria are thought of" just in case someone is as unfamiliar with established terminology as I am?

fleskesvor: Yeah, I think that wording is fine. :slightly_smiling_face: Maybe we should make a list of terminology to establish our use of terminology throughout the framework. I know the use of "scene" vs. "room" can also be confusing to newcomers.

doc/exits-and-hotspots.md
```diff
@@ -0,0 +1,58 @@
...
+The `exit.gd` script must be attached to either the `Area2D` or `TextureRect`.
+
+Because exits are meant for changing scenes, the event keyword is `:exit_scene` to avoid
+confusion with hotspots.
+
+An example of this would be
+
+```
+:exit_scene
+set last_exit bridge
```
fleskesvor: This should be `set_global last_exit bridge`, I think.

doc/exits-and-hotspots.md
```diff
@@ -0,0 +1,58 @@
...
+collide with the hotspots and exits.
+
+## Hotspots
+
+Hotspots are `Area2D`, with the required `CollisionPolygon2D` child and optionally also
+a `Sprite`) child. You may name them freely.
+
+The `hotspot.gd` script must be attached to the `Area2D`.
+
+The hotspots react only on the player character.
```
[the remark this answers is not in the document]
mjtorn: I was making another point than you. I meant how an NPC will not (currently) trigger an event there.

I expanded on it now, though, to include what you said. I preferred the word "within" because included sounds a bit different.

Force-pushed and while at it, reworded the commit slightly for better `git blame`.

StraToN:
Merged. Thanks for your contribution !

PR #128, opened 2018-04-20
Allow not having a terrain in player.gd

mjtorn:
This is useful when editing the player character. Without checking
for terrain, because player.gd is a tool script, the editor will
complain ceaselessly about missing a terrain.

### DIFF
--- a/device/globals/player.gd
+++ b/device/globals/player.gd
@@ -211,6 +211,9 @@ func _editor_transform_changed():
 	_check_bounds()
 
 func _check_bounds():
+	if !terrain:
+		return
+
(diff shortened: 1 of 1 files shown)


StraToN:
Thanks a lot !
Makes me think that we should find a way to warn the user that a room *should* have a terrain node if it also instances a player scene.

fleskesvor:
We could use `report_errors()` in `global_vm.gd` for that. @mjtorn added an option to have the scene abort on errors if "escoria/platform/terminate_on_errors" is set, using an assertion.

mjtorn:
Yeah, that sounds like a good sanity check :)

Issue #111, opened 2018-04-08
When zooming, camera appears to be off


mjtorn:
To reproduce this, zoom in on a location in a scene. The camera will be off, though nothing in the coordinates or anything will appear to be off.

My theory is that the scaling is done from `(0, 0)` by sort of stretching the visible area in front of the camera, which is why the camera is off without really being off.

Does anyone know of a phenomenon like this or is my theory entirely off? Anything to fix it?

fleskesvor:
> To reproduce this, zoom in on a location in a scene.

I'm not sure what you mean by this. Neither of the zoom functions take a positional parameter, I think. The camera has its origin in the upper right corner, so unless your esc script also calls `camera_set_pos` you will just zoom closer and closer to the upper left corner of the original viewport. 

I agree that this behaviour isn't great though, and that maybe ideally you'd want to be able to call it with a global_id which could be used to anchor the position of the zoom.

fleskesvor:
@mjtorn Can you confirm that #130 gives the desired behavior?

fleskesvor:
> There's something that looks weird to my eye (or brain) but I think it's me expecting this to be centered without having a camera position where the characters are in the middle.

The characters in the above screenshots are not in the middle. Maybe that's the issue?

> There's also the question of the blurry background. My image is 1350px high, so I thought the camera would use the background image's pixels as they are available instead of scaling the visible ones.

Make sure both filter and mipmaps is unchecked in the import dock, then press reimport if you had to change either.

mjtorn:
> The characters in the above screenshots are not in the middle. Maybe that's the issue?

Yeah, like I said, it's the brain thinking they should be centered but they aren't. This happens despite the camera following the player (I think that's how the camera works now), so there probably is that drift thing when the camera doesn't need to move.

Problem is that `camera_set_target 0 helga` or `camera_set_target 0 helga player` doesn't center them either, so is there still the possibility of a bug in the position code?

I think a command for centering an object and/or coordinates with or without a speed, as best as possible considering the screen limits, would help here either way.

> Make sure both filter and mipmaps is unchecked in the import dock, then press reimport if you had to change either.

Ooh, thanks! There is some pixelation, but I believe that's natural, because I don't have an insane amount to work with. This should probably be documented in #45 at least or somewhere :)

fleskesvor:
[...]

PR #130, opened 2018-04-22
Fix game camera

fleskesvor:
Fixes a bug where game scene camera isn't being used, as far as I could tell. This should also make the camera clamp code redundant, since the camera will respect the bounds set initially.

Additionally, I added a second room to test the camera, including parallax and zooming, since the current small room isn't a good test case for these things.


fleskesvor:
Ideally the parallax layer should use a larger image to give us greater flexibility with the motion scale without risk of the camera panning outside its edges, but I think this is good enough to demonstrate the effect.

device/globals/global_vm.gd
```diff
@@ -695,7 +659,7 @@ func save():
 
 func set_camera(p_cam):
 	camera = p_cam
-	camera.clear_current()
```
fleskesvor: This caused the game scene camera to not be used, which I believe was a bug. Instead, the default viewport camera must have been used, which was probably the reason why extra clamping code was needed. The camera behaves much better with this camera actually being set.

device/globals/game.gd
```diff
@@ -357,29 +357,26 @@ func set_inventory_enabled(p_enabled):
...
+				var pos = child.get_global_position()
+				var size = child.get_texture().get_size()
 				area = area.expand(pos)
 				area = area.expand(pos + size)
 
 		camera.limit_left = area.position.x
 		camera.limit_right = area.position.x + area.size.x
-		var cam_top = area.position.y # - main.screen_ofs.y
-		camera.limit_top = cam_top
-		camera.limit_top = cam_top + area.size.y + main.screen_ofs.y * 2
```
fleskesvor: I'm not sure what the purpose of `main.screen_ofs` is, but this puts the camera outside of the background.

StraToN: ofs usually means "offset" so that makes sense :)

fleskesvor: Hmm, why would that have been desirable?

fleskesvor:
I updated the PR with documentation on implementing a parallax background.

It is also possible to add background items by replacing the automatic background shape with one that doesn't cover the whole background, but this requires changes to `globals/item.gd`, and the shape and placement are affected by positioning and scaling of the item shape, so I will detail this in an enhancement issue for future implementation.

device/globals/game.gd
StraToN: Maybe a dirty hack for Dog Mendoza? @punto-  ?

device/globals/game.gd
```diff
@@ -357,29 +357,26 @@ func set_inventory_enabled(p_enabled):
...
-	if camera_limits.size.x == 0 && camera_limits.size.y == 0:
-		var p = get_parent()
+	if camera_limits.size.x == 0 and camera_limits.size.y == 0:
 		var area = Rect2()
-		for i in range(0, p.get_child_count()):
-			var c = p.get_child(i)
-			if c is preload("res://globals/background.gd"):
-				var pos = c.get_global_position()
-				var size = c.get_size()
+		for child in get_parent().get_children():
```
StraToN: Way cleaner. I like it!

device/globals/game.gd
```diff
@@ -357,29 +357,26 @@ func set_inventory_enabled(p_enabled):
...
+			if child is preload("res://globals/background.gd"):
```
StraToN: This will cause #132 to need a rebase.

device/globals/game.gd
StraToN: Probably a dirty hack for Dog Mendoza. Let's ask @punto- , but I think this can go away.

device/globals/global_vm.gd
StraToN: make_current() seems clearer anyway as the name is more self-explanatory.

device/globals/game.gd
fleskesvor: If #132 is merged first, I can do the rebase.

device/globals/global_vm.gd
fleskesvor: Looking into this, it seems to be a bug introduced by ee1d6ac106f3b9eb9bca1a56ba0ff28ac55670bb. With a custom camera it makes sense that the Escoria camera would need to be cleared, but with that gone, this line should probably have been removed. I'll update the PR.

device/globals/global_vm.gd
```diff
@@ -161,56 +161,20 @@ func update_camera(time):
...
-		pos.x = camera_limits.position.x + half.x
-
-	if pos.y + half.y > camera_limits.position.y + camera_limits.size.y:
-		pos.y = (camera_limits.position.y + camera_limits.size.y) - half.y
-	if pos.y - half.y < camera_limits.position.y:
-		pos.y = camera_limits.position.y + half.y
-
-	return [orig_pos != pos, pos]
+	if camera and camera is Camera2D and prev_state:
+		camera.zoom = prev_state["zoom"]
```
fleskesvor: The type checks in the two functions above were added because I was experimenting with a camera as a child of the player node, i.e. one that would have been set by a game developer, but with that gone, it's probably not needed.

fleskesvor:
As mentioned on IRC, I've rebased on master, edited the first commit to remove the redundant call to camera.make_current() and edited the second commit to fix the reference to hotspot1.esc that went missing from #127 before it was merged.

This should also fix #111.

@StraToN: Do you want to wait and hear what @punto- has to say about the way camera limits are calculated or should this be merged as-is?

StraToN:
@fleskesvor I think I'll merge as-is, @punto- may respond later here or on the irc channel. If required, we'll make another PR to adapt the modification.

fleskesvor:
Yeah, that sounds like a good plan. We should probably also make a settings menu template for changing resolutions eventually, and then we may need to revisit this function as well:

https://github.com/godotengine/escoria/blob/master/device/globals/main.gd#L91

mjtorn:
```
:ready
> [!room.002.inited]
 camera_set_target 0 player
 wait 0.1
 camera_set_target 500 helga
 wait 1.5
 camera_set_target 500 player
 wait 1.5
 say player player_room002_sees_helga:"Ooh, Helga is by the bridge!"
 set_global room.002.inited true
```

The camera doesn't move to `helga` and back anymore :(

Is there something I got to adjust in the room?

fleskesvor:
> The camera doesn't move to helga and back anymore :(
>
> Is there something I got to adjust in the room?

No, you shouldn't have to adjust anything. If it was working before this commit, I must have broken something.

PR #131, opened 2018-04-23
Hide tooltip debug if there is no tooltip

mjtorn:
I got annoyed by the amount of noise when trying to work on Spine animations, which requires a fair amount of temporary `printt()` debugging.

This is one of the main causes of useless prints, so I suggest we make it more relevant by not printing something that can be clearly seen in the game.


StraToN:
Nothing wrong with this commit, I suppose.

PR #132, opened 2018-04-23
Created global types

StraToN:


mjtorn:
I think this is two or three commits in one, did you squash them or something?

  1. Types singleton
  1. That input stuff in the project settings
  1. The test room changes

Would it be possible for you to `git reset HEAD~1` and commit this in parts or restore the commits if you squashed them accidentally somewhere?

Otherwise really cool!

fleskesvor:
Looks good! :slightly_smiling_face: I can't verify what the project settings and hud changes entail from looking at the code though. I agree with @mjtorn that it would make sense to have this be three separate commits if that isn't too much work, as per the [contributing guidelines](https://github.com/godotengine/escoria/blob/master/CONTRIBUTING.md) (that I think I wrote while you were away).

I also think there's a bit of a conundrum with the types singleton: On one hand it makes sense that it's named `escoria_types` because functionally it defines a set of types, but on the other hand when used, we refer only to a single type, so the verbalization looks a bit odd. Would it be a good compromise to keep the script name, but map it as `escoria_type` in the project settings, or would that cause confusion? That is, keep the name `escoria_types.gd`, but refer to it in code as eg. `escoria_type.ACTION_MENU`.

mjtorn:
Come to think of it, I would shorten the name further, if possible. `esc_type.ACTION_MENU` to keep it simple.

If it was possible to put these in the global namespace with `TYPE_INT` (or whatever they're called) that would be even more convenient, but messy for other reasons ;)

fleskesvor:
Yes, `esc_type` is great. Then it's very clearly not a misspelling of the script name. Fewer letters to type is also a bonus. :+1:

StraToN:
Ok, I'll move the UI relooking away (it's very small anyway and could also go in a small demo remake) and perform your change requests asap.

fleskesvor:
The changes in the project settings seem to be incompatible with Godot v3.0.2 stable. I'm not able to open Escoria at all with your branch. Would it be best to split those changes out into a new PR and hold off on merging until official builds of v3.0.3 or v.3.1.0 are available? Opening the branch using the tip of Godot master is not a problem.

fleskesvor:
I'm at work and can't test this, but this looks completely safe to me. :+1:

Issue #10, opened 2017-03-14
Code cleaning


StraToN:
Many scripts included in Escoria contain lines of code that are commented out. We should clean them if they are not needed.

lavaduderDev:
There is also a lot of strange code. For example in telon_music.gd you'll see that it extends 
`"res://demo/objects/item.gd"` Where in the world is that? If it's referring to `"res://globals/item.gd"` Then that's really old code. Is this even need? If so, what for?

fleskesvor:
> There is also a lot of strange code. For example in telon_music.gd you'll see that it extends
"res://demo/objects/item.gd" Where in the world is that? If it's referring to "res://globals/item.gd" Then that's really old code. Is this even need? If so, what for?

`telon_music.gd` is actually not being referenced at all, and may be safely removed. Good catch, @lavaduderDev. 🙂

PR #133, opened 2018-04-23
Clean up clicked() and split out right_clicked()

mjtorn:
In hunting down the bug where opening an action menu runs the previous event (no GH issue for this at the moment) I noticed that walking around between opening the menu prevented the bug.

This lead me to investigate the code around clicking, which was insanely complex.

Here I'm attempting to clean it up. There may be more cleaning to be done, but this works in my action-menu-and-collapsible-inventory proto and the test room.

`git diff -w --cached` output in the commit message because it's easier to review than a regular diff because of all the whitespace noise.

```diff
diff --git a/device/globals/game.gd b/device/globals/game.gd
index f268f32..4f7708f 100644
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -99,7 +99,20 @@ func set_current_tool(p_tool):
 	current_tool = p_tool

 func clicked(obj, pos, input_event = null):
+	if !vm.can_interact():
+		return
+
+	if !player:
+		return
+
+	if mode != "default":
+		return
+
 	var walk_context = null
+	var action = obj.get_action()
+
+	# XXX: Why does this disable joystick_mode?
+	joystick_mode = false

 	if input_event:
 		walk_context = {"fast": input_event.doubleclick}
@@ -110,23 +123,17 @@ func clicked(obj, pos, input_event = null):
 			if area.has_method("is_clicked") and area.is_clicked():
 				return

-	joystick_mode = false
-	if !vm.can_interact():
-		return
-	if player == null:
-		player = self
-	if mode == "default":
-		var action = obj.get_action()
 	# Hide the action menu (where available) when performing actions, so it's not eg. open while walking
 	if action_menu:
 		action_menu.stop()
+
 	if action == "walk":
 		#click.set_position(pos)
 		#click_anim.play("click")
-			if player == self:
-				return
+
 		if inventory and inventory.is_collapsible:
 			inventory.close()
+
 		player.walk_to(pos, walk_context)
 		# Leave the tooltip if the player is in eg. a "use key with" state
 		if !current_action:
@@ -135,14 +142,6 @@ func clicked(obj, pos, input_event = null):
 	elif obj.inventory:
 		# Use and look are the only valid choices with an action menu
 		if action_menu:
-				# Do not set `look` as permanent action
-				if input_event.button_index == BUTTON_RIGHT:
-					interact([obj, "look"])
-					# XXX: Moving the mouse during `:look` will cause the tooltip to disappear
-					# so the following is a good-enough-for-now fix for it
-					get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", obj.get_tooltip())
-					vm.hover_begin(obj)
-				else:
 			current_action = "use"
 			# XXX: Setting an action here does not update the tooltip like `mouse_enter` does. Compensate.
 			var text = tr("use.id")
@@ -157,34 +156,53 @@ func clicked(obj, pos, input_event = null):
 		player.interact([obj, action, current_tool])
 	elif current_action != "":
 		# Walking the player to perform current_action is fine only when inventory is closed
-			if (action_menu and !inventory.is_visible()) or !action_menu:
+		if inventory:
+			if !inventory.is_collapsible or (inventory.is_collapsible and !inventory.is_visible()):
 				player.interact([obj, current_action, current_tool])
-		elif action_menu == null:
-
-			# same as action == "walk"
-			if player == self:
-				return
-			if (action_menu and !inventory.is_visible()) or !action_menu:
-				player.walk_to(pos, walk_context)
-				# Leave the tooltip if the player is in eg. a "use key with" state
-				if !current_action:
-					get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", "")
-
-		elif obj.use_action_menu && action_menu != null:
-			if ProjectSettings.get_setting("escoria/ui/right_mouse_button_action_menu"):
-				if input_event.button_index == BUTTON_RIGHT:
+	elif obj is preload("res://globals/item.gd"):
+		if action_menu and obj.use_action_menu:
+			if !ProjectSettings.get_setting("escoria/ui/right_mouse_button_action_menu"):
 				spawn_action_menu(obj)
 			else:
-					# Left-clicking in this context causes `player` to move to `obj`
 				if obj.has_node("interact_pos"):
 					pos = obj.get_node("interact_pos").get_global_position()
 				else:
 					pos = obj.get_global_position()
 				player.walk_to(pos, walk_context)
-			# Have to verify left button because `clicked` reacts to any click
-			elif input_event.button_index == BUTTON_LEFT:
+
+func right_clicked(obj, pos, input_event = null):
+	if !vm.can_interact():
+		return
+
+	if !player:
+		return
+
+	if mode != "default":
+		return
+
+	var action = obj.get_action()
+
+	if action == "walk":
+		player.walk_to(pos)
+		return
+
+	# Hide the action menu (where available) when performing actions, so it's not eg. open while walking
+	if action_menu:
+		action_menu.stop()
+
+		if obj.use_action_menu:
+			if ProjectSettings.get_setting("escoria/ui/right_mouse_button_action_menu"):
 				spawn_action_menu(obj)
+				return

+	if inventory:
+		if obj is preload("res://globals/item.gd") and obj.inventory:
+			# Do not set `look` as permanent action
+			interact([obj, "look"])
+			# XXX: Moving the mouse during `:look` will cause the tooltip to disappear
+			# so the following is a good-enough-for-now fix for it
+			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", obj.get_tooltip())
+			vm.hover_begin(obj)

 func spawn_action_menu(obj):
 	if action_menu == null:
diff --git a/device/globals/item.gd b/device/globals/item.gd
index 7862fd1..e71f947 100644
--- a/device/globals/item.gd
+++ b/device/globals/item.gd
@@ -99,7 +99,11 @@ func input(event):
 	if event is InputEventMouseButton || event.is_action("ui_accept"):
 		if event.is_pressed():
 			clicked = true
+
+			if event.button_index == BUTTON_LEFT:
 				get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, event.get_global_position(), event)
+			elif event.button_index == BUTTON_RIGHT:
+				get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "right_clicked", self, event.get_global_position(), event)
 			_check_focus(true, true)
 		else:
 			clicked = false
```


device/globals/game.gd
```diff
@@ -99,7 +99,20 @@ func set_current_tool(p_tool):
...
+		return
+
+	if mode != "default":
+		return
+
 	var walk_context = null
+	var action = obj.get_action()
+
+	# XXX: Why does this disable joystick_mode?
+	joystick_mode = false
```
fleskesvor: I think it's because joystick and mouse input mouse don't work well together. Eg. when in joystick mode, the action menu is spawned when the player is within a certain distance of an object, and the code on lines 127 and 128 closes it on every click, which might be inconvenient. Presumably, if the mouse is clicked, it means that the game is not being played with a joystick.

mjtorn:
On Mon, Apr 23, 2018 at 03:19:35PM -0700, Eivind wrote:
>fleskesvor commented on this pull request.

>Since `clicked()` and `right_clicked()` are so similar, would it be better to
>check `event.button_index` in `clicked()` and only spawn the action menu if
>`ProjectSettings.get_setting("escoria/ui/right_mouse_button_action_menu")` is
>set, or are there other major differences?

Definitely maybe. I have a usability peeve with right-clicking that I hope
to address, ie. right-clicking the background. Currently it emits a signal,
but IMO it would be a better game if there's a "Use key with" situation and
right-clicking would reset the action.

I think this and potential other things would benefit from having the right
click in its own function.

Furthermore it keeps the basic left click a bit simpler, because I think the
mess of if/else in there could maybe be simplified. This is also something
I might look into after this "groundwork" PR goes in.

Does that make any sense?

StraToN:
This is an interesting PR. 
I think we have to take care not to close the door to other UI types. There are many of them, and it will be desirable in the near future to use a cleaner way to switch UI types without having to edit the code nor having to manage every UI in the same block code.

fleskesvor:
@mjtorn I'm not sure if that was before your time with Escoria, but there's some discussion about input handling in #44. We're painfully aware that input handling is difficult to maintain at the moment. :cry: It would for sure be easier if this functionality didn't have to implement both an ad-hoc state machine and decision making at once, but could instead just be aware that a click happened and check to see which action is associated with that click.

I'm a bit wary of adding to this code though, because I'm afraid that it will make it more difficult to undo later. But also because it's difficult to do proper regression testing, since there are currently three supported control schemes—verb menu, action menu and one-click input—and the demo uses only one, with no easy way to switch. That's also not counting joystick_mode (which might be broken as it is) and right-click action menu.

I'm not against merging this PR, but it's also not an easy decision.

device/globals/game.gd
mjtorn: This sounds like it requires more than a slight refactor... Which is outside the scope of this PR, IMO.

device/globals/item.gd
```diff
@@ -99,7 +99,11 @@ func input(event):
 	if event is InputEventMouseButton || event.is_action("ui_accept"):
 		if event.is_pressed():
 			clicked = true
-			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, event.get_global_position(), event)
+
+			if event.button_index == BUTTON_LEFT:
+				get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, event.get_global_position(), event)
+			elif event.button_index == BUTTON_RIGHT:
+				get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "right_clicked", self, event.get_global_position(), event)
```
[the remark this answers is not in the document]
fleskesvor: What if we introduce the concept of primary and secondary click here and use the button index and the right_mouse_button_action_menu property to check if the primary or secondary button was clicked? `right_clicked()` would be `secondary_clicked()` or something instead and you pass the boolean to those functions and have logic depend on that instead of the button_index. I think that will be more compliant with the future API, and should work better with other input devices as well, I think.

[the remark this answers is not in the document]
mjtorn: I don't really know what the future API is, is that related to the master-branch-of-Godot stuff we saw from Straton earlier?

I think that is definitely the direction to go towards, to decouple the literal event from the idea of the event, but is it relevant in the scope of this PR?

To me it seems like it could be on top of this and have one less PR in the queue.

I'm a bit unsure about what you mean with the right_mouse_button_action_menu being passed; do you mean that the check to see if an action menu should be spawned is a part of the function signature and not a check in the function body?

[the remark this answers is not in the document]
fleskesvor: > I don't really know what the future API is, is that related to the master-branch-of-Godot stuff we saw from Straton earlier?

Yeah, I meant the stuff that's been discussed in #44.

> I think that is definitely the direction to go towards, to decouple the literal event from the idea of the event, but is it relevant in the scope of this PR?

Maybe not, but if there's a simpler and more future-proof way to achieve the same thing, I think it's worth discussing before code is merged.

> I'm a bit unsure about what you mean with the right_mouse_button_action_menu being passed; do you mean that the check to see if an action menu should be spawned is a part of the function signature and not a check in the function body?

I was in a bit of a hurry, so some of the stuff I wrote didn't make sense. What I meant was that you'd have something like this in the input function:

    var primary_button = event.button_index == BUTTON_LEFT or (event.button_index == BUTTON_RIGHT and ProjectSettings.get_property("escoria/right_is_primary")
    var position = event.get_global_position()

    get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, position, primary_button)

and replace the logic in `clicked()` that depends on `event.button_index` with a check for primary_button.

I realize now that the "right click action menu" setting isn't exactly the same as flipping the buttons though, since you still want to walk with left-click, right?

So doing what I'm thinking of cleanly also require checking whether the click happened on an item or the background, and probably some other things, so I agree that it's outside of the scope of this PR. When that's in place though, there won't be a need for a second clicked menu, I think, but maybe it's a good idea to have one in the meantime, even though it duplicates a lot of code.

What do you think, @StraToN?

mjtorn:
On Thu, Apr 26, 2018 at 07:28:56AM -0700, Eivind wrote:
>fleskesvor commented on this pull request.
>
>
>
>> @@ -99,7 +99,11 @@ func input(event):
> 	if event is InputEventMouseButton || event.is_action("ui_accept"):
> 		if event.is_pressed():
> 			clicked = true
>-			get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, event.get_global_position(), event)
>+
>+			if event.button_index == BUTTON_LEFT:
>+				get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, event.get_global_position(), event)
>+			elif event.button_index == BUTTON_RIGHT:
>+				get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "right_clicked", self, event.get_global_position(), event)
>
>> I don't really know what the future API is, is that related to the master-branch-of-Godot stuff we saw from Straton earlier?
>
>Yeah, I meant the stuff that's been discussed in #44.

Will need to revisit that :)

>> I think that is definitely the direction to go towards, to decouple the literal event from the idea of the event, but is it relevant in the scope of this PR?
>
>Maybe not, but if there's a simpler and more future-proof way to achieve the same thing, I think it's worth discussing before code is merged.

Sure.

>> I'm a bit unsure about what you mean with the right_mouse_button_action_menu being passed; do you mean that the check to see if an action menu should be spawned is a part of the function signature and not a check in the function body?
>
>I was in a bit of a hurry, so some of the stuff I wrote didn't make sense.

Writing this on the bus after one hell of a day, I relate to hurried mistakes
and may even have made one or two myself, no worries ;)

> What I meant was that you'd have something like this in the input function:
>
>    var primary_button = event.button_index == BUTTON_LEFT or (event.button_index == BUTTON_RIGHT and ProjectSettings.get_property("escoria/right_is_primary")
>    var position = event.get_global_position()
>
>    get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "clicked", self, position, primary_button)
>
>and replace the logic in `clicked()` that depends on `event.button_index` with a check for primary_button.
>
>I realize now that the "right click action menu" setting isn't exactly the same as flipping the buttons though, since you still want to walk with left-click, right?

Yeah, exacly that. CoMI-style. So right_is_primary would be misleadimg, as
this is closer to keybindings that are made on an engine level, if you will.

>So doing what I'm thinking of cleanly also require checking whether the click happened on an item or the background, and probably some other things, so I agree that it's outside of the scope of this PR. When that's in place though, there won't be a need for a second clicked menu, I think, but maybe it's a good idea to have one in the meantime, even though it duplicates a lot of code.

You meant "a second clicked function"?

But yeah, heh, even I as the author of this PR see this as a stepping stone to something
even better ;P

fleskesvor:
> You meant "a second clicked function"?

Yeah, I did, although a second action menu on the other button does sound like a crazy concept. Come to think of it, we really should bind input to actions via the input API, like we talked about before, but I forgot about that again. I will do some testing on this branch tonight and get some first hand experience with the value of these changes. Either way though, I don't want us to get more married to the concept of left and right click than we already are, so I think the function should change name and/or a FIXME should be added.

mjtorn:
> Either way though, I don't want us to get more married to the concept of
> left and right click than we already are, so I think the function should
> change name and/or a FIXME should be added.

You got both in a force-push now. A quick amend.

device/globals/game.gd
```diff
@@ -110,81 +123,86 @@ func clicked(obj, pos, input_event = null):
...
-					else:
-						pos = obj.get_global_position()
-					player.walk_to(pos, walk_context)
-			# Have to verify left button because `clicked` reacts to any click
-			elif input_event.button_index == BUTTON_LEFT:
 				spawn_action_menu(obj)
+				return
 
+	if inventory:
+		if obj is preload("res://globals/item.gd") and obj.inventory:
```
fleskesvor: `esc_type.ITEM`

device/globals/game.gd
```diff
@@ -110,81 +123,86 @@ func clicked(obj, pos, input_event = null):
...
-		var action = obj.get_action()
-		# Hide the action menu (where available) when performing actions, so it's not eg. open while walking
+	# Hide the action menu (where available) when performing actions, so it's not eg. open while walking
+	if action_menu:
+		action_menu.stop()
+
+	if action == "walk":
+		#click.set_position(pos)
+		#click_anim.play("click")
+
```
fleskesvor: ^ This commented out code can be removed.

device/globals/game.gd
```diff
@@ -99,7 +99,20 @@ func set_current_tool(p_tool):
...
 
 func clicked(obj, pos, input_event = null):
+	if !vm.can_interact():
+		return
+
+	if !player:
+		return
+
+	if mode != "default":
+		return
```
fleskesvor: Might be better with:

    if not vm.can_interact() or not player or mode != "default":
        return

?

mjtorn:
On Thu, Apr 26, 2018 at 03:28:16PM -0700, Eivind wrote:
>+	if action == "walk":
>+		#click.set_position(pos)
>+		#click_anim.play("click")
>+
>
>^ This commented out code can be removed.

I actually think it should be `if click:` because of the code in `_ready_`. If you
have an animation like a ripple-on-a-pond or rotating-circle click effect.

I'll do that instead when I push soon. We can continue the discussion then if
it feels bad.

mjtorn:
On Thu, Apr 26, 2018 at 10:31:29PM +0000, Eivind wrote:
>fleskesvor commented on this pull request.
>
>Might be better with:
>
>    if not vm.can_interact() or not player or mode != "default":
>        return

I would like to make another case.

Having each `if` separate means many things:

  1. They are explicit
  1. They are very legible
  1. It's easier to alter their behavior
  1. Adding more cases doesn't mean adding (parentheses and overflowing) the lines into crazy long lines that are just worse on the cognitive restrictions in people's brains

mjtorn:
On Thu, Apr 26, 2018 at 03:23:36PM -0700, Eivind wrote:
>fleskesvor commented on this pull request.
>+	elif obj is preload("res://globals/item.gd"):
>
>You could use `esc_type.ITEM` here.

Yes, of course, I didn't think that the types PR was already merged.

My bad.

fleskesvor:
> I would like to make another case.
>
> Having each `if` separate means many things:
>
>  1. They are explicit
>  1. They are very legible
>  1. It's easier to alter their behavior
>  1. Adding more cases doesn't mean adding (parentheses and overflowing) the lines into crazy long lines that are just worse on the cognitive restrictions in people's brains

Yes, you're right. I think it would be better, and more DRY, if you separated those conditions into a separate function `can_click()` function though, since they're the same for both left and right clicks.

fleskesvor:
> I actually think it should be `if click:` because of the code in `_ready_`. If you
have an animation like a ripple-on-a-pond or rotating-circle click effect.
>
> I'll do that instead when I push soon. We can continue the discussion then if
it feels bad.

Is that documented anywhere though? I don't remember seeing any such nodes.

mjtorn:
>> I actually think it should be if click: because of the code in _ready_. If you have an animation like a ripple-on-a-pond or rotating-circle click effect.

> Is that documented anywhere though? I don't remember seeing any such nodes.

That's my interpretation of the code, but I didn't test it. This is why I'm so scared about issues like #10 going overboard. Some of the dead code may be valuable if it was documented and/or implemented.

mjtorn:
> Yes, you're right. I think it would be better, and more DRY, if you separated those conditions into a separate function can_click() function though, since they're the same for both left and right clicks.

I can't argue with that :D Pushed. We can figure out what to do if the checks ever drift apart between the functions, but it is much cleaner now, thanks for pointing it out.

device/globals/game.gd
```diff
@@ -110,81 +131,82 @@ func clicked(obj, pos, input_event = null):
...
+					pos = obj.get_node("interact_pos").get_global_position()
+				else:
+					pos = obj.get_global_position()
 				player.walk_to(pos, walk_context)
-				# Leave the tooltip if the player is in eg. a "use key with" state
-				if !current_action:
-					get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_tooltip", "")
 
-		elif obj.use_action_menu && action_menu != null:
+ func secondary_click(obj, pos, input_event = null):
```
fleskesvor: There's a space before `func` which is throwing a "misplaced func" error and aborts the game. Otherwise it looks good. 🙂

mjtorn:
On Sun, Apr 29, 2018 at 02:40:40PM -0700, Eivind wrote:
>fleskesvor commented on this pull request.

>-		elif obj.use_action_menu && action_menu != null:
>+ func secondary_click(obj, pos, input_event = null):
>
>There's a space before `func` which is throwing a "misplaced func" error and aborts the game. Otherwise it looks good. 🙂 

The joys of quick fixes. I'll be sure to fix this as soon as I'm
at a computer and also ping on irc when done.

Thanks for testing!

mjtorn:
Is this cleared for merging, @fleskesvor? 

For posterity, the original diff up there isn't the final merge, as may be expected.

PR #134, opened 2018-04-23
Spine support

mjtorn:
This code does not support every possible scenario at the moment, but I think it's fine because it doesn't seem to break anything either and it's very much an opt-in feature.

Please give it a spin, I couldn't break anything in the test room, but something may always come up.

I hope to work more on this in the future, our AdventureJam requires NPCs (implemented either as a new script or through `item.gd`) to be animated, so some future work is guaranteed.

### DIFF
--- a/device/globals/interactive.gd
+++ b/device/globals/interactive.gd
@@ -120,10 +120,11 @@ func _process(time):
 
 		#pose_scale = animations.directions[last_dir+1]
 
+	if self is esc_type.ITEM:
 		_update_terrain()
 
 func _find_sprites(p = null):
-	if p is Sprite || p is AnimatedSprite || p is TextureRect || p is TextureButton:
+	if p is Sprite || p is AnimatedSprite || p is TextureRect || p is TextureButton || p.get_class() == "Spine":
 		sprites.push_back(p)
 	for i in range(0, p.get_child_count()):
 		_find_sprites(p.get_child(i))
@@ -136,5 +137,6 @@ func _ready():
 
 	if Engine.is_editor_hint():
 		return
-	if has_node("animation"):
-		animation = get_node("animation")
+
+	animation = $"animation"
+
--- a/device/globals/item.gd
+++ b/device/globals/item.gd
@@ -16,6 +16,7 @@ export var active = true setget set_active,get_active
 export var placeholders = {}
 export var use_custom_z = false
 
+var animation_is_spine
 var anim_notify = null
 var anim_scale_override = null
 
@@ -185,6 +186,37 @@ func global_changed(name):
 	elif "global_changed" in event_table:
 		run_event(event_table.global_changed)
 
+func _animation_play(args):
+	if !animation:
+		return
+
+	assert(typeof(args) == TYPE_DICTIONARY)
+
+	## Then tackle the hard part, though name must always be present
+	var name = args["name"]
+
+	if !animation_is_spine:
+		# play(String name=”“, float custom_blend=-1, float custom_speed=1.0, bool from_end=false)
+		var custom_blend = args["custom_blend"] if args.has("custom_blend") else -1
+		var custom_speed = args["custom_speed"] if args.has("custom_speed") else 1.0
+		var from_end = args["from_end"] if args.has("from_end") else false
+
+		animation.play(name, custom_blend, custom_speed, from_end)
+	else:
+		# play(const String &p_name, bool p_loop, int p_track, int p_delay)
+		var loop = args["loop"] if args.has("loop") else false
+		var track = args["track"] if args.has("track") else 0
+		var delay = args["delay"] if args.has("delay") else 0
+
+		animation.play(name, loop, track, delay)
+
+func _animation_get_current_animation():
+	if !animation:
+		return
+
+	# XXX: Spine defaults to track 0
+	return animation.get_current_animation()
+
 func anim_get_ph_paths(p_anim):
 	if !(p_anim in placeholders):
 		return null
@@ -198,11 +230,8 @@ func anim_get_ph_paths(p_anim):
 	return ret
 
 func play_anim(p_anim, p_notify = null, p_reverse = false, p_flip = null):
-
-	if p_notify == null && (!has_node("animation") || !get_node("animation").has_animation(p_anim)):
-		print("skipping cut scene '", p_anim, "'")
+	if p_notify == null && (!animation || !animation.has_animation(p_anim)):
 		vm.finished(p_notify)
-		#_debug_states()
 		return
 
 	if p_anim in placeholders:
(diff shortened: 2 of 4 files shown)


fleskesvor:
As mentioned on IRC before the weekend, I'm very sceptical of adding software specific logic to Escoria. Especially when that software is both proprietary and only runs on proprietary platforms. Ideally, Spine would either export directly to Godot or to an intermediate open format that could be used by Godot.

They also have an issue for adding support for Godot, which has been marked as "ready", though there is no further news:

https://github.com/EsotericSoftware/spine-runtimes/issues/728

What are your thoughts on this, @StraToN?

mjtorn:
Even better would be if Godot allowed method overloading so the Spine module could be beat into API compliance and Escoria wouldn't have to care.

I would really hate to have to permanently fork Escoria and maintain it over a feature that some others may actually use as well.

Does it really matter that much that the JSON exports have been made with closed software and the runtime is a bit so-so on the opensource/free front?

mjtorn:
There was a reply on the above issue that "ready" means "ready to start development", so let's see. If this doesn't go in, maybe I'm stuck with keeping this commit around, but @StraToN 's opinion would also be valuable.

fleskesvor:
> Even better would be if Godot allowed method overloading so the Spine module could be beat into API compliance and Escoria wouldn't have to care.

Yes, I agree that this is more of a concern for Godot projects in general and not one specific to Escoria.

> I would really hate to have to permanently fork Escoria and maintain it over a feature that some others may actually use as well.

I understand that this would be very inconvenient. However, would it not be almost as convenient to just make a copy of `player.gd` with your changes that can live with your project files instead of Escoria upstream? That way you would only need to diff against Escoria's `player.gd` the few times there `player.gd` has some upstream changes that you want to include in your project. Either that or apply your changes from stash or checkout that one file from a branch.

> Does it really matter that much that the JSON exports have been made with closed software and the runtime is a bit so-so on the opensource/free front?

My concern is that by adding software specific logic we help maintain a reliance on that specific proprietary software. I feel the same way about accepting docx as a de-facto document standard. It may be convenient but I feel uneasy about it.

mjtorn:
On Mon, Apr 23, 2018 at 01:33:24PM -0700, Eivind wrote:
>> Even better would be if Godot allowed method overloading so the Spine
>> module could be beat into API compliance and Escoria wouldn't have to care.
>
>Yes, I agree that this is more of a concern for Godot projects in general and
>not one specific to Escoria.

I agree, men det er livet :/ Gotta play the cards we're dealt.

>> I would really hate to have to permanently fork Escoria and maintain it
>> over a feature that some others may actually use as well.
>
>I understand that this would be very inconvenient. However, would it not be
>almost as convenient to just make a copy of `player.gd` with your changes
>that can live with your project files instead of Escoria upstream? That way
>you would only need to diff against Escoria's `player.gd` the few times there
>`player.gd` has some upstream changes that you want to include in your
>project. Either that or apply your changes from stash or checkout that one
>file from a branch.

You make a good point. Basically a stash in a no-go, because by implication
I would never commit the Spine code anywhere. It could be in a separate branch
from which I cherry-pick it. Then I'd track master (and feature branches) but
constantly rebase my gamedev branch, and maintain a Spine branch, and re-apply
it from the branch.

In this case it would be easier to simply rebase constantly.

However, a spine_player.gd file could maybe work the easiest, but patch
management halfway outside git is also a hassle. This is why PRs are made ;)

Are you completely opposed to merging a spine_player.gd file into master?

Or some other kind of architecture that separates Spine support? I don't
know what it would be beside that one file. Problem is, though, that whoever
touches player.gd should also touch spine_player.gd which can be easy to
miss.

It could also be that someone gets into Dragonbones and would we then have
a dragonbones_player.gd? Would it be more acceptable because it's more
open-source? Probably not because that's reliance too.

>> Does it really matter that much that the JSON exports have been made with
>> closed software and the runtime is a bit so-so on the opensource/free
>> front?

>My concern is that by adding software specific logic we help maintain a
>reliance on that specific proprietary software. I feel the same way about
>accepting docx as a de-facto document standard. It may be convenient but I
>feel uneasy about it.

I feel uneasy too, but getting things done has to trump uneasy feels. The
upside here is that the API used from the Spine module appears very stable
and Escoria would use a limited subset of it.

.docx sucks golf balls compared to the Spine API, which is beautiful and
well documented.

The other upside is that getting things done has to trump massaging a pull
request, if the answer to everything is no. A leadership decision needs to
be made, like the Esoteric Software guy on the other issue, and then again
c'est la vie and I gotta deal with it.

I also feel uneasy because I may have inadvertedly cracked a can of worms,
like some kind of Pandora, but with code :/

Thanks for the good feedback even if it probably ends up hurting me! :D

mjtorn:
I had a new idea as well. Could there be a `spine` branch in the main repository? It would have a notice in the readme that it gets rebased often and force-pushed, so it should be used for cherry-picking only, and then it would be sort of supported but off-master.

Then people like me who still have to rebase their gamedev branches/repos onto Escoria master could drop the spine commit and cherry-pick HEAD from that branch.

Nothing is hassle-free but then it would be hosted.

However, that branch should never contain more than one or two commits (if the README notice is separate) and I don't know if you want to give me that kind of power to the main repo, even if I swear every branch that goes to master will continue to respect the PR process.

fleskesvor:
> It could be in a separate branch
from which I cherry-pick it. Then I'd track master (and feature branches) but
constantly rebase my gamedev branch, and maintain a Spine branch, and re-apply
it from the branch.

> In this case it would be easier to simply rebase constantly.

> However, a spine_player.gd file could maybe work the easiest, but patch
management halfway outside git is also a hassle. This is why PRs are made ;)

A separate branch sounds like a good idea, because if `spine_player.gd` is the only file changed on that branch, you can always do a `git rebase -i master` without conflicts. I wouldn't cherry-pick. As for using `spine_player.gd` in your project based on Escoria master, I think `git checkout spine-branch -- device/globals/spine_player.gd` would be easier than using cherry-pick, especially if you have more than one commits on that branch.

> Are you completely opposed to merging a spine_player.gd file into master?

> Could there be a spine branch in the main repository? 

I'm not completely opposed to it and maybe a separate branch would be a good idea, since I do agree that this might benefit more game developers. I do, however, think it's a difficult decision for the reasons mentioned before, and ultimately it's up to either @punto- or @StraToN.

fleskesvor:
@akien-mga suggested on IRC that it's fine to add support for Spine, provided that it's done in a modular way, eg. by extending `player.gd` or creating an interface that can be inherited by the current `player.gd`, `player_spine.gd` and any other software specific implementation that is added in the future. The implementation can live in the main repo as long as it can be contributed under the current MIT license and isn't limited by a Spine license.

Another possible problem that just occured to me is, how will we support this? Are there any exported Spine assets provided freely that can be used for implementation and testing? Some quick googling didn't turn up much.

mjtorn:
I can't look into this more in-depth but I think [this](https://github.com/EsotericSoftware/spine-superspineboy) is a free test. Gave the link to @StraToN on the irc channel too, but here for posterity.

PR #140, opened 2018-04-26
Fix localization helper script, lockit.gd

fleskesvor:
The script can be run with `godot --script ../lockit.gd` from the project root, or by using the `-s` short option, but not `-script` like the Floss Manuals tutorial says:

https://fr.flossmanuals.net/creating-point-and-click-games-with-escoria/i18n/

There's room for improvement here. Using a shell command to tidy up .lockit files doesn't seem very convenient. Maybe it would be better to support an in-place option like `sed` does. Ie. overwrite the .esc file but give an optional file extension that will be used to create a backup of the original.

### DIFF
--- a/tools/lockit.gd
+++ b/tools/lockit.gd
@@ -3,10 +3,10 @@ extends MainLoop
 func _iteration(time):
 	return true
 
-
 func get_input_files(argv):
 
 	var list = []
+
 	if "-esc" in argv:
 		for i in range(argv.size()):
 			if argv[i] == "-esc":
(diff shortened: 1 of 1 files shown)


StraToN:
> Using a shell command to tidy up .lockit files doesn't seem very convenient. Maybe it would be better to support an in-place option like sed does. Ie. overwrite the .esc file but give an optional file extension that will be used to create a backup of the original.

I agree with this. I'll create an issue for this. However we can still merge these changes for now, as they clean the script a bit.

fleskesvor:
> we can still merge these changes for now, as they clean the script a bit

They also make the script not broken. 😛Thanks, @StraToN. 🙂

PR #141, opened 2018-04-27
Music stuff

lavaduderDev:
bg_music was not working as intended. So I made a hotfix.

I also added in a little melody (Not the best.) and had it load in when the game starts.


lavaduderDev:
`demo_melody.mid` I'd figured people want to see the source for music. A midi was the best compromise. If you want me to remove it. I will.

As for the hotfix. I wasn't quite sure how `:load` worked, so maybe I'll change it to `:ready`.

fleskesvor:
> `demo_melody.mid` I'd figured people want to see the source for music. A midi was the best compromise. If you want me to remove it. I will.

No, if midi is a more accessible format than ogg, it's a good idea to include it. If it isn't used in the demo, it should be in a directory outside of the project root though, eg. `assets/music/` (which doesn't currently exist), and it needs to include a license. If you didn't make it yourself, make sure it's in compliance with the original license. If you made it, you should [choose a license](https://choosealicense.com/). In either case, you can add a new "Music credits" section at the bottom of the README, similar to the "Art credits" section.

Regarding `:load` vs `:ready`, if `:ready` is good enough for now, you can go with that, and I'll make an issue about fixing it in a way so that `telon` and `bg_music` can be registered and used without being subject to race conditions.

fleskesvor:
The `bg_music` issue has been fixed by #170, so if the changes to `device/globals/vm_level.gd` is reverted and the feedback in my previous comment is addressed, this can be merged.

lavaduderDev:
I removed the Midi. I wasn't quite sure how to handle it, but what I did was `--force` it. The hotfix was also removed.

fleskesvor:
Thanks, @lavaduderDev. :slightly_smiling_face:

This could be merged as-is, but I still need clarification on whether the music is something you wrote yourself, and if so, what license you want it made available under. Had it been a pure code contribution, the MIT license which is used for Escoria would have applied, but since that is a software license, I think that means any assets need explicit licensing, or they will be unlicensed. In my understanding, this adds restrictions, since users would have to ask us for permission to use in their own projects.

Maybe @akien-mga can chime in on this?

akien-mga:
No license and OP's account was deleted, so not much can be done with this PR.

If we want a demo music piece, we could pick a CC-BY or CC-BY-SA one on OpenGameArt.

PR #147, opened 2018-05-02
Changing costumes

mjtorn:
I could not test this (edit: beyond the point of seeing the `set_costume` function is called) because I have no idea how to make the "hat" animation work out in the player scene. At least the default costume works!

@fleskesvor you wanna nudge me in the right direction or even send me some code?

The original code I stole this from uses binary scenes so there's little I can do right now.


fleskesvor:
> @fleskesvor you wanna nudge me in the right direction or even send me some code?

I made a PR on your branch, which adds the second costume to the player scene.

An issue, which I'm not sure if have practical consequences or not, is that if the selected frame set in the `AnimatedSprite` is different from the current costume, the correct costume won't be set until the player starts moving or the costume is set explicitly from an esc script.

As punto mentioned on my original PR, #33, we also need to save the current costume, or else the default will be used when a save is loaded.

fleskesvor:
I made a new PR on your branch (mjtorn/escoria#4) and I'm re-posting my comments on it here for posterity:

> I moved `player.gd`'s functionality into `set_state` which was currently stubbed and rewrote the example esc scripts to use `set_state` for items instead of `set_costume`. The main advantage of this is that it converges the two very similar concepts and there is already code in place for saving state, so we get that one for free.
>
> I also added the floppy disk, which is a simple example of using `item.gd`'s `set_state` for something other than just swapping textures and it also makes it easy to verify that state has been saved by `:use`-ing that, quitting the game and choosing to "Continue" from the menu. (Just playing the scene in the editor does not load the autosave.)
>
> I also think the PR should have a more integrated example of changing player costumes than `:use`-ing the can and ice cream. Maybe there should be a hat in the room that you can pick up and use from the inventory to toggle the costume? I think a pink bycocket with a green feather on one side would be awesome, but I'm not picky. We can always improve it later.

mjtorn:
As I said in the original PR, all this is great, but I would respectfully ask that improving the demo scene is a separate effort.

Due to reasons, I would also hope that this PR gets merged without having to squash the `set_state` code into the history, as that's an insane effort as well.

fleskesvor:
> As I said in the original PR, all this is great, but I would respectfully ask that improving the demo scene is a separate effort.

Yes, I agree.

> Due to reasons, I would also hope that this PR gets merged without having to squash the set_state code into the history, as that's an insane effort as well.

As discussed on IRC, both @StraToN and I are fine with this.

Thanks for the contribution! :smile:

fleskesvor:
I should also mention that punto's original concerns from #33 are adressed in this PR.

PR #148, opened 2018-05-05
Make zoom take optional time to reach target

fleskesvor:
This PR supports an optional time in second that can be provided as a second parameter to `camera_set_zoom` and `camera_set_zoom_height` to control how long it takes for the camera to zoom into place.

Since neither this command nor `camera_set_target` are blocking they can be used together and in combination with `wait` to achieve better control over the camera, and this hopefully resolves the remaining issues of #111. 

I experimented with using a speed similar to `camera_set_target`, but it wasn't very useful, since `camera_set_target` is dependent on distance, while `camera_set_zoom` is not. There should possibly be a command similar to or a replacement for `camera_set_target` which uses time reach target instead of speed to control the camera, but that's a separate issue.

I also removed the zoom stack, which didn't feel very useful, and which is not compatible with fixed zoom levels. Instead, I implemented a simple zoom stack with esc script in demo room 2.


mjtorn:
I love this PR!

In my proto, the zoom does look a bit wonky because it pushes in on the NPC, but despite `set_target`, the NPC is not in the center. What happens is the two characters are a bit to the upper right, but having a slow zoom, say 5s, shows that the camera pushes in on its center, until it (apparently) hits the camera limit, and zooms in a way that respects the limit.

Anyway, the wonkiness is dealt with in #111 and for what it's worth, I hope this gets merged asap, really good work! :)

PR #149, opened 2018-05-06
Add interactive items in parallax layer

fleskesvor:
This makes the sun in demo room 2 interactive, where it is in a parallax layer.

My original plan was to achieve this by making `background_area.gd` backgrounds (`Sprite`) hand editable, because I thought that would make it possible to interact with items behind it. I also thought I had done exactly this while creating room 2 in the first place, but I could find no such implementation in any of my branches or stash.

The solution I ended up with was making the sun use a `TextureButton` with a click mask, because that just works. Naturally, that comes with the side effect of being visible through the click shape used for the "foreground background" so there also needs to be a `Control` type item in the foreground to block input. That is why I also added a `TextureButton` with a click mask for the mountain. I also attached the `item.gd` script to it, but that is not a requirement. Simple foreground shapes could just be `TextureRects` with no texture attached.

One remaining issue is that interacting with items in the background doesn't work well at the moment. Because of the offset, the player character will often walk far to the side of the sun before interacting with it. I should probably also write some basic documentation on how to create a click mask, since it might not be straightforward.


fleskesvor:
> One remaining issue is that interacting with items in the background doesn't work well at the moment. Because of the offset, the player character will often walk far to the side of the sun before interacting with it.

I've found a solution to this, which I think is acceptable. I extend `item.gd` with an exported `NodePath` variable, which can be used to set an `interact_pos` in the foreground.

PR #150, opened 2018-05-08
Polishing of set_state

mjtorn:
The endless spam on every finished animation had to be silenced.

I also explained in detail how to use `AnimationPlayer` to create states.


StraToN:
Thanks a lot ! :+1:

fleskesvor:
Good job. :heart:

Issue #142, opened 2018-04-30
Test and document exports


fleskesvor:
Test that exports are working on

- [x] Linux
- [x] macOS
- [x] Windows

HTML5 exports are know to not be working at the moment.

If possible, test that autosaves are working, and if not, create bug report.

When exporting, *.esc needs to be included explicitly, and that must be documented. Documentation should possibly also include settings/saves paths as documented in http://docs.godotengine.org/en/3.0/tutorials/io/data_paths.html (I think the Mac path is wrong though, re: godotengine/godot#13064.)

fleskesvor:
While testing, I noticed that selecting "No" does not work, neither on "Quit" nor on "New Game" from an already running game. Moreover, the latter seems to leave the game in a very odd state when pressing "Yes" after first having pressed "No.

As far as I can tell, this seems to always have been an issue in 3.0, while the tip of 2.1 works (with Godot 2.x). We might perform a bisect to check when the issue was introduced, but that will require a first pass with bisect to find the first commit which works with Godot v3.0.2 stable.

mjtorn:
I just started setting up the Git stuff for our Jam game, so I can test the material I got from the animators and such, and seeing this kind of an update is nearly panic-worthy.

Do you have some more info on this?

Export the demo scene? Does that mean no menus work?

fleskesvor:
@mjtorn It should be fixed in my latest PR.

mjtorn:
@fleskesvor I thought this was only with exported games, but because I was mistaken, I got to test your PR and it works for me as well.

fleskesvor:
I discovered the bug while testing exports, and assumed so at first, but it turns out that I just hadn't performed the actions required to trigger the bug while testing locally.

Thanks for testing the PR. :slight_smile:

fleskesvor:
I've tested the Linux export now, and it seems to be working without noticeable issues. Things I tested:

- "New game" from the start menu loads the test room as defined by `res://demo/game.esc`
- "Continue" from the start menu loads an autosave if `~/.local/share/godot/app_userdata/Escoria\ Prototype/quick_save.esc` exists. Otherwise the option is grayed out.
- States and globals are correctly added and loaded from `quick_save.esc`
- "Continue" from in-game menu closes the menu
- "New game" from in-game menu makes a prompt appear. Pressing "Yes" restarts the game; pressing "No" closes the prompt (but not the menu).
[...]

PR #152, opened 2018-05-10
Fix confirm popup to close on animation_finished

fleskesvor:
Fixes this issue mentioned in #142:

> While testing, I noticed that selecting "No" does not work, neither on "Quit" nor on "New Game" from an already running game. Moreover, the latter seems to leave the game in a very odd state when pressing "Yes" after first having pressed "No.
>
> As far as I can tell, this seems to always have been an issue in 3.0, while the tip of 2.1 works (with Godot 2.x), and it looks like it was an issue in 2.1 too. We might perform a bisect to check when the issue was introduced, but that will require a first pass with bisect to find the first commit which works with Godot v3.0.2 stable.

### DIFF
--- a/device/demo/ui/confirm_popup.tscn
+++ b/device/demo/ui/confirm_popup.tscn
@@ -5,18 +5,18 @@
 [sub_resource type="Animation" id=1]
 
 resource_name = "close"
-length = 1.0
+length = 0.1
 loop = false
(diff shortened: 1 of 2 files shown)


mjtorn:
I actually hit this sooner than I thought. Go forth and merge, it works!

PR #153, opened 2018-05-10
Make the "boot" script a ProjectSetting

mjtorn:
With this change, you don't need to commit a change into Escoria's
code to start your own game instead of the demo.


fleskesvor:
I might revert this in my open UI PR, since the menu script is highly project specific and should live in `demo/ui`/`game/ui/`, but until that happens, this is a much better solution than editing framework scripts. Thanks, @mjtorn. :slightly_smiling_face:

PR #154, opened 2018-05-10
Do not look for inventory if it's not enabled

mjtorn:
I bumped into this while setting up things for the jam game. I have a path for an inventory but it's not created yet, which obviously leads into problems. That's when I noticed there's an `export var` to disable the inventory altogether!

Feels like a super-useful feature in early development and why not respect it in real games as well.


fleskesvor:
Sounds like a good idea. Thanks! :slightly_smiling_face:

PR #155, opened 2018-05-11
Fix checking inventory.is_visible() when inventory is null

mjtorn:
Because the inventory might be disabled, we would call `is_visible()` on null. Let's not do that.


device/globals/game.gd
```diff
@@ -37,7 +37,7 @@ func mouse_enter(obj):
 		tooltip.set_position(pos)
 
 	# We must hide all non-inventory tooltips and interactions when the inventory is open
-	if action_menu and inventory.is_visible():
+	if action_menu and inventory and inventory.is_visible():
```
fleskesvor: There's also a usage of `inventory.get_action()` on L45 which might be problematic. All other accesses of properties on `inventory` seems to be guarded by null checks, as far as I can tell.

PR #156, opened 2018-05-11
Shuffle elif inventory to place so tooltips work

mjtorn:
I should have tested the previous commit for real, but it seemed so obvious.

This is what you get when the panic kicks in. Sorry.


fleskesvor:
95% of the blame is on me. 😖Thanks for the fix!

PR #157, opened 2018-05-11
Renamed doc/ folder to docs/

StraToN:
In order to use the doc/ folder for Github Pages, one needs to rename it docs/.


fleskesvor:
Looks good to me. :sunglasses:

docs/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+Installation
+------------
+
+[Git submodules](git_submodule.md)
+
+Reference
+---------
+
+[Background](background.md)
```
fleskesvor: There should be empty lines between these links so that they are not inlined.

fleskesvor: The same goes for the Scripting and Tools sections.

StraToN: Fixed

PR #158, opened 2018-05-11
Drop the hard requirement on arrow_prev/arrow_next

mjtorn:
It occurred to me that we're most likely doing a scumm/daiza inventory, with so few items we don't need arrows. Why make them and hide them when we can make them optional?

### DIFF
--- a/device/globals/inventory.gd
+++ b/device/globals/inventory.gd
@@ -141,10 +141,13 @@ func _ready():
 
 	page_size = get_node("slots").get_child_count()
 	sort_items()
-	get_node("arrow_prev").connect("pressed", self, "change_page", [-1])
-	#get_node("arrow_left").set_focus_mode(Control.FOCUS_NONE)
-	get_node("arrow_next").connect("pressed", self, "change_page", [1])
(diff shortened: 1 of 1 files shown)


fleskesvor:
Good point. :+1: Can you also remove the commented out calls to `set_focus_mode`? The constants have moved to the `FocusNode` enum, but if it's not in use, we might as well drop it, I think.

mjtorn:
Thanks, @fleskesvor! I didn't know what they were so I left them in just in case they're something that could be uncommented to (re-)enable a feature. We've had that before, and sounds like `FocusNode` not in use either could or could not be useful, I'd have to understand what it is first :D

It's gone now and I updated to the new get syntax too!

fleskesvor:
The purpose of `FOCUS_NONE` is to make the arrow keys unselectable with controller mode. Presumably, there's some custom logic for cycling the inventory when using the controllers, but I think we will have to rethink how that works if we want to make the abstractions we've talked about. If we do need it in the future, we can find it with Git.

Thanks for the fixes, @mjtorn. :slightly_smiling_face:

PR #159, opened 2018-05-14
Adds plugin management for Spine animations

StraToN:
Ok, this one should supersede #134 by placing all Spine-specific code in a plugin that needs to be activated.

The idea is that every plugin activated in Godot is loaded by the `main` singleton into a list.

An Escoria plugin defines the `escoria_plugin_type()` function that returns a value defined in esc_types. This is mandatory for Escoria to be able to recognize the plugin (if it doesn't, then it is a "classic" plugin). I currently added one plugin type: Animation (the only one we need for now)

Each plugin has a `managed_types()` function that returns a list of node types that the plugin is able to deal with (enabling one plugin to deal with multiple types in different manners).

For each animation node, `interactive.gd` builds a list of sprites. Escoria will first try to find the sprites in AnimationPlayer and AnimatedSprite nodes. Then, it will call each animation plugin's function dedicated for the construction of the sprites list (`_find_sprites()`).

Then, at needed points in the animation management (in `player.gd` and `item.gd`), we call specific plugins functions:
- `_connect_animation_finished_node(animation_node, manager_node, call_func_name)`: connects the given animation_node's animation_finished event to the manager_node's call_func_name.
- `_animation_play(animation_node, last_dir, args)`: plays the animation defined in animation_node according to given parameters
- `_plugin_animation_finished_task_is_blocking(task, animation)`: returns true whether the given task name needs to block the animation. When this is useless, this function should return false.


device/addons/animations-spine/animations_spine.gd
```diff
@@ -0,0 +1,70 @@
...
+		animation_node.play(name, loop, track, delay)
+		animation_node.add(animation_node.idles[last_dir], true)
+
+
+# Function called when animation finishes, to prevent glitches
+# depending on the 'task' value.
+# The return of this function depends on every plugin. It is better to explain here the reasons
+# of the returned value. By default, if there is no reason to have a specific blocking at this moment,
+# this function should return false by default.
+func _plugin_animation_finished_task_is_blocking(task, animation):
```
StraToN: Should be renamed `_animation_finished_task_is_blocking`. 
`_plugin` prefix is for Escoria functions that make calls to plugin.

StraToN:
I've been thinking about this PR while swimming and I've come to the conclusion that the animation plugin design is bad as it currently is:

Currently, the animation management is shared between Escoria scripts `global/item.gd` and `global/player.gd` **and** the plugin script since these 2 scripts call certain plugin functions when required.

This is not good. 
The animation of a custom node (Spine and others) should be **totally** deferred to the plugin. 

Thus, an animation plugin shoud only define 2 functions:
- `_find_sprites()` : initialize the animation system with sprites to load.
- `_animation_play()` : plays the animation. This function is called instead of item.gd/player.gd's one if the node is managed by the plugin.

And both functions should define **entirely** their own way to manage the animation, even if some parts are common with the way Escoria currently does for `AnimationPlayer` and `AnimatedSprite`. 

I'll try to rework this soon.

PR #161, opened 2018-05-15
Work on clamping UI elements

mjtorn:
Tooltips are hard due to how big they usually are, but clamping vertically is a good start and worth it.

Dialogs should never overflow. I think it's possible to mess with the camera so they still do, but in our jam game we see only great success.

Therefore this is one of those may-not-solve-world-hunger-but PRs. At least test it, please!

### DIFF
--- a/device/globals/dialog_instance.gd
+++ b/device/globals/dialog_instance.gd
@@ -69,6 +69,26 @@ func finish():
 			anim.play("hide")
 	_queue_free()
 
+func clamped_position(dialog_pos):
+	var my_size = $"anchor/text".get_size()
+	var center_offset = my_size.x / 2
+
+	var dist_from_right = vm.camera_limits.size.x - (dialog_pos.x + center_offset)
+	var dist_from_left = dialog_pos.x - center_offset
+	var dist_from_bottom = vm.camera_limits.size.y - (dialog_pos.y + my_size.y)
+	var dist_from_top = dialog_pos.y - my_size.y
+
+	if dist_from_right < 0:
+		dialog_pos.x += dist_from_right
+	if dist_from_left < 0:
+		dialog_pos.x -= dist_from_left
+	if dist_from_bottom < 0:
+		dialog_pos.y += dist_from_bottom
+	if dist_from_top < 0:
+		dialog_pos.y -= dist_from_top
+
+	return dialog_pos
+
 func init(p_params, p_context, p_intro, p_outro):
 	character = vm.get_object(p_params[0])
 	context = p_context
@@ -104,6 +124,8 @@ func init(p_params, p_context, p_intro, p_outro):
 			pos = character.get_node("dialog_pos").get_global_position()
 		else:
 			pos = character.get_position()
+
+		pos = clamped_position(pos)
 		set_position(pos)
 
 	if has_node("anchor/avatars"):
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -31,6 +31,31 @@ var tooltip
 func set_mode(p_mode):
 	mode = p_mode
 
+func tooltip_clamped_position(tt_pos):
+	var vp_size = get_viewport().size
+	var tt_size = tooltip.get_size()
+	tt_pos -= tt_size / Vector2(2, 1)
+
+	# var dist_from_right = vp_size.x - (tt_pos.x + tt_size.x)
+	# var dist_from_left = tt_pos.x - tt_size.x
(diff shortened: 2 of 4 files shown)


device/globals/dialog_instance.gd
```diff
@@ -69,6 +69,27 @@ func finish():
 			anim.play("hide")
 	_queue_free()
 
+func check_clamp(dialog_pos):
+	var rtl_size = $"anchor/text".get_size()
+	var rtl_offset = rtl_size / Vector2(2, 2)
```
fleskesvor: This can be written as `var rtl_offset = rtl_size / 2`, I think.

device/globals/dialog_instance.gd
```diff
@@ -69,6 +69,27 @@ func finish():
 			anim.play("hide")
 	_queue_free()
 
+func check_clamp(dialog_pos):
```
fleskesvor: Maybe `clamped_position` would be a better name, since it's not a void function?

fleskesvor: Or maybe `clamp_position` and update it in place.

device/globals/game.gd
```diff
@@ -26,14 +26,44 @@ var tooltip
 func set_mode(p_mode):
 	mode = p_mode
 
+func tooltip_check_clamp(tt_pos):
```
fleskesvor: Same. Might not need calling if there's no tooltip either.

device/globals/game.gd
```diff
@@ -26,14 +26,44 @@ var tooltip
...
 func mouse_enter(obj):
 	var text
 	var tt = obj.get_tooltip()
 
 	# When following the mouse, prevent text from flashing for a moment in the wrong place
 	if ProjectSettings.get_setting("escoria/ui/tooltip_follows_mouse"):
 		var pos = get_viewport().get_mouse_position()
-		pos -= tooltip.get_size() / Vector2(2, 1)
+
+		pos = tooltip_check_clamp(pos)
```
fleskesvor: Here, I mean. Might be a tad bit easier to follow if the presence of tooltip can be checked here.

device/globals/game.gd
```diff
@@ -216,7 +246,7 @@ func spawn_action_menu(obj):
 		player.walk_stop(player.position)
 
 	var pos = get_viewport().get_mouse_position()
-	var am_pos = action_menu.check_clamp(pos, camera)
+	var am_pos = action_menu.check_clamp(pos)
```
fleskesvor: But it looks like we're already using this confusing naming convention. It's understandable from context though.

device/globals/game.gd
mjtorn: Let's go with `tooltip_clamped_position`.

device/globals/game.gd
mjtorn: I'll move the check upward, so it's 'if tooltip and ProjectSettings....`

mjtorn:
This is _almost_ there, but fortunately I tested setting the camera to the right edge of the screen with the player on the left, and the dialog is not clamped. This is because I'm testing against the background, when I used to be testing against the viewport, but now it seems some kind of best-of-both-worlds approach is required...

mjtorn:
I've given this more thought and toyed around with code between the day's other work.

It's very hard to determine where the viewport is located on the background. This is something I seem to recall we struggled with when doing the original zoom/scale implementation. What this means is it's quite hard to determine when the dialog is outside the viewport, ie. when an NPC or the player is near the edge and speaking.

It would make sense to merge this and look at viewport clamping separately, so we don't forget that this already is a definite improvement over the current state.

@fleskesvor do you concur?

mjtorn:
This uses the camera limits now, as discussed on IRC, and I fixed the commit message to match.

device/globals/dialog_instance.gd
```diff
@@ -69,6 +71,26 @@ func finish():
...
+		dialog_pos.x += dist_from_right
+	if dist_from_left < 0:
+		dialog_pos.x -= dist_from_left
+	if dist_from_bottom < 0:
+		dialog_pos.y += dist_from_bottom
+	if dist_from_top < 0:
+		dialog_pos.y -= dist_from_top
+
+	return dialog_pos
+
```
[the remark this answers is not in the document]
mjtorn: I wish they could, but I'm not so certain because the action menu should be centered on the cursor click whereas dialog requires only the x-axis to be accounted for.

Tooltips are also horribly broken until there's a way to get the text-width size of the tooltip RichTextLabel so they can't be clamped except vertically.

So hopefully in the future!

PR #163, opened 2018-05-15
ESC function to toggle HUD visibility

mjtorn:
If your inventory is overlaid onto the game, and you want to
do camera-movement trickery, it's hard to set the scene so
the inventory doesn't get in the way.

It's also confusing for the brain because having an interactive
element feels like the game would be interactive during the
camera movement.

### DIFF
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -191,6 +191,9 @@ func inventory_set(p_obj, p_has):
 func say(params, level):
 	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "dialog", "say", params, level)
 
+func set_hud_visible(p_visible):
+	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "hud", "set_visible", p_visible)
+
 func dialog_config(params):
 	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "dialog", "config", params)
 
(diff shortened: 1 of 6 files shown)


device/globals/global_vm.gd
```diff
@@ -191,6 +191,9 @@ func inventory_set(p_obj, p_has):
 func say(params, level):
 	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "dialog", "say", params, level)
 
+func set_hud_visible(is_visible):
+	get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "set_hud_visible", is_visible)
```
fleskesvor: Instead of storing a reference to the hud and calling this additional function, you could just call `get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "menu_closed")` (or "menu_opened") instead? Currently I think only the hud will be triggered by that call, but it could be used by other ui elements which need to close programmatically in the future.

fleskesvor: The HUD has its own group too, so I think a clean solution would be to add another function, eg. `toggle_visibility(p_visible)` to hud.gd which takes a boolean to show/hide and set the `visible` property when called via group call.

mjtorn:
I implemented this as @fleskesvor said, and it is much cleaner now. Only change from that comment is that I called the function in hud.gd `set_visible(p_visible)` to match #164.

Also when something is called "toggle", it reminds me of a function taking no arguments, simply toggling an internal state to what it wasn't. Explicit for the win!

StraToN:
Ok for me since @fleskesvor 's change request was fulfilled.
Also we'll need to document this new function in the documentation.

mjtorn:
@StraToN there's a note in the ESC reference now. Hope it's enough :)

StraToN:
Perfect ! Thanks @mjtorn !

Issue #41, opened 2017-12-19
Dialog options' areas are not aligned with labels


StraToN:
We need to check dialogs options generation, it possibly shows a bug.
https://github.com/flossmanualsfr/escoria/issues/14

fleskesvor:
Fixed by #166. I will create a PR on Daïza to address the issue as soon as possible.

PR #166, opened 2018-05-15
Improve dialog trees, documentation and default font

### LINKED ISSUE #40: cyrillic symbols in esc files
I want to use cyrillic symbols in esc files but when I run scene I can not see them.

fleskesvor:
Adds font with better [Unicode support](https://developers.googleblog.com/2016/10/an-open-source-font-system-for-everyone.html), which finally fixes #40.

I'm not 100% sure how to best respect the license (SIL Open Font License), but I included the original license file and readme in a separate font directory and put a note in our readme. Do you know if this is the best way, @StraToN, or should we ask Akien?


mjtorn:
Tested in our jam game, it works with colors and shadow-as-outline.

IIRC the .tres files are project-specific, but I always find myself searching
for https://godotengine.org/qa/20870/how-do-you-bring-fonts-into-godot-3-0

Should that maybe be pointed out or have a pointer to the official docs
if they explain the thing?

Either way, good work, much appreciated!

fleskesvor:
> Tested in our jam game, it works with colors and shadow-as-outline.

Thanks for testing. :slightly_smiling_face: 

> IIRC the .tres files are project-specific, but I always find myself searching
for https://godotengine.org/qa/20870/how-do-you-bring-fonts-into-godot-3-0
>
> Should that maybe be pointed out or have a pointer to the official docs

Good point. I will add a small doc on using and customizing dialogue, including fonts.

fleskesvor:
Thanks guys. ❤️ 

My second commit should address #41. It's not perfect, but it's much better than what we have.

fleskesvor:
I moved the section on speech dialog into the new document on dialogs, and adapted it to not repeat the things from the previous section. Anyone want to look over the changes before it's merged?

PR #167, opened 2018-05-17
Make set_state bg_music accept decibel values for AudioStreamPlayer

mjtorn:
This is a real requirement, especially for remote teams busy with an adventure jam competition. It's easiest to figure out the decibel values when putting the game together than nagging the musician about mixing.

Besides this, we should have a mechanism to alter the volume during the game, like (at least) Broken Age does, so that when a player speaks, background music and sound effects get lowered. But this is a good start!


device/globals/bg_music.gd
```diff
@@ -7,8 +7,8 @@ func game_cleared():
 	if global_id != "":
 		vm.register_object(global_id, self)
 
-func set_state(p_state, p_force = false):
-
+func set_state(p_state, p_volume = 1, p_force = false):
```
fleskesvor: As you found out and we discussed on IRC, this second parameter to `set_state` is problematic, because the state is not saved. I think the idiomatic, and most pragmatic, solution might be to use `set_global` to set volume and add a listener for the `global_changed` event, which you'll find examples of in eg. `item.gd` and `inventory.gd`.

mjtorn:
FWIW I dropped this code entirely from my adventure jam branch as it's too problematic.

One of these time units, I'll write up an issue to replace this PR.

fleskesvor:
Ok. :+1: I'm closing this one then.

PR #168, opened 2018-05-17
Retrieve bg_music from the tree to make music work

mjtorn:
Because otherwise it's loaded way late, so even `:ready`
won't work with it.

This code needs some TLC later.


fleskesvor:
Fixed by #170.

PR #169, opened 2018-05-17
Allow textured content in confirmation popup

mjtorn:
Sometimes you want to use a `TextureRect` in your
confirmation popups, like if the artist wanted
to add decorations to the text.

This makes removing the `message` node possible
and replacing it with nodes named after the translation
identifiers.


StraToN:
Seems ok to me :) Thanks a lot @mjtorn !

PR #171, opened 2018-05-18
Allow Position2D to be item without errors about signals

mjtorn:


fleskesvor:
I think you mentioned on IRC that this is for use with `camera_set_target`. Is that correct? If so, it might be better to add another extension of `interactive.gd`, maybe `target.gd`, which only calls `vm.register_object(global_id, self)` on ready, given that a `global_id` has been set. If it's not to be used as an item, I don't think it should be implemented as one.

mjtorn:
>I think you mentioned on IRC that this is for use with `camera_set_target`.
>Is that correct?

That and similar cases, yes.

> If so, it might be better to add another extension of
>`interactive.gd`, maybe `target.gd`, which only calls
>`vm.register_object(global_id, self)` on ready, given that a `global_id` has
>been set. If it's not to be used as an item, I don't think it should be
>implemented as one.

This is a much cleaner idea, and really appreciate that you're looking
into these PRs, because I obviously haven't had time to really do anything
except throw ideas out there.

Really like the name of `target.gd`. Wonder if it should have that
`interact_angle` or whatever `export var` as well?

One use case apart from `camera_set_target` is to teleport the player near
the exit and it might seem silly if he always faces the camera, although
the door would be to the side.

Do you know if that variable would work for this or should there be something
else? Because the target isn't really interacted with. Or maybe an ESC
command to turn the player.

fleskesvor:
> This is a much cleaner idea, and really appreciate that you're looking
into these PRs, because I obviously haven't had time to really do anything
except throw ideas out there.

No, worries. The ideas are appreciated. :slightly_smiling_face: 

> Really like the name of `target.gd`. Wonder if it should have that
`interact_angle` or whatever `export var` as well?
> 
> One use case apart from `camera_set_target` is to teleport the player near
the exit and it might seem silly if he always faces the camera, although
the door would be to the side.
>
> Do you know if that variable would work for this or should there be something
else? Because the target isn't really interacted with. Or maybe an ESC
command to turn the player.

Yes, that's a good idea, and it should work, because the `interact_angle` is checked when teleporting, and is used to set the right idle frame:
  
https://github.com/godotengine/escoria/blob/master/device/globals/player.gd#L295

We should probably also update `trigger.gd` to use `teleport` instead of `set_position` to make use of this when double-clicking.

mjtorn:
>> Do you know if that variable would work for this or should there be something
>>else? Because the target isn't really interacted with. Or maybe an ESC
>>command to turn the player.
>
>Yes, that's a good idea, and it should work, because the `interact_angle` is checked when teleporting, and is used to set the right idle frame:

>https://github.com/godotengine/escoria/blob/master/device/globals/player.gd#L295
>
>We should probably also update `trigger.gd` to use `teleport` instead of `set_position` to make use of this when double-clicking.

Hmm! I think we cut out a piece of gameplay from the jam game because of this.

The player is half-way off-screen in a small terrain, but the exit has to be
in view. I would now guess that `set_position` respects terrains, whereas
`teleport` may not. Did I get it? We worked around the issue due to time
constraints.

-- 
mjt

fleskesvor:
`set_position` is a function in `Node2D`, so it shouldn't care about the terrain, but there's probably some custom logic that constrains the player to the terrain, which would probably be a problem for `teleport` as well. Should a player technically be able to go somewhere where there isn't a terrain though? My intuition says this should probably be solved by using multiple `NavigationPolygon`s if needed, but this might not work out of the box because of checking for polygons by node name. I will have to investigate.

mjtorn:
#181 can replace this PR, I would then adapt our jam game code for it.

fleskesvor:
Ok, closing this one too. :+1:

PR #173, opened 2018-05-20
[WIP] Make kind of functional web build

fleskesvor:
This is nowhere near ready to be merged, but it makes it possible to walk around in the first demo room and look at the errors that are thrown in the web console.

The web build apparently is a debug build by default, which means the `test_width` and `test_height` properties are used to set the size of the viewport, not `width` and `height`.

I have no idea why commenting out the `start()` function in `save_data.gd` makes the web build not halt, or why commenting out the call to it makes it halt. Doesn't make a whole lot of sense to me.

I also haven't been able to work around an issue where the game crashes when clicking outside of the canvas used to display the game or figured out why the hud isn't visible.

### DIFF
--- a/device/export_presets.cfg
+++ b/device/export_presets.cfg
@@ -1,0 +1,20 @@
+[preset.0]
+
+name="HTML5"
+platform="HTML5"
+runnable=true
+custom_features=""
+export_filter="all_resources"
+include_filter="*.esc"
+exclude_filter=""
+patch_list=PoolStringArray(  )
+
+[preset.0.options]
+
+texture_format/s3tc=true
+texture_format/etc=false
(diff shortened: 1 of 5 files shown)


Beuc:
Currently a major blocker is the use of threads in device/globals/resource_queue.gd which are [not supported](https://docs.godotengine.org/en/latest/getting_started/workflow/export/exporting_for_web.html#limitations) due to [browser limitations](https://emscripten.org/docs/porting/pthreads.html). We can either wait a few years or rework resource management without threads.

StraToN:
Not relevant after switch to new escoria in branch "main".
Web export to be retested though. Thanks!

PR #174, opened 2018-05-24
Fix hud z index

mjtorn:
Our Jam Game had a lot of z-index issues in the HUD. Tooltips and dialog were positioned underneath the player and other objects.

The reverted commit should never have happened. It made sense at the time, because it wasn't apparent in any known demo or proto. The positioning is off regardless of the change, at least in our Jam Game, so reverting it has a negligible effect at best for positioning and makes the text appear on top, fixing a very relevant regression.

Tooltips were mostly correct for items, but broke down hard with the inventory, which uses `Node2D`. Having a layer node for it makes it possible to have the tooltips where you'd expect them: readable on top of the item.

I don't think there's much to discuss on this one, please kick it in.

### DIFF
--- a/device/demo/ui/dialog_default.tscn
+++ b/device/demo/ui/dialog_default.tscn
@@ -4,7 +4,7 @@
 [ext_resource path="res://demo/ui/dialog_theme.tres" type="Theme" id=2]
 
 
-[node name="dialog_instance" type="Container"]
+[node name="dialog_instance" type="Node2D"]
 
 anchor_left = 0.0
 anchor_top = 0.0
@@ -24,7 +24,7 @@ typewriter_text = true
 characters_per_second = 45.0
 fixed_pos = false
 
-[node name="anchor" type="Container" parent="." index="0"]
+[node name="anchor" type="Node2D" parent="." index="0"]
 
 anchor_left = 0.0
 anchor_top = 0.0
--- a/device/globals/dialog_instance.gd
+++ b/device/globals/dialog_instance.gd
@@ -1,4 +1,4 @@
-extends Container
+extends Node2D
 
 var context
 var text
--- a/device/globals/dialog_player.gd
+++ b/device/globals/dialog_player.gd
@@ -10,6 +10,7 @@ func say(params, callback):
 		type = params[2]
 	type = type + ProjectSettings.get_setting("escoria/platform/dialog_type_suffix")
 	var inst = get_resource(type).instance()
+	var z = inst.get_z_index()
 	get_node("/root").get_child(0).add_child(inst)
 	var intro = true
 	var outro = true
@@ -17,6 +18,7 @@ func say(params, callback):
 		intro = types[type][0]
 		outro = types[type][1]
 	inst.init(params, callback, intro, outro)
+	inst.set_z_index(z)
 
 func config(params):
 	types[params[0]] = [params[1], params[2]]
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -441,7 +441,7 @@ func load_hud():
 	#else:
 	#	get_node("hud_layer/hud/inv_toggle").hide()
 
-	tooltip = get_node("hud_layer/hud/tooltip")
+	tooltip = get_node("hud_layer/hud/tt_layer/tooltip")
 
 func _ready():
 	add_to_group("game")
--- a/device/globals/hud.gd
+++ b/device/globals/hud.gd
@@ -5,7 +5,7 @@ var background = null
 func set_tooltip(text):
 	if text:
 		printt("hud got tooltip text ", text)
-	get_node("tooltip").set_text(text)
+	get_node("tt_layer/tooltip").set_text(text)
 
 func inv_toggle():
 	#get_node("inventory").toggle()
@@ -48,3 +48,4 @@ func _ready():
 	# Hide verb menu if hud layer has an action menu
 	if has_node("../action_menu"):
 		$verb_menu.hide()
+
--- a/device/ui/hud.tscn
+++ b/device/ui/hud.tscn
@@ -15,11 +15,16 @@ margin_bottom = 40.0
 rect_pivot_offset = Vector2( 0, 0 )
 rect_clip_content = false
 mouse_filter = 0
+mouse_default_cursor_shape = 0
 size_flags_horizontal = 2
 size_flags_vertical = 2
 script = ExtResource( 1 )
 
-[node name="tooltip" type="Label" parent="."]
+[node name="tt_layer" type="Node2D" parent="." index="0"]
+
+z_index = 2
+
+[node name="tooltip" type="Label" parent="tt_layer" index="0"]
 
 anchor_left = 0.0
 anchor_top = 0.0
@@ -32,6 +37,7 @@ margin_bottom = 115.0
 rect_pivot_offset = Vector2( 0, 0 )
 rect_clip_content = false
 mouse_filter = 2
+mouse_default_cursor_shape = 0
 size_flags_horizontal = 2
 size_flags_vertical = 0
 custom_colors/font_color = Color( 0, 0, 0, 1 )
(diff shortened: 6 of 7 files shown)


fleskesvor:
Is the additional "layer" needed? I think the root issue is that the hud layer isn't drawn on top of the viewport, so this might potentially cause similar issues if that isn't fixed. Also, have you made sure that your game node is at the bottom of the room scene? Remember that draw order is from root to bottom unless overridden with z index.

mjtorn:
I don't really know about drawing onto the viewport, this just seemed like a legitimate fix. Can you tell me more about what that implies? If the HUD was drawn onto the viewport, it would always - by definition - be top-most?

My `game` node is at the bottom, that's not the problem here.

fleskesvor:
I'm not awfully familiar with it myself, but my understanding of CanvasLayers is that all its nodes should be on top of nodes from a lower layer, so that you don't have to worry about individual z indices. This doesn't seem to be working, and I think we should fix that so that a developer doesn't have to know about the z indices in the framework to position tveit nodes correctly.

mjtorn:
You'd think that, I'd think that, but that's not what we saw.

I'll look into reproducing it in the demo at some point, but if you have access to a Linux machine, probably something like VirtualBox may suffice, you can download versions 0.1.0 and 0.1.4 at https://gamejolt.com/games/ironsky-lunar-adventure/339236 to see how the tooltips and dialog drop below everything else in 0.1.0.

In defense of this code, the developer shouldn't have to know about Z, because they're relative to the parent. This pushes them up by definition because they're related to the last node (`game`) in the tree.

fleskesvor:
> You'd think that, I'd think that, but that's not what we saw.

Yes, but I'm thinking that's probably a bug in Escoria. My main concern with adding a quick fix for tooltips is that it's only addressing one visible symptom and might require additional quick fixes for other UI elements in the future. If it isn't urgent, I prefer to take some extra time to investigate what's causing this and come up with an idiomatic solution.

> I'll look into reproducing it in the demo at some point, but if you have access to a Linux machine, probably something like VirtualBox may suffice, you can download versions 0.1.0 and 0.1.4 at https://gamejolt.com/games/ironsky-lunar-adventure/339236 to see how the tooltips and dialog drop below everything else in 0.1.0.

My main computer is a Linux machine, but unfortunately I had to send it in to service a few weeks ago and I've not been able to play the demo on my old Linux machine because of dependency issues (libssl and libcrypto). I will give it a try on the work Mac and try to reproduce in the demo.

> In defense of this code, the developer shouldn't have to know about Z, because they're relative to the parent. This pushes them up by definition because they're related to the last node (`game`) in the tree.

They might have to adjust z indices to arrange sprites correctly in their game scenes, but it's up to us to make sure the hud is always on top. If this can't be solved without setting a very high z index on hud elements, developers need to know which z index we've chosen for those elements, so that they can make sure to use a lower index on their other sprites.

mjtorn:
Hope you get your computer back soon! My work laptop may face a similar situation soon too, it runs real hot. Fortunately I don't rely on it too much.

And I very much doubt they'd adjust the z-indices, because they're relative. Using a scalemap on a terrain sets crazy z-index values on the player, like _1500_, which changes as the player moves. Yet still the tooltips work as expected with a more reasonable value.

mjtorn:
As stated on IRC, I tried moving the camera to be a child of `hud_layer` to see if that would make the HUD elements always the most high. Naturally I also removed the `tt_layer` `Node2D`.

The tooltips were rendering beneath everything else.

I had the idea of making `camera` a child of `hud_layer/hud` as well, but this broke everything with an error about a previously freed resource - even if I added a camera to the hud scene itself.

So simply having a Node2D to help out with the Control nodes feels like an even better idea now that I tried the alternatives :/

fleskesvor:
Just to clarify, are both of these statements accurate for the current state of Escoria?

> Our Jam Game had a lot of z-index issues in the HUD. Tooltips and dialog were positioned underneath the player and other objects.

> Tooltips were mostly correct for items, but broke down hard with the inventory, which uses `Node2D`.

The first one is what has me the most worried though, since according to the docs, http://docs.godotengine.org/en/3.0/classes/class_canvaslayer.html and http://docs.godotengine.org/en/3.0/classes/class_canvaslayer.html, everything in the `hud_layer` in `game.tscn` and `game_am.tscn` should render above anything in the "default 2D scene" , since the former is in layer 1 and the latter is in layer 0.

The second one is expected, since the `tooltip` is higher up than the `inventory` in `hud.tscn`. Placing `tooltip` at the bottom should fix this, but I'll need to verify once I've fixed the demo to be able to reproduce the issue. Though come to think of it, isn't it expected that the inventory should prevent interaction with items beneath it, even mouse over events?

Is this PR meant to address both issues or only the second?

mjtorn:
On Mon, May 28, 2018 at 12:58:11PM -0700, Eivind wrote:
>Just to clarify, are both of these statements accurate for the current state of Escoria?
>
>> Our Jam Game had a lot of z-index issues in the HUD. Tooltips and dialog were positioned underneath the player and other objects.
>
>The first one is what has me the most worried though, since according to the docs, http://docs.godotengine.org/en/3.0/classes/class_canvaslayer.html and http://docs.godotengine.org/en/3.0/classes/class_canvaslayer.html, everything in the `hud_layer` in `game.tscn` and `game_am.tscn` should render above anything in the "default 2D scene" , since the former is in layer 1 and the latter is in layer 0.

You can play version 0.1.0 and see the breakage. The NPC dialog in the first
room goes underneath the player.

Tooltips also, iirc. Actually I'm too tired to remember exactly in which
instaces they did...

>> Tooltips were mostly correct for items, but broke down hard with the inventory, which uses `Node2D`.
>
>The second one is expected, since the `tooltip` is higher up than the `inventory` in `hud.tscn`. Placing `tooltip` at the bottom should fix this, but I'll need to verify once I've fixed the demo to be able to reproduce the issue. Though come to think of it, isn't it expected that the inventory should prevent interaction with items beneath it, even mouse over events?

...But the inventory is surely one of them.

An inventory should of course prevent interaction, but now I don't have the
game at hand and it's really too late to be awake and try to remember
if it's possible to interact "through" the inventory...

I think the inventory background had a z-index, as did the items, but not
the tooltips. You may be right in that the tooltip should be lower
in the tree, but I think it's actually in a different branch of the scene
tree. Maybe that's not a problem, I'd need the code and the editor to see.

You tell me when you're done ;)

>Is this PR meant to address both issues or only the second?

It fixed everything, as you may see in v0.1.4 of the game. Both
tooltips and dialog.

-- 
mjt

fleskesvor:
I created a new HUD branch (#180) which demonstrates the issue with the tooltips appearing behind the inventory when hovering over an item in the inventory with `escoria/ui/tooltip_follows_mouse` set to `true` in the project settings. Placing the `tooltip` node at the bottom of `hud.tscn` fixes that issue.

The `dialog_player` in `game.tscn`/`game_am.tscn` actually isn't in a `CanvasLayer` at all, so that explains why it's not working as expected. The `dd_player` in `main.tscn`, however, is. I wonder what the rationale for that is and also why both dialog players aren't just in `main.tscn` to begin with.

My recommendations would be to move the `tooltip` node to the bottom of `hud.tscn` and move `dialog_player` to `main.tscn`, which should hopefully address these issues. What do you think, @StraToN?

> An inventory should of course prevent interaction, but now I don't have the
game at hand and it's really too late to be awake and try to remember
if it's possible to interact "through" the inventory...

I checked, and the inventory blocks interaction with the background, so that's working as expected.

StraToN:
@mjtorn Can you please test if your problem still occurs in current master please?

fleskesvor:
> @mjtorn Can you please test if your problem still occurs in current master please?

I don't think we did anything to fix this yesterday, but I believe this will:

> move the `tooltip` node to the bottom of `hud.tscn` and move `dialog_player` to the `dialog` `CanvasLayer` in `main.tscn`

( ^ edited from original quote to clarify)

fleskesvor:
@mjtorn: Can you confirm that these issues are fixed by #180?

mjtorn:
On Fri, Jun 15, 2018 at 02:02:46PM -0700, Eivind wrote:
>@mjtorn: Can you confirm that these issues are fixed by #180? 

I'll attempt to do so on Monday. I'm sort of neck-deep in making
the Spine integrator script that would work with Godot in general
and not require Escoria to hack it. The stuff I've been talking
about on IRC ;P

So much to do, so little time...

-- 
mjt

fleskesvor:
That's OK. No rush. :slightly_smiling_face:

mjtorn:
I think #180 was covered in a recent rebase because I had a ton of pain resetting everything related to the tooltip etc :D Also thought this PR was closed already when I deleted my branch, so all is ok for now.

PR #175, opened 2018-05-24
Do not clear params_queue until walk_stop()

mjtorn:
There were cases where the params_queue was cleared, with the side
effect that nothing happened until opening the action menu again.

@fleskesvor suggested this has been a problem with the verb menu
as well.

My investigations lead me to believe it has to do with disjointed
terrains or other unwalkable destinations that caused the queue
reset to happen too soon.

This should also be a somewhat straight-forward merge.

### DIFF
--- a/device/globals/player.gd
+++ b/device/globals/player.gd
@@ -39,7 +39,6 @@ func walk_to(pos, context = null):
 	walk_context = context
 	if walk_path.size() == 0:
 		task = null
-		params_queue = null
 		walk_stop(get_position())
 		set_process(false)
 		return
@@ -49,7 +48,6 @@ func walk_to(pos, context = null):
 	path_ofs = 0.0
 	task = "walk"
 	set_process(true)
-	params_queue = null
 
 func walk(pos, speed, context = null):
 	walk_to(pos, context)
@@ -148,8 +146,10 @@ func interact(p_params):
 	else:
 		pos = p_params[0].get_global_position()
 	if !telekinetic && get_global_position().distance_to(pos) > 10:
-		walk_to(pos)
+		# It's important to set the queue before walking, so it
+		# is in effect until walk_stop() has to reset the queue.
 		params_queue = p_params
+		walk_to(pos)
 	else:
 		if animations.dir_angles.size() > 0 && p_params[0].interact_angle != -1:
 			last_dir = _get_dir_deg(p_params[0].interact_angle)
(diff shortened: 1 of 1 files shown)


mjtorn:
The bug is that an unreachable item cannot be interacted with properly. Using the action menu will not do anything, until the action menu is opened again on that or any other item.

This should also be evident in 0.1.0 of https://gamejolt.com/games/ironsky-lunar-adventure/339236 when you are able to enter the third room (not spoiling more of it!)

One of the NPCs gets teleported into a disjoint segment of the terrain, which causes the bug.

Because the player will always attempt to walk to, or as close to, the item/npc as possible, in this case he cannot. There is no path to a disjoint segment, by definition. This is one of the code paths: https://github.com/godotengine/escoria/pull/175/commits/23105307fd8cfe828d39010bda4a46dba59a1a5f#diff-e7397f5f52cd6ad0a6dba10a0be9c680L40

So when `walk_to(pos)` was called first, it cleared the `params_queue` variable, and caused nothing to happen when we reach `walk_stop(pos)`.

This code removes the clearing, and makes sure the `params_queue` is in effect even if the player can't reach the item/npc. It's so important, I wrote a comment about it in the code.

The solution is ok, because `walk_stop(pos)` does this: https://github.com/godotengine/escoria/commit/23105307fd8cfe828d39010bda4a46dba59a1a5f#diff-e7397f5f52cd6ad0a6dba10a0be9c680L169

The `params_queue` variable is either there so `interact` can be called and _then_ the queue cleared, or it's null and nothing needs to happen.

----

I do not know where the action came from when opening the action menu again in the original code. I simply couldn't figure it out, nor is it very relevant, because this logic is clearer: "If told to do something, do it, even if you can't reach the destination".

mjtorn:
```
mjt@lechuck:~/src/git_checkouts/isala-escoria (isala *%=)$ git grepn params_queue
device/globals/player.gd:22:var params_queue
device/globals/player.gd:171:           params_queue = p_params
device/globals/player.gd:190:   if typeof(params_queue) != typeof(null):
device/globals/player.gd:191:           if animations.dir_angles.size() > 0 && params_queue[0].interact_angle != -1:
device/globals/player.gd:192:                   last_dir = _get_dir_deg(params_queue[0].interact_angle)
device/globals/player.gd:199:           get_tree().call_group_flags(SceneTree.GROUP_CALL_DEFAULT, "game", "interact", params_queue)
device/globals/player.gd:201:           params_queue = null
```

I don't think it can be anything but null. Should there be something around L190 that says `if !params_queue:` and `assert(params == null)` without that silly `typeof` overhead?

When I get back to coding, I could try that.

fleskesvor:
Yeah, I think just changing it to check that it's not null is good enough, and it makes it much easier to tell what's going on with params_queue.

mjtorn:
I simplified the `!= null` check

StraToN:
Seems good to merge. Thanks @mjtorn !

PR #178, opened 2018-05-28
Adaptible RMB behavior, with documentation

mjtorn:
This allows `game` to have a child node, basically just for the class/self behavior, that enables adaptible right-mouse-button behavior.

With this merged, you can alter the behavior of the RMB. This is especially useful on `background`, where you may want to highlight items and/or disable the current tool.

After 2.5 hours of trying to figure out what the best design is, I'm fairly confident it's this.


StraToN:
Perfect! Thanks a lot @mjtorn :)

fleskesvor:
This is great stuff. :+1: It adds an idiomatic way of writing custom functionality in GDScript, which is something we use a lot in The Farrier, but this is a better way of doing it, since as a child of `game.tscn`, the same node and script can be re-used in multiple rooms.

PR #180, opened 2018-05-28
HUD positioning

fleskesvor:
Anchor verb menu and inventory to the bottom left and right, respectively, to make the HUD less dependent on screen sizes.

The inventory takes up too much screen space in room2 and needs adjusting.

### DIFF
--- a/device/demo/ui/inventory.tscn
+++ b/device/demo/ui/inventory.tscn
@@ -1,22 +1,30 @@
-[gd_scene load_steps=3 format=2]
+[gd_scene load_steps=4 format=2]
 
-[ext_resource path="res://globals/inventory.gd" type="Script" id=1]
-[ext_resource path="res://demo/ui/inventory_items.tscn" type="PackedScene" id=2]
+[ext_resource path="res://globals/white.png" type="Texture" id=1]
+[ext_resource path="res://globals/inventory.gd" type="Script" id=2]
+[ext_resource path="res://demo/ui/inventory_items.tscn" type="PackedScene" id=3]
 
-[node name="inventory" type="Control" index="0"]
+[node name="inventory" type="TextureRect" index="0"]
 
+self_modulate = Color( 0, 0, 0, 0.196078 )
 anchor_left = 0.0
 anchor_top = 0.0
 anchor_right = 0.0
 anchor_bottom = 0.0
-margin_right = 40.0
-margin_bottom = 40.0
+margin_right = 600.0
+margin_bottom = 280.0
 rect_pivot_offset = Vector2( 0, 0 )
+rect_clip_content = false
 mouse_filter = 0
 mouse_default_cursor_shape = 0
 size_flags_horizontal = 2
 size_flags_vertical = 2
-script = ExtResource( 1 )
+texture = ExtResource( 1 )
+expand = true
+stretch_mode = 0
+script = ExtResource( 2 )
+_sections_unfolded = [ "Visibility" ]
+is_collapsible = false
 
 [node name="slots" type="Control" parent="." index="0"]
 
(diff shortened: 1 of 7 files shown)


fleskesvor:
The issue with dialog appearing behind items should now be fixed by moving `dialog_player` into the `dialog` `CanvasLayer` in `main.tscn`, instead of its previous placement in `game.tscn`/`game_am.tscn`, where it was not in a `CanvasLayer`. It could have been placed within a `CanvasLayer` in `game.tscn`/`game_am.tscn` too, but I felt it made sense to have both kinds of dialog be in the same node, and dialogs are already modular by being configurable within their respective `ResourcePreloaders`.

The issue with tooltips on inventory items appearing behind the inventory should be fixed by moving `tooltip` further down the scene tree in `hud.tscn`.

mjtorn:
I tested this, and the HUD has dimensions 1920x0, so the tooltips aren't visible. When enlarging the HUD the diff looks like this:

```diff
diff --git a/device/globals/game.tscn b/device/globals/game.tscn
index 6bb43ba..0434ff9 100644
--- a/device/globals/game.tscn
+++ b/device/globals/game.tscn
@@ -2,7 +2,7 @@
 
 [ext_resource path="res://globals/game.gd" type="Script" id=1]
 
-[node name="game" type="Node" index="0"]
+[node name="game" type="Node"]
 
 script = ExtResource( 1 )
 fallbacks_path = "res://demo/fallbacks.esc"
@@ -19,6 +19,9 @@ transform = Transform2D( 1, 0, 0, 1, 0, 0 )
 
 [node name="hud" parent="hud_layer" index="0" instance_placeholder="res://ui/hud.tscn"]
 
+margin_top = -1080.0
+_sections_unfolded = [ "Rect" ]
+
 [node name="wait_timer" type="Timer" parent="." index="1"]
 
 process_mode = 1
```

Although position became (0, 0) and the size 1920x1080.

With tooltip_follows_mouse the tooltip does indeed get rendered underneath the inventory items. Dragging the tooltip to the bottom of the node tree, it would appear items are rendered behind the semi-transparent background of the inventory area.

So I'd say this solution may be in the correct direction but does not work now.

fleskesvor:
> I tested this, and the HUD has dimensions 1920x0, so the tooltips aren't visible.

Yeah, I remember this now. I had initially made it span the entire screen and set `mouse/filter` to `IGNORE`, which works nicely in-game, however it makes moving items around in the editor impossible, without first hiding the hud. I then decided to just make it span the bottom of the screen, since that works nicely for positioning the verb menu and the inventory, but I forgot about tooltips.

I'll see about just using a `CanvasLayer` instead.

fleskesvor:
> it would appear items are rendered behind the semi-transparent background of the inventory area

Yeah, I thought so too, but I knew it wasn't the case, since I could interact with them. I mentioned on IRC about a week ago that I was wondering if this was some kind of optical illusion, but I didn't check up on it. Turns out the actual issue was that I was modulating both the inventory and its children, so the items just blended in with the background. I've fixed the problem in my last commit, which I plan on adding into the previous commits where appropriate.

The last commit also fixes tooltips by going back to the previous solution of making the container fill the entire viewport. It's set to hidden so that it doesn't cover nodes in the editor, but is set to visible on _ready. I don't know how I feel about that solution, but at least it makes it testable.

fleskesvor:
The following issues reported on IRC should now be fixed:

- Inventory items blending with inventory: use self-modulate instead of modulate, otherwise all children will also be affected
- Tooltip not visible: use global position because having a bottom anchored hud make tooltip appear below viewport

StraToN:
2 little conflicts before I test this PR asap.

fleskesvor:
@StraToN: I've rebased onto master now, so it's ready for testing. The easiest way to test the layering is to enable `Escoria/Ui/Tooltip Follows Mouse` from the project settings and make sure that text stays on top of everything in the room, including items in the inventory.

We should also look into improving the contrast of the default dialog and tooltips, but that's outside the scope of this PR.

StraToN:
All good for me, tests pass fine.

fleskesvor:
Thanks, @StraToN. :slightly_smiling_face:

PR #181, opened 2018-06-04
Small fixes, target.gd, and turn

mjtorn:
I don't think splitting this up is worth-while as every change is very simple and they're tied together in the end by extending the first demo room.

I found `walk_context` may be missing `"fast"` so that's a trivial check.

@fleskesvor suggested in #171 that targets would only be registered if they have a global id. Thinking about it, that introduces implicit behavior if the global id is forgotten. `register_object()` should be very robust either way, so that is hardened now.

There's a new type for `target` which is essentially a named `Position2D`, useful for moving stuff around the room.

After teleporting or moving something, it may be necessary to turn the object. There's an ESC command for it.

Please review!


docs/esc_reference.md
```diff
@@ -164,6 +164,9 @@ Items can also change state by playing animations from an `AnimationPlayer` name
 - `walk_block object1 object2`
   Moves object1 towards the position of object2, at the speed determined by object1's "speed" property. This command is blocking.
 
+- `turn object degrees`
+  Turns `object` to a `degrees` angle. The `degrees` value must be between [0, 360] or an error is reported.
```
StraToN: Do you mean this command turns the object for the given `degrees` amount? Or does it specify a target "look angle" (0 = front, 90 = east, 180 = turned back...) as defined in the player_animations.gd file?

mjtorn: No, it turns it to a given angle, ie. it specifies the angle.

Let's see if I can make it clearer somehow...

device/globals/global_vm.gd
```diff
@@ -366,6 +366,9 @@ func get_object(name):
 	return objects[name]
 
 func register_object(name, val):
+	if !name:
+		report_errors("register_object", ["Name not given for " + val.get_class()])
```
fleskesvor: Can this happen? All of the calls to `register_object` seems to be guarded by `if global_id != ""` checks.

device/globals/global_vm.gd
```diff
@@ -370,14 +370,17 @@ func register_object(name, val):
...
-	if name in states:
-		val.set_state(states[name])
-	else:
-		val.set_state("default")
-	if name in actives:
-		val.set_active(actives[name])
 	val.connect("tree_exited", self, "object_exit_scene", [name])
 
+	# Most objects have states/animations, but targets do not
+	if not val is esc_type.TARGET:
```
fleskesvor: Isn't it better to stub `set_state` in `target.gd`, as was the case for `player.gd` until recently, in case we want to implement it in the future? I can't think of a reasonable implementation of state for a `Position2D` though.

device/globals/target.gd
```diff
@@ -0,0 +1,10 @@
+extends Position2D
```
fleskesvor: Oh, right. It doesn't extend any of the other scripts. Probably makes less sense to stub `set_state` then.

device/globals/esc_compile.gd
```diff
@@ -25,6 +25,7 @@ var commands = {
 	"teleport_pos": { "min_args": 3 },
 	"walk": { "min_args": 2 },
 	"walk_block": { "min_args": 2 },
+	"turn": { "min_args": 2 },
```
fleskesvor: I think it would be better to name the command (and the corresponding GDScript functions something like `set_angle`/`set_degree`, since `turn`, to me at least, implies that the item/player will always turn when it is called.

device/globals/global_vm.gd
mjtorn: It's more robust this way. I nearly passed in an empty `global_id` and someone else might in the future.

It's also a bit weird that you'd attach `target.gd` to a `Position2D` and wonder why it doesn't work, if you forgot to set the `global_id`. This is where the hardening is useful, to prevent bugs.

I would personally prefer this to be the only check for it and go so far as to remove the `if global_id != ""` checks altogether and receive errors instead. Errors are nice and clearly actionable.

device/globals/global_vm.gd
mjtorn: As per the comment below, there really isn't a need.

device/globals/esc_compile.gd
mjtorn: This goes back to what @StraToN once said about preferring "humane" names. `set_angle` and `set_degree` are very explicit, but `turn` does always turn the item/player just as much as `set_angle` would.

device/globals/player.gd
```diff
@@ -317,6 +317,16 @@ func teleport_pos(x, y):
 	set_position(Vector2(x, y))
 	_update_terrain()
 
+func turn(deg):
+	if deg < 0 or deg > 360:
+		vm.report_errors("player", ["Invalid degree to turn " + String(deg)])
```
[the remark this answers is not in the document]
mjtorn: Circles are divided into 360 degrees, there is no logic to turn the player round and round and round, and -1 is a special case in Escoria. Something like -2 doesn't make any sense, unless there's weird imaginary-number-type things going on.

Again, it's about robust code that tells the developer something's not right. So I say yes.

device/globals/interactive.gd
```diff
@@ -120,6 +120,16 @@ func _process(time):
 	if self is esc_type.ITEM:
 		_update_terrain()
 
+func turn(deg):
+	if deg < 0 or deg > 360:
```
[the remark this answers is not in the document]
mjtorn: As above, yes ;)

device/globals/global_vm.gd
fleskesvor: Yes, I agree. The call to `register_object` in `item.gd` should also be updated. The others (`telon.gd`, `bg_music.gd` and `player.gd`) hardcode the global_id, I think, so there's no need there.

device/globals/esc_compile.gd
fleskesvor: I'm not 100% convinced  this case is equivalent. `turn` doesn't turn the item/player if the player is already facing that direction. It sounds more like a toggle to me, while the others clearly define what will be achieved by running the command. I welcome @StraToN's input though.

fleskesvor: Or even worse, it might lead the developer to think they will turn the item/player by x degrees, which could be useful in some cases.

mjtorn: But... `set_angle` doesn't do anything either if the player is already at that angle, unless the entire implementation is changed. And how would the player turn or have his angle set if he's already there?

If you (try to) open the can in the demo room, you can see that the player is also animated while turning, so there shouldn't be any visual confusion either.

If anything, `set_*` sounds more "toggly" to me.

mjtorn: > Or even worse, it might lead the developer to think they will turn the item/player by x degrees, which could be useful in some cases.

The documentation says otherwise, and if we really find a need for that functionality, it'd be `turn_by` anyway.

device/globals/player.gd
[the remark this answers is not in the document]
fleskesvor: Yes, a circle is divided into 360 degrees, but you can easily turn 1000 or -3000 degrees. This is not so useful now perhaps, but it could be if commands could take globals (which might be very hard to achieve) and a command let you turn an item/player by x degrees.

My question was more about how likely it is that a developer would use an invalid value though, to warrant this kind of check.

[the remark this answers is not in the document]
mjtorn: I concede that you can rotate a dial by however many degrees and represent directions by using negative numbers. This does not, however, exist in Escoria. IMO if `turn_by` gets implemented, it should have this possibility and the animation support for it too.

In this case, where it does effectively "set" an angle (albeit with animations) there's no real reason to allow numbers. And -1 is still a special case.

The developer may put in a typo or have read the documentation a bit sloppily or want to see what happens. Especially junior hackers learn by trying to break the system.

What do you propose as an alternative?

Note also that doing an integer check is probably a CPU cycle's worth of computational power.

device/globals/global_vm.gd
mjtorn: Turns out `bg_music` had it in the scene so I added it there for `telon` as well. Squashed into the commit that was about the item only.

device/globals/global_vm.gd
mjtorn: I remembered seeing `has_method()` so I used that instead of having those types. Quacks like a duck and all that...

mjtorn:
Dropping here what I said on the IRC channel. The code for turning items is broken, which may be a side effect of Spine. I don't know where the code lives for setting the direction based on 1 or -1 in player's `animations` but I'm thinking it must be missing from items. In that case it really isn't Spine.

The code that calculates directions in `item.gd` is almost, but not quite, like the functions in `player.gd`. Optimally this kind of "just enough of the same, yet not the same, to fool the developer" copypasta would be replaced by using methods moved to a static `utils.gd` script.

device/globals/esc_compile.gd
fleskesvor: Who reads documentation though? :stuck_out_tongue: In that case, wouldn't it make sense to have the commands be `turn_to` and `turn_by`?

device/globals/player.gd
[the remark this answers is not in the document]
fleskesvor: I thought only a return value of -1 would be a special case; not input value of -1? I'm not awfully familiar with how this works, but my intuition was that this function would work just fine with any valid input and that if a value of < 0 or > 360 it would be for a reason, though it probably makes no difference at the moment.

mjtorn:
I force-pushed the rename and a refactor to interactive.gd so that the angle-getters correspond better to player.gd and overall have the benefit of working in theory.

The demo room does not have anything to test turning items, but probably should, so let this be a lesson about not cramming too much stuff into one PR "because of course it'll be a breeze" :( Right now there's some kind of blow but it's not breeze.

device/demo/rooms/test/test.tscn
```diff
@@ -12,10 +12,12 @@
 [ext_resource path="res://demo/rooms/test/ice_cream.tscn" type="PackedScene" id=10]
 [ext_resource path="res://demo/rooms/test/box.tscn" type="PackedScene" id=11]
 [ext_resource path="res://demo/rooms/test/can.tscn" type="PackedScene" id=12]
-[ext_resource path="res://demo/rooms/test/chain_saw.tscn" type="PackedScene" id=13]
-[ext_resource path="res://demo/rooms/test/floppy.tscn" type="PackedScene" id=14]
-[ext_resource path="res://globals/game.tscn" type="PackedScene" id=15]
-[ext_resource path="res://demo/ui/rmb_hints.gd" type="Script" id=16]
+[ext_resource path="res://demo/rooms/test/can_anims.gd" type="Script" id=13]
```
fleskesvor: This file is missing from the PR.

fleskesvor: I created the following file for testing, based on your animations, which I think should look similar to your file:

    # angle "0" is facing the camera, rotates counter clock-wise, angle 90 is profile right, etc

    const dir_angles = [ 90, 180, 270, 360 ]

    const directions = ["face_right", 1, "face_up", 1, "face_left", 1, "face_down", 1]

    const idles = ["face_right", 1, "face_up", 1, "face_left", 1, "face_down", 1]

    const speaks = ["face_right", 1, "face_up", 1, "face_left", 1, "face_down", 1]

`idles` and `speaks` is probably not necessary though, but I haven't looked into that.

device/globals/interactive.gd
```diff
@@ -109,17 +111,27 @@ func _process(time):
...
+		vm.report_errors("interactive", ["Invalid degree to turn to " + String(deg)])
+
+	last_dir = _get_dir_deg(deg)
+
+	if animation:
+		if !animation.get_current_animation() or animation.get_current_animation() != animations.directions[last_dir]:
+			animation.play(animations.directions[last_dir])
+
+			set_state(animations.idles[last_dir])
```
fleskesvor: Does this have any effect?

device/globals/player.gd
```diff
@@ -317,6 +317,16 @@ func teleport_pos(x, y):
 	set_position(Vector2(x, y))
 	_update_terrain()
 
+func turn_to(deg):
+	if deg < 0 or deg > 360:
+		vm.report_errors("player", ["Invalid degree to turn to " + String(deg)])
```
fleskesvor: I don't think there's any overhead in using the `String()` constructor, but `str()` might be more idiomatic?

device/globals/interactive.gd
```diff
@@ -109,17 +111,27 @@ func _process(time):
...
 				animation.play(animations.directions[last_dir])
 
-		#pose_scale = animations.directions[last_dir+1]
-
 	if self is esc_type.ITEM:
 		_update_terrain()
 
+func turn_to(deg):
+	if deg < 0 or deg > 360:
+		vm.report_errors("interactive", ["Invalid degree to turn to " + String(deg)])
```
fleskesvor: I don't think there's any overhead in using the `String()` constructor, but `str()` might be more idiomatic?

device/demo/rooms/test/can.esc
```diff
@@ -1,12 +1,46 @@
...
-
 say player "A can"
 
 :use
 set_state player hat
+
+:open
+set_global turning_can true
+> [eq can_angle 180, turning_can]
+	turn_to can 90 [eq can_angle 180]
```
fleskesvor: Are the postfix expressions here and in the "270" and "90" blocks necessary?

device/demo/rooms/test/can.esc
```diff
@@ -1,12 +1,46 @@
...
-
 :look
-
 say player "A can"
 
 :use
 set_state player hat
+
+:open
+set_global turning_can true
```
fleskesvor: What is this used for?

device/demo/rooms/test/can.esc
```diff
@@ -1,12 +1,46 @@
...
+	turn_to can 270
+	set_global can_angle 270
+	debug "!can_angle, now 270"
+	set_global turning_can false
+
+> [eq can_angle 360, turning_can]
+	turn_to can 270
+	set_global can_angle 270
+	debug "can_angle was 360, now 270"
+	set_global turning_can false
```
fleskesvor: Would it be better to write these two blocks as `[!eq can_angle 90, !eq can_angle 180, !eq can_angle 270, turning_can]` instead? We could use `OR` or some kind of switch statement.

device/globals/telon.tscn
```diff
@@ -106,7 +106,7 @@ light_on_map = false
 tooltip = ""
 action = ""
 events_path = ""
-global_id = ""
+global_id = "telon"
```
fleskesvor: Should be safe to remove this from `globals/main.tscn` now, right?

device/demo/rooms/test/test.tscn
mjtorn: The file is force-pushed, though you pretty much got it.

device/globals/player.gd
mjtorn: Fixed.

device/globals/interactive.gd
mjtorn: Fixed.

device/demo/rooms/test/can.esc
mjtorn: They are left-overs from when I thought I could do this easily with postfix expressions. Removed.

device/demo/rooms/test/can.esc
mjtorn: I'll take a look to see if I really need it. I was running into a situation where turning the can evaluated the next block. Naturally you can't call `stop` on it or the player wouldn't move, as `stop` terminates all execution, not just the block.

Any order of the turns will cause two checks of only `[eq can_angle XYZ]` to be true, unless I'm too intellectually challenged to figure out what the order would be. But I'm pretty sure we need this helper boolean. At least it makes the code explicit!

device/demo/rooms/test/can.esc
mjtorn: We don't have an OR right now apart from helper booleans and `stop`, right?

Besides the can is turned to fixed angles, so every "not 90 degrees" would require a "well, what is it, then?" check to see where it goes next.

device/globals/telon.tscn
mjtorn: Tested removing it, force-pushing the removal, because it works.

device/globals/interactive.gd
fleskesvor: Ok, I believe you; it just wasn't clear to me what purpose it would serve in this context. :slightly_smiling_face: Might be a good place to add a comment?

mjtorn: Fine, I'll be force-pushing a comment.

device/demo/rooms/test/can.esc
mjtorn: I hope this brain game isn't considered a blocker, because I did fidget around with the ordering until I came up with the helper.

fleskesvor: No, it's not a blocker, but it means I will have to dig into the esc parser again in the future to make this stuff easier. :worried:

mjtorn: Mmmmaybe. I never thought I'd hear me say this, but I might actually prefer not having an OR. That would lead to operator precedences and maybe grouping etc. Right now things are simple and doable. Maybe a discussion issue?

If you or anyone can think of an order for these that never breaks, and I'm not saying there isn't one, would be sweet if it could be after merging this :)

fleskesvor:
It looks to me like all concerns have been addressed now. Thanks for the contribution, @mjtorn! :slightly_smiling_face:

PR #182, opened 2018-06-04
Export camera drag margins in game.gd

mjtorn:
This controls how close to the edge of the screen the camera
has to be before starting to move.

The default value of 0.2 is from Godot's own default values.


fleskesvor:
This seems like a good idea. 🙂 Thanks, @mjtorn!

PR #183, opened 2018-06-05
Some UI Fixes

mjtorn:
Most of this should be pretty obvious, except maybe the vertical align one.

But Godot is really bad at handling textual dimensions and margins and such, so enforcing
a newline is the only way to do it.

### DIFF
--- a/device/globals/dialog_instance.gd
+++ b/device/globals/dialog_instance.gd
@@ -145,7 +145,7 @@ func init(p_params, p_context, p_intro, p_outro):
 
 	var parsed_ok = label.parse_bbcode(text)
 	assert(parsed_ok == OK)
-	label.bbcode_text = text
+	label.bbcode_text = "\n" + text
 	label.set_visible_characters(0)  # This length is always adjusted later
 
 	if self is Node2D:
--- a/device/globals/action_menu.gd
+++ b/device/globals/action_menu.gd
@@ -30,7 +30,7 @@ func check_clamp(click_pos, camera):
 	if dist_from_top < 0:
 		click_pos.y -= dist_from_top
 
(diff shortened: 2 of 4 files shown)


device/globals/dialog_instance.gd
```diff
@@ -145,7 +145,7 @@ func init(p_params, p_context, p_intro, p_outro):
 
 	var parsed_ok = label.parse_bbcode(text)
 	assert(parsed_ok == OK)
-	label.bbcode_text = text
+	label.bbcode_text = "\n" + text
```
fleskesvor: What is this for?

fleskesvor: I just noticed the commit message. Godot has margins though, but position, margin and offsets have been a huge mess in the demo dialogs. In #180, which is pending testing, I fixed those issues for dd_dialogs. I haven't touched this one yet though.

mjtorn: I haven't had the opportunity to look at #180 because everything in Lunar Adventure's codebase is such a mess, but if vertical alignment ("always center this, ok?") was so easy, I wonder why it isn't better documented. At least Godot doesn't have anything simple for it.

mjtorn: Do you think you will achieve proper vertical alignment in #180 ? I have talked about this with @mhilbrunner, that it would be very nice to have better support for dealing with text, and it does appear to be done best in Godot itself, with that newline hack as a temporary workaround.

If this PR is otherwise fine, we can just drop that one commit and I can keep it around my branch for personal use.

fleskesvor:
As discussed on IRC, this should be good to go now. Thanks, @mjtorn. :slightly_smiling_face:

PR #184, opened 2018-06-05
Fix teleporting an item to use Godot 3.0 API

mjtorn:
This is a critical yet trivial fix.


fleskesvor:
Thanks! :smile: Can't believe we missed that.

PR #185, opened 2018-06-05
Toggle trigger/exit activity with `visible`

mjtorn:
With this commit, esc scripts can use `set_active` on triggers and exits. Using only visible for this is a much cleaner implementation than adding an extra `active` boolean as I originally did. Thanks to @fleskesvor for the idea!


fleskesvor:
This looks great! Thanks, @mjtorn. :smile:

PR #187, opened 2018-06-14
Drop esc_type.BACKGROUND for esc_type.BACKGROUND_AREA

mjtorn:
As can be seen in the removed documentation, using `TextureRect`
for the background caused problems that are far worse than having
to un-center a `Sprite`. These include click masks and not being
able to use parallax.

Beside that, the old background.gd file extended `Control`, which
could lead to situations where something completely bizarre was
attempted for the background and break all the things.

Now there is only one very clear-cut and documented way of doing
things.


fleskesvor:
Yes, I agree. :slightly_smiling_face: 

There are also some words about which background to choose in the item doc, which should be rewritten to reflect this change.

mjtorn:
I updated the first commit with the note about item's documentation.

This change will obviously cause me pain, but I have to agree with it.

I decided that the commit history looks cleaner if the rename is a separate commit. From `master`s point of view it doesn't matter, but `git blame` will be nicer.

docs/background.md
```diff
@@ -2,19 +2,15 @@
...
 
 ### Parallax
 
 To create an illusion of perspective in a 2D game, a technique called parallax scrolling effect can be used. This effect is achieved by adding one or more additional background layers, which are moved at a different speed than the foreground when the camera pans.
 
-To implement a [ParallaxBackground](http://docs.godotengine.org/en/3.0/classes/class_parallaxbackground.html) in Escoria, a `Sprite` with the `globals/background_area.gd` script must be used for the foreground, as described in the previous section. Since the camera limits are calculated from the size of the `Sprite` texture, the image used must cover the visible space of the room, with transparent regions where the parallax layers will be visible.
+To implement a [ParallaxBackground](http://docs.godotengine.org/en/3.0/classes/class_parallaxbackground.html) in Escoria, a `Sprite` with the `globals/background.gd` script must be used for the foreground, as described in the previous section. Since the camera limits are calculated from the size of the `Sprite` texture, the image used must cover the visible space of the room, with transparent regions where the parallax layers will be visible.
```
StraToN: Either correct `globals/background.gd` to `globals/background_area.gd` here, or rename the remaining `globals/background_area.gd` to `globals/background.gd`.
I'd go for file rename personally.

docs/background.md
```diff
@@ -2,19 +2,15 @@
 
 The background node acts as a catch-all for input events. Any input event that isn't captured by another interactive element is processed by this node.
 
-To quickly add a background to your scene, add a `TextureRect` node as the first child of the `scene` node, name it "background" and add the `globals/background.gd` script.
-
-The drawback of this method, however, is that adding items to your scene with complex shapes becomes difficult, since you have to use `TextureButton`s with click masks, which have to be made with external drawing tools.
-
-For such use cases, the alternative is to use a `Sprite` for your background, name it "background", make sure the `Offset/Centered` property is not checked and add the `globals/background_area.gd` script.
+To add a background to your scene, add a `Sprite` node as the first child of the `scene` node, name it "background", make sure the `Offset/Centered` property is not checked and add the `globals/background.gd` script.
```
StraToN: Same as L13.

PR #188, opened 2018-06-19
Fix player angles/turning

mjtorn:


mjtorn:
I'm not 100% sure about the angles in the demo room, because the player character acts all wonky there.

The second test room looked better, but there are so many strange direction definitions that I'd appreciate someone checking this out.

Our Adventure Jam game uses very simple angles and this was required to make it work, so I don't know why it wouldn't work in general.

StraToN:
I approve this PR, however we really will have to look at demo scenes and check player angles.

fleskesvor:
It looks wrong with the player character in both demo rooms:

- moving and idling left, uses the up sprites
- moving and idling up, uses the right sprites
- moving and idling right, uses the down sprites
- moving and idling down, uses the left sprites

I don't think the demo angles are completely right either though, since eg. clicking slightly above the player character's right uses the down sprites. If the description in the `player_anims.gd` file is right, I think the correct angles should be 45 and 360 for walk_down, 135 for walk_right, 225 for walk_up and 315 for walk_left (ie. walk_right reversed).

PR #190, opened 2018-06-20
Do not version control .import/ folder, sync with Godot 3.0.3

akien-mga:
The .import/ folder contains imported binary files which are generated
from the *.import files and the source files. The contents of the .import/
folder should therefore not be version controlled, but left to Godot to
generate whenever you edit the project for the first time.

Godot 3.0.3 (and the master branch) now store the source file's MD5 sum
in the .import/ folder instead of each source file's corresponding
*.import file, so this commit synces this change (Godot 3.0.3 or later
should therefore be used to edit Escoria going forwards).

CC @fleskesvor who did 6b41a366, but I don't know what the rationale was.


StraToN:
Since we want to use the current stable version of Godot, this is good.

PR #192, opened 2018-06-21
Clean up how player's `animation` node is set

mjtorn:


device/globals/player.gd
```diff
@@ -345,12 +345,14 @@ func _ready():
...
-	animation = get_node("animation")
 	vm = $"/root/vm"
-	vm.register_object("player", self)
-	#_update_terrain();
+
 	if has_node("animation"):
-		get_node("animation").connect("animation_finished", self, "anim_finished")
+		animation = $"animation"
+		animation.connect("animation_finished", self, "anim_finished")
+		assert(animations)
```
fleskesvor: This is a good improvement. Might be better to use `vm.report_errors()` though, to provide the developer with a more helpful error message?

mjtorn:
>+		assert(animations)
>
>This is a good improvement. Might be better to use `vm.report_errors()` though, to provide the developer with a more helpful error message?

Good one! It started off as protection against me forgetting to set the
animations, but of course it's more useful if we use the correct error
reporting.

I'll do this as soon as I have real computer time.

mjtorn:
Is this better now?

fleskesvor:
Looks good; thank you, @mjtorn. 🙂

PR #193, opened 2018-06-21
Implement `set_angle` for interactives and player

mjtorn:
Essentially `turn_to` without the animations

### DIFF
--- a/device/globals/esc_compile.gd
+++ b/device/globals/esc_compile.gd
@@ -26,6 +26,7 @@ var commands = {
 	"walk": { "min_args": 2 },
 	"walk_block": { "min_args": 2 },
 	"turn_to": { "min_args": 2 },
+	"set_angle": { "min_args": 2 },
 	"change_scene": { "min_args": 1 },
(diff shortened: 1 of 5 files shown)


mjtorn:
On Thu, Jun 21, 2018 at 02:26:29PM -0700, Eivind wrote:
>Maybe you explained this in regards to `turn_to`, but what are the use cases for this? Both functions seem like they could be implemented with `set_state` and an animation, but if this has general use, I agree that it's a good idea to add a function for it.

Would `set_state` set the angle/last_dir correct? I doubt it would.

I found that when fading from black in `:ready`, and with an NPC's `turn_to`
executing, the animation was played in vain. That stuff has to be timed very
precisely to hide the animation.

So this is much more convenient when setting things up before showing them.

Decided against making it an option for `turn_to`, because esc is sometimes
a bit of a vague language to start with and I couldn't think of anything
better than this.

As a tangent: there should be some kind of `:setup` event, which would
set things up before `:ready` starts executing.

PR #194, opened 2018-06-21
Implement pose_scale for interactive and item

mjtorn:
Also drops the `set_state()` call, because it had
side effects. Be sure to set the state first!

### DIFF
--- a/device/globals/item.gd
+++ b/device/globals/item.gd
@@ -52,6 +52,10 @@ func anim_finished(anim_name):
 	if cur != state:
 		set_state(state, true)
 
+	if animations and "idles" in animations:
+		pose_scale = animations.idles[last_dir + 1]
+		_update_terrain()
+
 func set_active(p_active):
 	active = p_active
 	if p_active:
@@ -244,7 +248,9 @@ func set_speaking(p_speaking):
 		if get_node("animation").is_playing():
 			get_node("animation").stop()
 		set_state(state, true)
-	pass
+		if animations and "idles" in animations:
+			pose_scale = animations.idles[last_dir + 1]
+	_update_terrain()
 
 func set_state(p_state, p_force = false):
 	if state == p_state && !p_force:
@@ -282,10 +288,24 @@ func _update_terrain():
 	var color = terrain.get_terrain(pos)
 	var scale_range = terrain.get_scale_range(color.b)
 
+	# The item's - as the player's - `animations` define the direction
+	# as 1 or -1. This is stored as `pose_scale` and the easiest way
+	# to flip a node is multiply its x-axis scale.
+	scale_range.x *= pose_scale
+
 	if scale_on_map && (self is Node2D) && scale_range != get_scale():
-		var c = terrain.get_terrain(pos)
-		var s = terrain.get_scale_range(c.b)
-		set_scale(s)
+		# Check if `interact_pos` is a child of ours, and if so,
+		# take a backup of the global position, because it will be affected by scaling.
+		var interact_global_position
+		if has_node("interact_pos"):
+			interact_global_position = interact_pos.get_global_position()
+
+		self.scale = scale_range
+
+		# If `interact_pos` is a child, it was affected by scaling, so reset it
+		# to the expected location.
+		if interact_global_position:
+			interact_pos.global_position = interact_global_position
 
 	if light_on_map:
 		var c = terrain.get_light(pos)
--- a/device/globals/interactive.gd
+++ b/device/globals/interactive.gd
@@ -51,6 +51,8 @@ func walk_stop(pos):
 	set_state(state)
 
 	task = null
+	if "idles" in animations:
+		pose_scale = animations.idles[last_dir + 1]
 	_update_terrain()
 
 	if walk_context != null:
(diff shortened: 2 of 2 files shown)


device/globals/item.gd
```diff
@@ -282,10 +288,22 @@ func _update_terrain():
...
-		var c = terrain.get_terrain(pos)
-		var s = terrain.get_scale_range(c.b)
-		set_scale(s)
+		# Scan children to see if `interact_pos` is a child of ours, and if so,
+		# take a backup of the global position, because it will be affected by scaling.
+		var interact_global_position
+		for child in get_children():
+			if child == interact_pos:
+				interact_global_position = interact_pos.get_global_position()
+				break
```
fleskesvor: I think it's OK to assume that `interact_pos` is a direct child of the item scene, unless we're assuming that it might not be elsewhere. If a developer runs into problems with this, it will be obvious enough that they'll check on the documentation, and I wrote a doc on the typical anatomy of an item where it could be emphasised that it should be a direct child. I'm mentioning it because it would probably save some complexity.

device/globals/item.gd
```diff
@@ -282,10 +288,22 @@ func _update_terrain():
 	var color = terrain.get_terrain(pos)
 	var scale_range = terrain.get_scale_range(color.b)
 
+	scale_range.x *= pose_scale
```
fleskesvor: This could do with a comment, since it's not clear what its purpose is.

fleskesvor:
Does setting a different state this, eg. for changing a sprite have side effects? Does this work with multiple rotations, and flipping back to the opposite direction again? This is related to `turn_to` and `set_angle`, right?

mjtorn:
>-		var c = terrain.get_terrain(pos)
>-		var s = terrain.get_scale_range(c.b)
>-		set_scale(s)
>+		# Scan children to see if `interact_pos` is a child of ours, and if so,
>+		# take a backup of the global position, because it will be affected by scaling.
>+		var interact_global_position
>+		for child in get_children():
>+			if child == interact_pos:
>+				interact_global_position = interact_pos.get_global_position()
>+				break
>
>I think it's OK to assume that `interact_pos` is a direct child of the item scene, unless we're assuming that it might not be elsewhere.

You mean that we'd check this in `_ready`? I wanted to avoid introducing
global variables...

> If a developer runs into problems with this, it will be obvious enough that they'll check on the documentation, and I wrote a doc on the typical anatomy of an item where it could be emphasised that it should be a direct child. I'm mentioning it because it would probably save some complexity.

This isn't documented because it's fixed in code in this PR, and I'd prefer it
never being a problem over having to assume the user reads documentation
instead of giving up.

The only real improvement I could think of, if gdscript supports this, is
something like `if interact_pos in get_children()`, but depending on the
interpreter, there might be a significant difference here.

mjtorn:
>
>> @@ -282,10 +288,22 @@ func _update_terrain():
> 	var color = terrain.get_terrain(pos)
> 	var scale_range = terrain.get_scale_range(color.b)

>+	scale_range.x *= pose_scale
>
>This could do with a comment, since it's not clear what its purpose is.

Can we rename `pose_scale` in the future to something like
`target_direction` or `flip_direction` or something?

I will add a comment when I have time for it, but having better
variable names would be ok too :D I named it for consistency
here.

mjtorn:
On Thu, Jun 21, 2018 at 02:41:09PM -0700, Eivind wrote:
>Does setting a different state this, eg. for changing a sprite have side effects?

Dunno, I use Spine through `AnimationPlayer`, the stuff I'll document and push
as soon as A Lunar Adventure is ported over to this new scheme.

>Does this work with multiple rotations, and flipping back to the opposite
>direction again? This is related to `turn_to` and `set_angle`, right?

Yes, it works, by flipping the NPC and keeping her `interact_pos` intact.

mjtorn:
Now I force-pushed some commenting and the `if interact_pos in get_children()` version. Sadly I don't have the proper game code on me, but this shouldn't really be a problem, as the logic's the same.

fleskesvor:
>> I think it's OK to assume that `interact_pos` is a direct child of the item scene, unless we're assuming that it might not be elsewhere.

> You mean that we'd check this in `_ready`?

No, I didn't read your code properly, sorry. I saw the word "scan" in your comment, noticed the loop, and thought you were looking through the entire hierarchy of the interactive node, when you could have just used `has_node("interactive_pos")` to check for it.

fleskesvor:
> Can we rename `pose_scale` in the future to something like
`target_direction` or `flip_direction` or something?

Yes, that would be a good improvement. 👍

mjtorn:
> No, I didn't read your code properly, sorry. I saw the word "scan" in your comment, noticed the loop, and thought you were looking through the entire hierarchy of the interactive node, when you could have just used has_node("interactive_pos") to check for it.

I changed the code and the comment in the last (force-)push. Is it safe to say only `has_node("interactive_pos")`? It looks like it could be, because that's the only condition I can find in which the position is taken from a child.

Would you prefer I force-push such a change?

device/globals/item.gd
```diff
@@ -282,10 +288,24 @@ func _update_terrain():
...
+
 	if scale_on_map && (self is Node2D) && scale_range != get_scale():
-		var c = terrain.get_terrain(pos)
-		var s = terrain.get_scale_range(c.b)
-		set_scale(s)
+		# Check children to see if `interact_pos` is a child of ours, and if so,
+		# take a backup of the global position, because it will be affected by scaling.
+		var interact_global_position
+		if interact_pos in get_children():
+			interact_global_position = interact_pos.get_global_position()
```
fleskesvor: We're already saving a reference to `interact_pos` at the bottom of `_ready()`, if one is present. All current code should access the interact position via the getter function. Are there cases where external consumers of that function might get the wrong, flipped position?

fleskesvor: No, probably not.

mjtorn: `has_node("interact_pos")` does seem to work just as well, force-pushed it. Thanks for the tip!

mjtorn:
Because `has_node("interact_pos")` seems to work and should offer some benefit, I propose this PR gets merged now.

PR #195, opened 2018-06-25
Do not report errors if animations is actually set in player

mjtorn:
Sorry about that, chief.


fleskesvor:
Thanks for the fix. :slightly_smiling_face:

PR #196, opened 2018-06-25
Allow sched_event to have multi-word events

mjtorn:
Because sched_event is so useful for testing, when you temporarily
set a game state in the room's event script, it should be possible
to test an object's `use inventory_item` events as well.

This commit brings that.


StraToN:
Good bring :) Thanks @mjtorn  !

PR #197, opened 2018-06-25
[RFC] Make player's set_state actually set a state, not just a costume

mjtorn:
This aligns the behavior with NPC behavior without breaking
how simple items can still be altered with `set_state`. The difference
lies in how idle animations are handled and when the costume
change comes into effect.

Therefore a character's costume change must be a combination of
`cut_scene` to allow the costume change animation to finish and then
an explicit `set_state char idle` or whatever you want your character to
do next.

### DIFF
--- a/device/globals/item.gd
+++ b/device/globals/item.gd
@@ -45,12 +45,9 @@ func anim_finished(anim_name):
 		set_scale(get_scale() * anim_scale_override)
 		anim_scale_override = null
 
-	var cur = animation.get_current_animation()
-
-	# `cur` is always "" for a finished animation,
-	# this forces the state to be reset and the animation played again
-	if cur != state:
-		set_state(state, true)
+	# Although states are permanent until changed, the underlying animations are not,
+	# so we must re-set the state for the appearance of permanence
+	set_state(state, true)
 
 func set_active(p_active):
 	active = p_active
@@ -252,9 +249,8 @@ func set_state(p_state, p_force = false):
 
 	# printt("set state ", "global_id: ", global_id, "state: ", state, "p_state: ", p_state, "p_force: ", p_force)
 
-	if has_node("animation"):
-		get_node("animation").stop()
 	state = p_state
+
 	if animation != null:
 		if animation.is_playing() && animation.get_current_animation() == p_state:
 			return
--- a/device/globals/player.gd
+++ b/device/globals/player.gd
@@ -28,6 +28,8 @@ var anim_scale_override = null
 var sprites = []
 export var placeholders = {}
 
+var state
+
 func set_active(p_active):
 	if p_active:
 		show()
@@ -61,7 +63,10 @@ func anim_finished(anim_name):
 		set_scale(get_scale() * anim_scale_override)
 		anim_scale_override = null
 
-	animation.play(animations.idles[last_dir])
+	# Although states are permanent until changed, the underlying animations are not,
+	# so we must re-set the state for the appearance of permanence
+	set_state(state, true)
+
 	pose_scale = animations.idles[last_dir + 1]
 	_update_terrain()
 
@@ -100,7 +105,7 @@ func anim_get_ph_paths(p_anim):
 	return ret
 
 func play_anim(p_anim, p_notify = null, p_reverse = false, p_flip = null):
-	if p_notify != null && (!has_node("animation") || !get_node("animation").has_animation(p_anim)):
+	if p_notify != null and (!animation || !animation.has_animation(p_anim)):
 		vm.finished(p_notify)
 		return
 
@@ -114,7 +119,6 @@ func play_anim(p_anim, p_notify = null, p_reverse = false, p_flip = null):
 			node.replace_by_instance(res)
 			_find_sprites(get_node(npath))
 
-
 	pose_scale = 1
 	_update_terrain()
 	if p_flip != null:
@@ -125,10 +129,9 @@ func play_anim(p_anim, p_notify = null, p_reverse = false, p_flip = null):
 		anim_scale_override = null
 
 	if p_reverse:
-		get_node("animation").play(p_anim, -1, -1, true)
+		animation.play(p_anim, -1, -1, true)
 	else:
-		get_node("animation").play(p_anim)
-		#get_node("animation").seek(0, true)
+		animation.play(p_anim)
 
 	anim_notify = p_notify
 	var dir = _find(p_anim, animations.directions, p_flip.x)
@@ -307,12 +310,17 @@ func teleport(obj):
 	set_position(pos)
 	_update_terrain()
 
-func set_state(costume):
-	var node = get_node(costume)
-	# You can `set_state player default` with no animation by that name, and get "animation"
-	animation = node if node else $"animation"
-	assert(animation is AnimationPlayer)
-	animation.play(animations.idles[last_dir])
+func set_state(p_state, p_force = false):
+	if state == p_state and !p_force:
+		return
+
+	state = p_state
+
+	if animation != null:
+		if animation.is_playing() and animation.get_current_animation() == p_state:
+			return
+		if animation.has_animation(p_state):
+			animation.play(p_state)
 
 func teleport_pos(x, y):
 	set_position(Vector2(x, y))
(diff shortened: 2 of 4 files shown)


mjtorn:
There is no support for changing costumes as such in Godot, like how Spine nodes contain "skins" that can easily be changed in an animation. `cut_scene` is a bit of a funny name for it, as it brings something else to mind than "play this animation in a blocking manner", but c'est la vie.

Would it be possible to replace the sprite sets in an animation called `skin_hat` and `skin_default` and then use the normal `idle_down` and `walk_right` style animations - all in one AnimationPlayer? This would be analogous to changing skins in a Spine node. I don't have the time to test it now :(

If this `set_state` change is unacceptable, I would like to hear what kind of semantic would be acceptable for making the player do things, in an animated loop, like a state.

Over to you, @StraToN, @fleskesvor and others, cheers!

fleskesvor:
With this change, is it the actual state that gets saved to the save game, and not the costume? I don't mind using a different mechanism to change costumes, but then we'd need to save the current custom in a global and the developer must explicitly check for that and run the same cut scene again when loading a save, right?

I don't have any strong feelings about how to solve this either way, but I think we can re-implement this stuff in a better, more declarative way once the AnimationTree/AnimationGraph/Animator node is implemented, so I was hoping to hold off a bit on it.

mjtorn:
You're probably right in your concern here, but I'll double check at some point.

Because you're probably right, what's the best fix? I might be able to duplicate `AnimationPlayer`s in the Python script, but that might not be enough. If `set_state player float` (like Shay-in-space) is a valid state, which gets saved in a savegame, how do we know the costume?

To put it in demo-room terms, you'd need to save both `set_state player hat` and `set_state player float` either way.

Should we just add a variable to player.gd and item.gd called "costume"? The most intuitive way would be to have `set_costume player blue` even if the implementation would be an animation that sets the variable and changes the skin/costume/whatever. This is to avoid figuring out when to save two states or saving a mandatory animation run and a state.

I don't yet know what to propose as a fix, not knowing what vanilla Godot is capable of. Could we have one `AnimationPlayer` for costumes and another for all other animations?

fleskesvor:
> If `set_state player float` (like Shay-in-space) is a valid state, which gets saved in a savegame, how do we know the costume?

We don't know the costume; it's the `AnimationPlayer` that we know.

> To put it in demo-room terms, you'd need to save both `set_state player hat` and `set_state player float` either way.

No, but you need `AnimationPlayer`s for all possible scenarios with the current solution, like `hat_walk` when the character is walking with a hat, `hat_float` when the character is floating with a hat, `default` when the character is walking without a hat, etc.

I agree that it's not great, and it's tedious to configure a bunch of `AnimationPlayer`s, but it does work with saving the state.

mjtorn:
I suppose it's up to me to see if the animation sprites can be changed, because that would be compatible with vanilla-godot and enable us to track `costume` as its own thing, and thus provide a clean interface without multiple `AnimationPlayer`s.

Either that or have a separate `AnimationPlayer` for the animations that change a Spine characters skin (= costume) and another for basic animations. Or duplicate the shit out of everything, but that's - as you say - tedious.

mjtorn:
One thing occurred to me about saving costumes. It doesn't save the outcome of the `cut_scene` call, but because you need to track costume changes in every room (which you can enter in a special costume), you have something like

```
> [got_helmet]
   cut_scene player skin_blue
   set_state player idle
```

in all those rooms anyway. Obviously `got_hekmet` will be stored with the other globals. So that's not a problem.

I haven't looked at changing which set of sprites to use from animation, but I would really hate to have a solution which duplicates animation players, duplicates animations and has different semantics from NPCs. There has to be a better way.

fleskesvor:
> One thing occurred to me about saving costumes. It doesn't save the outcome of the cut_scene call, but because you need to track costume changes in every room (which you can enter in a special costume), you have something like

The way it works now, you would have to use

    > [got_helmet]
       set_state player skin_blue

Since you use state to change which `AnimationPlayer` is being used, not which animation is playing. That is only handled by code. I feel that setting specific animations should mostly be a consequence of commands (like `say` and `walk`) and not directly manipulated from esc scripts. I would welcome an easier way to change costumes without using multiple `AnimationPlayer`s though.

mjtorn:
But that's not a viable solution.

I tried explaining this before, and I'll try again, but there'd need to be two states for your approach.

First you do `set_state player skin_blue` which looks up an `AnimationPlayer` named "skin_blue" and plays its idle animation.

Then you want the player to float in space, `set_state player float`, which looks for an `AnimationPlayer` named "float", which it doesn't find, so it reverts the costume and plays the idle animation.

Is that what you really want to happen?

-----

That's why there has to be a global that tracks the conditions (= state of the game) in which to change costumes and have only one (1) `AnimationPlayer` to set the state like "float" or "jump_up_and_down_screaming".

Alternatively we'd find a solution that uses animations to change the set of frames used by animations like "idle" or "float" to look like a costume-change.

I haven't had time to find such a solution, because it Just Works(tm) as expected with Spine.

fleskesvor:
> Is that what you really want to happen?

No, it's not what I want to happen, but I was saying that's the way it works now. The way the player character moves (walk/float/run or whatever) isn't a separate concept from the way the PC looks (eg. default/dress/fancy_hat, etc.). Both movement and looks have to be a part of the same "costume" or `AnimationPlayer`, which leads to a lot of duplication. I agree that this isn't ideal, so I don't think we disagree about what we want to achieve, but your example code isn't how player states work at the moment, unless that got changed in a recent commit that I didn't test. There's no setting states to idle/speaking/walk. Those states are handled by scripts, and float is just a fancy walk cycle.

mjtorn:
That's the problem, it doesn't work that way in Escoria at the moment, but I implemented it using Spine. I seriously haven't had time to see if the sprites can be changed to match the behavior, which is why this is RFC ;) And I do appreciate your comments!

I can't also help but think about how to cope with NPCs and costumes. It's a valid use case, if you have a game where NPCs can wear disguises, for example.

We should think hard about NPCs and how to make them less distinct from the PC, because items are so overloaded they're hard to reason about.

-----

This seems like something that's hard to fix unless we agree on an implementation and I would have to find the time to do a reference implementation for the demo scene that's more palatable. I just don't know when.

The alternative is to figure out a scheme that uses multiple `AnimationPlayer`s.

It would have to have something like your

```
> [got_helmet]
   set_state player skin_blue
```

but in case the player is set into a completely different state, it would be like

```
> [got_helmet, is_floating]
  set_state player skin_blue_float
```

Though in that case there'd be an `AnimationPlayer` called "skin_blue_float".

How do we reason about that and the `animations` attribute? Obviously the idle animation in that case would have to be a  floating animation, but I couldn't find a way to define that.

Can we move `animations` into the `AnimationPlayer`s?

mjtorn:
I added diagonal walking to our game and noticed some problems, like the fact that if I change the costume with `cut_scene` as in this PR, and then `set_state player idle`, the player always returns to `idle` regardless of the direction he's facing once `anim_finished` kicks in.

This happens also with the rope-pulling sequence. If I `set_state` the character to pull the rope, the same state gets repeated after `anim_finished`, which is the expected behavior. Then calling `set_state player idle` will cause him to idle in the wrong direction.

Not sure if this is the latest version in this PR because of all the hate it's got, but if I try to change the costume using `set_state`, which is a 0.001s or so animation, Godot/Escoria will waste a lot on repeating that animation. And the character won't idle.

I may have to draw a diagram of exactly how complex this thing would get with all the states and `AnimationPlayer`s just to see if something else is more viable, but it would be nice to achieve a resolution here that doesn't blow the characters completely out of proportions with the duplication.

This approach of not setting a state (as an animation that loops until interrupted) for the player is also problematic for cut scenes, as something like the rope-pulling animation suddenly requires that a copy be made with very precise timing to match the NPCs that simply are in a state. That also makes life hard.

mjtorn:
Close this once #219 is merged instead.

mjtorn:
I'm confident enough that #219 or a variant of it will go in and cause less of a political shitstorm, so I'll just close this and delete the branch. Our games are running code from #219 anyway now.

PR #198, opened 2018-06-25
Fix dialog

fleskesvor:
This reverts the change that placed the dialog player into a canvas layer in main.tscn, since unless position was fixed, dialog would be placed outside of the viewport in most cases, because of the difference between the scene size and the actual size of the viewport. There would have to be code in place to calculate position from the ratio between the two, but the previous solution is likely more predictable.


mjtorn:
Please merge this! It made the problems go away.

The rationale of having a single place for the dialog player is valid and legit, but someone has to have the spare time to write that ratio code.

Thanks a lot for the fix :)

PR #200, opened 2018-06-26
Clean up item states, especially useful for NPCs

mjtorn:
In interactive.gd, don't stop a playing animation, but let item.gd
do that. Having a stopped animation can lead to weird side effects,
which I can't explain, but where `state` is set though
`animation.get_current_animation()` returns `""`.

Now there's no need to call `animation.stop()` before setting a state,
so that's cleaner. I also cleaned up some of the `get_node("animation")`
mess because that's already bound to a variable anyway.

### DIFF
--- a/device/globals/interactive.gd
+++ b/device/globals/interactive.gd
@@ -46,8 +46,7 @@ func walk_stop(pos):
 	set_position(pos)
 	walk_path = []
 
-	if animation != null && animation.is_playing():
-		animation.stop()
+	# Walking is not a state, but we must re-set our previous state to reset the animation
 	set_state(state)
 
 	task = null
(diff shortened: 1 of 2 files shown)


mjtorn:
This has some slight overlap with #197 because I touched item there, but I believe this is a clean PR to merge. It seems to have fixed my problems where the NPCs get stuck in strange states, which I still can't explain, but the appearance is that an NPC who should be idle - and according to `printt()` is idle! - has a walking animation.

mjtorn:
This should be fine with anything, because it only interacts with `AnimationPlayer` and doesn't care much more. I think I tested it with the demo scene, but if you want, I can test it later to be sure. Unless you happen to have it handy. I think the spinning autosave is the test case to go with, as not many items are animated currently in the demo.

For what it's worth, room1 in our jam game does use an `item` with `AnimationPlayer` to change a sprite, no Spine involved, and it works. This is by calling `set_state` which plays the animation, hitting the code in `anim_finished` when done.

StraToN:
Ok, if you have it tested and working in both cases then I'm fine with it. Once again, we'll need pretty demo scenes soon to have a better time testing.

PR #202, opened 2018-06-26
Revert "Move `dialog_default` out of demo as a `Container`"

mjtorn:
This reverts commit 28dea4d2ad5662dc2f27565c972965a68a79d1a2.

Because canvas layers didn't happen yet, a Container with the
dialog will be below regular items. My gut, and experimenting
with the ordering in the scene tree, tells me that this is
how it's supposed to work. Either you use Z or a separate
canvas layer.

### DIFF
--- a/device/demo/ui/dialog_default.tscn
+++ b/device/demo/ui/dialog_default.tscn
@@ -4,7 +4,7 @@
 [ext_resource path="res://demo/ui/dialog_theme.tres" type="Theme" id=2]
 
 
-[node name="dialog_instance" type="Container"]
+[node name="dialog_instance" type="Node2D"]
 
 anchor_left = 0.0
 anchor_top = 0.0
@@ -24,7 +24,7 @@ typewriter_text = true
 characters_per_second = 45.0
 fixed_pos = false
 
-[node name="anchor" type="Container" parent="." index="0"]
+[node name="anchor" type="Node2D" parent="." index="0"]
 
 anchor_left = 0.0
 anchor_top = 0.0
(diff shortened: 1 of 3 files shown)


mjtorn:
@fleskesvor do you remember what happened here? I was so sure a PR like this existed at some point... Was it closed in favor of #180 or something?

fleskesvor:
Yeah, I think that's how it's supposed to work. Usually you use a separate canvas layer with Container nodes for a UI, in which case z index isn't a problem. Using Node2Ds also works, but then you have to make sure the z index is higher than other elements in the scene tree.

I think you're thinking of #174, which had a similar change, and which was closed in favor of #180.

mjtorn:
Having dialog in a canvas layer probably makes sense also for the use case where `telon` fades out but you want to have someone `say` something, so that the dialog text is visible. We have that in our game and I was thinking of making a `not_telon` to achieve this, but no idea how to approach the problem.

I do still propose merging this, because it takes care of a problem, until canvas layers replace this. At least leave the PR hanging until a viable solution is implemented. This is because I don't want crap in my fork that doesn't have a PR, as that makes maintainability harder.

StraToN:
Canvas layers definitely have my preference overall.
I am fine with merging this PR if it solves the problem, but using Canvas Layers should definitely be a priority.

fleskesvor:
Ok, I'm creating an issue to revert this and implement a solution with canvas layers so that we don't forget about it.

Thanks for the PR, @mjtorn. :slightly_smiling_face:

PR #204, opened 2018-06-26
Python script for creating Spine nodes, and documentation

mjtorn:
The script might not win beauty contests any more than the documentation might win literary contests, but at least we should be able to close #134 and #159!

And this one after you merge it ;)


fleskesvor:
Great job on getting this to work! 🙂 I'm wondering if it wouldn't be better to have this be a separate project from Escoria though, since it's not a core part of the framework, and link to it under a "resources" section in the readme. What do you think, @StraToN?

StraToN:
Let's ask @akien-mga for his opinion, but I think I have an idea of his answer: this script, even very interesting, is not specifically related to Escoria itself and thus should not be merged in Escoria, nor in Godot editor because we cannot deal with all animation softwares of the industry ourselves ; this is the community's work.

I'd personally advise to keep this script separated, but make it visible (in the README or somewhere else). Another good idea would be to propose Spine to include this script with their official tools so they can evolve it.

akien-mga:
Since it seems useful for non-Escoria Godot projects, I think it might be better hosted separately indeed and linked to in the docs (the added docs file for Spine usage in Escoria could stay, with relevant edits to link to where that script lives).

Otherwise if it were Escoria-specific, but not a dependency of Escoria (i.e. you don't need it to use the framework), I wouldn't mind having it in the main tree but it should likely go in a specific folder like `contrib/spine` or something like that.

mjtorn:
Thanks for the encouraging and nice comments :)

Sadly it's not Escoria-specific. In order to put all the Escoria goodness in the script, I'd have to define some kind of DSL/jsonschema where you type in all the things, run the script, and have it blow up because who the hell gets that stuff right on the first try. Nor would I have time to write unit tests for something so complex. On top of that, it would probably require a Godot scene parser in Python to compare results with.

Instead the unfolded sections are such that it's not too painful to do the Escoria stuff after creating a scene with this, which also makes it more general.

I understand and actually agree with your assessment and hope this PR can hang in here until I have time to organize it somewhere else.

StraToN:
Let's keep this one pending then, until we find the best place to put this script.

mjtorn:
I created a repo https://github.com/mjtorn/spine-to-godot for the tool, so I'll drop everything but the documentation from this PR and force-push.

Showed it at https://github.com/EsotericSoftware/spine-runtimes/issues/728 as well.

mjtorn:
@StraToN, clear to merge?

StraToN:
Yep! Great job on this @mjtorn  :)

PR #205, opened 2018-07-04
Settings for HiDPI/Retina, fix aspect ratio

mjtorn:
Although the documentation for
[Multiple Settings](http://docs.godotengine.org/en/3.0/tutorials/viewports/multiple_resolutions.html)
makes it sound like the `Expand` aspect ratio doesn't do
what we want, godotengine/godot#19912 states that this is the correct way.

~Further documentation may be warranted for the scene tree to
avoid having a clear-color bar at the top of a Retina game.~

~This commit does not yet address that documentation.~

It appears that it was a bug in Escoria that caused the problems with other aspect ratios.

### DIFF
--- a/device/globals/main.gd
+++ b/device/globals/main.gd
@@ -88,34 +88,6 @@ func wait(time, level):
 	wait_timer.set_one_shot(true)
 	wait_timer.start()
 
-func check_screen():
-#	var vs = OS.get_video_mode_size()
-	var vs = OS.get_screen_size()
-	if vs == vm_size:
-		return
-	vm_size = vs
-
-	var rate = float(vs.x)/float(vs.y)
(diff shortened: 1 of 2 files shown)


mjtorn:
I tested removing `_check_screen()` on OSX (Retina) and Linux (4k) and it works, but not exports or Windows.

fleskesvor:
Interestingly, this also make the hud canvas layer scale with the window size, which might it easier to keep positioning of dialogs when they are moved and if the window is scaled. I found no regressions in testing either, so this looks like a good improvement. Thanks, @mjtorn. 👍

fleskesvor:
No worries; easy mistake. :slightly_smiling_face:

PR #208, opened 2018-07-13
Use the "keep" stretch mode for HiDPI/Retina

mjtorn:
If I'd have had time to properly test these during the past two weeks or so, this would have been a part of #205 :/

### DIFF
--- a/device/project.godot
+++ b/device/project.godot
@@ -23,11 +23,14 @@ esc_type="*res://globals/escoria_types.gd"
 
 window/size/width=1920
 window/size/height=1080
+window/size/resizable=false
+window/size/fullscreen=false
 window/size/test_width=1920
 window/size/test_height=1080
 window/dpi/allow_hidpi=true
 window/stretch/mode="2d"
-window/stretch/aspect="expand"
+window/stretch/aspect="keep"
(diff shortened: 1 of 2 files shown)


fleskesvor:
Why does your PR also set "resizable" and "fullscreen" to false? I agree that you usually shouldn't be able to resize a game window, but for convenience and until we have a settings template to change screen sizes, it might make testing easier to be able to resize windows?

fleskesvor:
Ok, that makes sense. Do you think you could add a line or two to the export docs about which settings to set to what before exporting?

mjtorn:
@fleskesvor are you ok with that documentation? I added some other tacit knowledge I've accrued over the months as well.

mjtorn:
I hope this isn't seen as an abuse of my new superpowers, but this has been hanging for a while and is proven to work so I think it should go in. We can address documentation issues later if need be, but let's keep the balls rolling!

fleskesvor:
Sorry for the late reply. Yes, that documentation is perfect. :+1: 

> I hope this isn't seen as an abuse of my new superpowers

No, there's no controversy here. Welcome onboard. :slightly_smiling_face:

PR #209, opened 2018-07-17
Option for items to set only z-index from scalemap

mjtorn:
This de-couples z-indexing from scaling. The use case is when
gfx artists draws a room with different "element items" as layers
in the same scale, and using the sizemap to get the z-index would
shrink the item.

The alternative to this is to debug what the scale factor is,
divide 1 by it, multiply by 100, scale that image separately,
and have it look ridiculously yet uselessly big in the editor.

Now it's possible to retain the artist's scales but also have
the player move behind the element item!

### DIFF
--- a/device/globals/item.gd
+++ b/device/globals/item.gd
@@ -13,7 +13,7 @@ export(Color) var dialog_color = null
 export var talk_animation = "talk"
 export var active = true setget set_active,get_active
 export var placeholders = {}
-export var use_custom_z = false
+export var dynamic_z_index = true
 
 var anim_notify = null
 var anim_scale_override = null
@@ -268,8 +268,9 @@ func teleport_pos(x, y):
 	_update_terrain()
 
 func _update_terrain():
-	if self is Node2D && !use_custom_z:
+	if self is Node2D and dynamic_z_index:
 		set_z_index(get_position().y)
+
 	if !scale_on_map && !light_on_map:
 		return
 
@@ -278,6 +279,7 @@ func _update_terrain():
 	var terrain = $"../terrain" if has_node("../terrain") else null
 	if terrain == null:
 		return
+
 	var color = terrain.get_terrain(pos)
 	var scale_range = terrain.get_scale_range(color.b)
 
@@ -286,7 +288,7 @@ func _update_terrain():
 	# to flip a node is multiply its x-axis scale.
 	scale_range.x *= pose_scale
 
-	if scale_on_map && (self is Node2D) && scale_range != get_scale():
+	if self is Node2D and scale_on_map and scale_range != get_scale():
 		# Check if `interact_pos` is a child of ours, and if so,
 		# take a backup of the global position, because it will be affected by scaling.
 		var interact_global_position
@@ -300,7 +302,7 @@ func _update_terrain():
 		if interact_global_position:
 			interact_pos.global_position = interact_global_position
 
-	if light_on_map:
+	if self is CanvasItem and light_on_map:
 		var c = terrain.get_light(pos)
 		modulate(c)
 
--- a/docs/items.md
+++ b/docs/items.md
@@ -20,7 +20,17 @@ It may be that you want these default actions to require double-clicks. This is 
 
 An item-local default action overrides a global default action, as you'd expect.
 
-### Background items
+## Z-index
+
+By default every room in the game is considered to have faux "depth". This is done by setting the z-index based on the y-coordinate.
+
+If you want things to be flat or manually controlled, you can use Godot's z-indexes by unchecking the `Dynamic Z Index` checkbox.
+At least for certain inventory configurations, this is something you must do. If your items don't show up on the HUD, uncheck!
+
+The system isn't perfect, so sometimes you may have to use an `AnimationPlayer` to tweak the z-index given by Godot. Likewise
+if your item consists of many nodes, they will have different positions, so some of them may need manual tweaking.
+
+## Background items
 
 There are also times when you want to enable the player to interact with items which are not separate scenes, but drawn directly on the background. For such use cases, you will typically use a `TextureRect` to frame the item, with an optional click mask, and no "area" child node.
 
@@ -33,7 +43,7 @@ You can also use an `Area2D` for your background item with a `CollisionPolygon2D
     * `CollisionPolygon2D` which defines the shape of the item
     * ...
 
-### Parallax items
+## Parallax items
 
 Items can also be added to a parallax layer, but this comes with a few caveats. First of all, you need to use an item with the shape defined by a `Control` node, since the mouse cursor cannot interact with an `Area2D` through the foreground layer. Secondly, since these items will be visible to the cursor through the foreground, any items meant to be obstructed by foreground elements will need to be masked by a `Control` node. These masks don't need to be items, but can be `TextureRects` with click masks for more fine grained control of their shape. Do take care not to cover any items in the foreground with these masks though, as that will make them impossible to interact with.
 


StraToN:
This could also be of use for the player (if it's not done yet, can't verify now). 

In some cases, depending on the room (usually exterior, but not only), the player may have a different scale on the left and on the right side, even if its Y position is the same.
This currently avoids the effects we observe at the Vortex place in LucasArts's Sam & Max Hit the Road game.

StraToN:
This allows items to react similarly as the player as you move them in the editor as well, right?

fleskesvor:
I agree with decoupling the z-index from scaling imposed by the scalemap. What are the use cases for changing the z-index anyway? Something that was used in Dog Mendonça? I mean, if there are no valid use cases, it might be less confusing to just remove that code.

mjtorn:
@StraToN sadly, it does not. The player is a `tool` so it gets dynamically updated. There's been talk for almost forever that we need a similar npc.gd script but it's never been important enough. I tried sketching it out and there was a ton of duplication, which would need to be abstracted out into a static script with methods for the objects. That's a separate discussion.

@fleskesvor The case is that NPCs can move behind other items, for example. Or items that are moved in-game without being characters.

So z-indexing is a valuable feature for players and non-players alike, but having forced scaling in the same is not ok.

mjtorn:
It is derived, yes. The y coordinate and z-index have basically the same value, as the y-axis value increases downwards, and thus items that appear at the foreground will most often behave as expected because their positions are at a lower-position higher-y-value coordinate.

I don't think it's always 100% correct but this is where `use_custom_z` comes in. It hard-codes a z value to be used. So it's good for element-type items, but can never work for NPCs and such, obviously, if they move around.

--------

This is good to merge now?

mjtorn:
It does not duplicate it, because `use_custom_z` uses a fixed value. This takes the z value from the scalemap without actually scaling the item.

fleskesvor:
Could you point out where that happens, because I cannot find it? To me `z_from_scale` just looks like `use_custom_z` by another name, or rather `not_use_custom_z`, to be precise.

If they are completely different things, having both `z_from_scale` and `use_custom_z` set to `true` should probably result in a warning, since a developer might expect that setting one of them is enough to control how an item is scaled, while it might not be.

mjtorn:
z-01-default-behavior

That's how you'd expect it to be with current master, that you scale it from the scalemap _and_ get the z-index for free. Note that when I was in a bad place in many ways pushing the game out for AdvJam2018, I flattened the UFO onto the background and the character walked only around it.

Then when I made it an item, in this case the scales match, but taking the z-index from the scalemap makes it smaller. This led me to implement the new boolean, instead of calculating an upscale factor for the UFO item image just so Escoria can scale it back down again.

Obviously broken, right?

z-02-nothing-set

Here we can see what happens when nothing is set. The player happily walks on the UFO.

z-03-z-from-scale

This is how it's supposed to be \\:D/

--------

Determining the `use_custom_z` value is painful extra work, so that's why this is a significant improvement.

I do agree they are in conflict, I'll add a call to `report_errors()` about it, thanks!

fleskesvor:
Yes, I can see what you mean, but my opinion is that you'd get the same result by setting both `scale_on_map` and `use_custom_z` to `false`, without introducing another variable.

If I'm reading the code correctly, in the three cases above you have, in order:

```
use_custom_z = false
z_from_scale = false
...
if false and not false:
    set_z_index(get_position().y)
```

```
use_custom_z = false
z_from_scale = false
...
if false and not false:
    set_z_index(get_position().y)
```

```
use_custom_z = false
z_from_scale = false
...
if true and not false:
    set_z_index(get_position().y)
```

Only the last expression evaluates to true, which makes it set the UFO to a higher z index than the player, since it's y coordinate is greater.

mjtorn:
The middle image has `scale_on_map`, `use_custom_z` (and `z_from_scale`) as `false`.

So obviously it does not get the same result, because the player is standing on the UFO :D

This is why I said _"Determining the `use_custom_z` value is painful extra work, so that's why this [`z_from_scale`] is a significant improvement."_

Do you have a better solution than adding `printt()` calls while running the game to see where the player is positioned at all times and where the UFO might be?

fleskesvor:
Yes, but in the first image `scale_on_map` is set to `true`, and in the second it's set to `false`. I still fail to see how `z_from_scale` doesn't duplicate `use_custom_z`, only with the opposite meaning. I agree that `z_from_scale` might be a more descriptive name though.

mjtorn:
Because `use_custom_z` hard-codes a value (from `Z Index/Z Index`) and `z_from_scale` takes the dynamic z-index value from the scale map _without scaling the image_ - as can be seen in the first image. 

It's broken because the UFO is tiny as a consequence of `scale_on_map`.

mjtorn:
@fleskesvor did you maybe mean a bug I found?-) It would be super-nice if A Lunar Adventure was the demo game so maybe these "omg I need everything done two weeks ago" situations would get more eyes...

There was a problem where `scale_on_map` wasn't always respected, so I force-pushed that now as a fix.

fleskesvor:
I mean that using the `Z Index/Z Index` and the derived `z index` from the scalemap are mutually exclusive, so I still don't understand why we need to different booleans for what, to me, appears to do the same thing.

And yes, I agree that a more fleshed out game like A Lunar Adventure would be great as a demo game, but the source would need to be available for us to test with.

mjtorn:
The code does issue an error if `use_custom_z` and `z_from_scale` are both set, so that's done. We can't really add that check for `scale_on_map` because we must be able to use only the scaling feature of that map and use `use_custom_z`.

@StraToN do you see where I'm coming from and going to with this? We did talk about z-index stuff on IRC but it was before our holidays so I'm not sure how deep we went into exactly this topic ;)

------

#222 for discussion, but not to supplant a merge.

-----

As for releasing the source of A Lunar Adventure, as much as part of me might want to do it, I wouldn't hold my breath. The negotiations about these sorts of things last forever, and then it'd have to be a very shut-down license to avoid fork hell. Also the voice overs would never be released.

StraToN:
I had difficulties to understand the problem, but I think I got it by re-reading the discussion above.

IIUC, you need and option for items (and only items I suppose?) to NOT scale themselves with the scalemap value because this results in wrong sizing. The "good" way to do this then, would be to resize the image in your sprite editor to the right dimensions so it finally scales correctly with the scalemap.
But this is boring.
And you can't just discard the built-in scalemap feature, because if you do then you'll loose Z management (and players may actually walk over the item). So you actually need to "correct" the Z value obtained for the scalemap.

If this summary is correct, please confirm @mjtorn :)
If yes, then you indeed have 2 solutions:

1. Set a Z value yourself, hard-coded. That's what you did I think with your "custom" Z.
2. Set an optional "Z correction" that will multiply the scalemap value by the "Z correction" value. Disadvantage here is that if you modify the scalemap, then you'll have to find the according Z correction again.

mjtorn:
@StraToN I think you got it! Only items. And it's not "boring" because items aren't `tool` scripts, so they completely mess up the editor view by covering everything of importance underneath them. It's une espice de merde.


For solution 1, I keep repeating _"Determining the use_custom_z value is painful extra work, so that's why this [z_from_scale] is a significant improvement."_ ;)

For solution 2, I don't know why adding such code is any better than recycling the z-value from the scalemap o_O :confused:  Determining that value is just as hard as determining a value to hard-code, because it's essentially just another hard-coded value by itself. And what you said.

StraToN:
> For solution 1, I keep repeating "Determining the use_custom_z value is painful extra work, so that's why this [z_from_scale] is a significant improvement." ;)
> 
> For solution 2, I don't know why adding such code is any better than recycling the z-value from the scalemap o_O confused Determining that value is just as hard as determining a value to hard-code, because it's essentially just another hard-coded value by itself. And what you said.

Both work, but 1 is indeed easier. Maybe it is just a naming problem. What about "Z override value" instead of "Custom Z" @fleskesvor? Would it make it more understandable?

Anyway, since this value is exported, Godot allows you to move the cursor value in the inspector, but if `item` is not a `tool` then the editor will not resize it. This is not so good... What about making `item.gd` a tool?

mjtorn:
@StraToN and I discussed this on irc, that there should be a variable `fixed_z_index` where -1 means "automatic".

I think at least I lost the perspective a bit, because if we set a `fixed_z_index` it's a redundancy of Godot's own z index. I'll try to propose a `dynamic_z_index` boolean which is essentially `z_from_scale` but better decoupled from the scaling logic, to avoid this redundancy.

mjtorn:
Added a last-minute update to the documentation.

I believe this is as we discussed so I will be merging it fairly soon, unless @StraToN or @fleskesvor says anything. A Lunar Adventure was not broken by this and suddenly things make more sense in the editor.

mjtorn:
Confirmed ok on IRC, thanks!

PR #211, opened 2018-07-19
Ask ClassDB about connecting signals

mjtorn:
Noticed some missing signals were issuing errors for Sprite
and Node2D nodes. This just asks if the signals are supported.

### DIFF
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -301,14 +301,20 @@ func compile_str(p_str):
 
 	return ev_table
 
+func report_warnings(p_path, warnings):
+	var text = "Warnings in file "+p_path+"\n\n"
+	for e in warnings:
+		text += e+"\n"
+
+	print("warning is ", text)
+
 func report_errors(p_path, errors):
-	#var dialog = preload("res://demo/globals/errors.xml").instance()
 	var text = "Errors in file "+p_path+"\n\n"
 	for e in errors:
 		text += e+"\n"
-	#dialog.set_text(text)
+
 	print("error is ", text)
-	#main.get_node("layers/telon").add_child(dialog)
+
 	# The only way to - optionally - make errors matter
 	if ProjectSettings.get_setting("escoria/platform/terminate_on_errors"):
 		assert(false)
--- a/device/globals/item.gd
+++ b/device/globals/item.gd
@@ -361,13 +361,17 @@ func _ready():
 		area = get_node("area")
 	else:
 		area = self
-	if area is Area2D:
-		area.connect("input_event", self, "area_input")
-	else:
-		area.connect("gui_input", self, "input")
 
-	area.connect("mouse_entered", self, "mouse_enter")
-	area.connect("mouse_exited", self, "mouse_exit")
+	if ClassDB.class_has_signal(area.get_class(), "input_event"):
+		area.connect("input_event", self, "area_input")
+	elif ClassDB.class_has_signal(area.get_class(), "gui_input"):
+		area.connect("gui_input", self, "input")
+	else:
+		vm.report_warnings("item", ["No input events possible for global_id " + global_id])
+
+	if ClassDB.class_has_signal(area.get_class(), "mouse_entered"):
+		area.connect("mouse_entered", self, "mouse_enter")
+		area.connect("mouse_exited", self, "mouse_exit")
 
 	if events_path != "":
 		event_table = vm.compile(events_path)
--- a/device/globals/trigger.gd
+++ b/device/globals/trigger.gd
@@ -72,20 +72,24 @@ func _ready():
 		area = get_node("area")
 	else:
 		area = self
-	if area is Area2D:
+
+	if ClassDB.class_has_signal(area.get_class(), "input_event"):
 		area.connect("input_event", self, "area_input")
-	else:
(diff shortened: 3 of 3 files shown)


StraToN:
In which cases do these errors occur?

mjtorn:
On Thu, Jul 19, 2018 at 03:49:55PM +0000, Julian Murgia wrote:
>In which cases do these errors occur? 

I have a scene node in A Lunar Adventure that's maybe malformed by Escoria's
specs but works. I can get back with the details after the vacation, but
I think it makes sense to suppress these errors, because if the game is
broken, it's going to be visibly broken.

Item can be attached to pretty much anything, like the Polygon2D thing, so
maybe this approach is fine.

Alternatively someone could look into hardening the code to do type checking,
but meh ;P

StraToN:
@mjtorn My question was more relative to the fact I'm surprised these errors occur at all, since they can only happen if the node which `item.gd` script is attached to is an `Area2D`.

At least, your PR is very good to have anyway since it allows both `Area2D` and `Control` nodes to have this script attached, and behave the same way.

fleskesvor:
Another option could be to change the `else` clause to `elif area is Control` and add a new `else` which throws an error if the item is neither an `Area2D` nor a `Control`, instead of silently failing.

mjtorn:
But eg. Polygon2D does not inherit from Area2D. The case can of course be made that we should coach gamedevs into proper behavior, but I'm also a bit worried that we're (meaning mainly me) will be beating our (mine) head(s) against a wall over these slight changes.

If it's guaranteed we can always add an Area2D child into obscure cases, without any detriment to game development, I really don't care and would just fix that scene in A Lunar Adventure.

The pragmatic approach is to let the developer just see if the game works as expected, since a lot of the things that go on are "smoke and mirrors" anyway. That is, IMO, this is not a similar error case as missing translation identifiers or global identifiers or such.

@StraToN and @fleskesvor, just give me the word on what you want so we can get this going. Once my vacation is officially over, I don't want to wallow too long in PRs or in updating a game that will never generate revenue.

fleskesvor:
> But eg. Polygon2D does not inherit from Area2D.

Did you mean `Position2D`, which you added support for to use as camera targets, IIRC? A `Polygon2D` only gives an `Area2D` its shape when used for items.

My thinking in suggesting to explicitly check for `Area2D` and `Control` (and maybe `Position2D`?) is that it would make it easier to provide the developer with a meaningful error message in line with [our documentation](https://godotengine.github.io/escoria/items.html), instead of possibly allowing unsupported nodes which happen to have either of those signals.

What do you think, @StraToN?

mjtorn:
I may get back to this soon, but I do mean wonky items, not `Position2D`, which actually uses `target.gd` and not `item.gd`. The lander in room3 of A Lunar Adventure is technically broken but it looks good enough in the game. There is a strong argument for me fixing it, but it's just frustrating to spend time on a game that's now very close to a sunk cost and should be released out of early access, as the team should move on.

What if I add an `else:` that calls `vm.report_errors()`, would that be an acceptable compromise?

Though it would necessarily force me to spend time fixing that one item...

mjtorn:
@fleskesvor I updated the PR with a matching `target.gd` change and an `else:` to report errors.

Turned out the case in A Lunar Adventure was doable, so let's have this merged IMO.

fleskesvor:
I don't really have any objections to this, though using ClassDB like this feels a bit like blowing smoke to let unsupported nodes slip by to me, instead of explicitly checking for Area2D or Control. I'm still not sure if I completely understand the use case though. Is it to be able to register nodes and be able to control them with esc commands without them actually being "proper" items?

mjtorn:
I commented on IRC but leaving it here for posterity... This may be a bit iffy and hard to test as I dropped some of the "broken" Area2Ds from A Lunar Adventure, but I don't know that there's a problem either.

Area2D is very specific and required, but Control is such a base type that it can be almost anything. The `ClassDB` code simply extends the palette to make other node types possible, as long as they are configured correctly.

It also drops the assumption "If it isn't an `Area2D` it must be `Control`" by issuing a warning if things are broken.

Especially for items, it could make sense to have non-interactive items that can be controlled via ESC commands. I don't know what an example would be, maybe a crane/hook on the ceiling that moves around when you play with a lever? There's no reason for that hook to have Area2D and you could argue it has to be a `TextureRect` in that case, but it might as well be a `Sprite`.

Using a sprite makes more sense to me, because we were told by the Godot people that `Control`s are for UI elements and such.

I have to confess that I mix things up a bit in A Lunar Adventure because it wasn't always clear which to prefer and some of this Escoria code was written during or after assembling the game and I never had (and probably never will) have the time to visit it.

Hoping Iron Sky: Cold War will prefer Sprites all the way, unless there's a reason not to, because of that Control-is-UI idea.

mjtorn:
To make development on _Iron Sky: Cold War_ easier, and as most of the issues have been addressed, I'm thinking it's enough to be give out warnings here. `No input events possible for global_id lander`.

In _A Lunar Adventure_ this is fine, because at 1/6g and with the shape of the lander, the player can happily jump anywhere on it to reach the door. But still it needs to be an item to have eg. `use_custom_z`.

This way if we wanted to enforce having `Area2D`s everywhere, the effort should go into fixing #221 by someone who has the time and money required. I will just ignore the warning until then ;)

PR #212, opened 2018-08-02
Overload _find_sprites() to look for anything to modulate

mjtorn:
This is because eg. Spine nodes, although not officially supported,
are inherited from CanvasItem. In fact, CanvasItem is what implements
light modulation, and `_find_sprites()` is used for nothing else,
it makes sense to apply the light map everywhere.


fleskesvor:
Looks perfect to me. :+1: Thanks, @mjtorn.

PR #213, opened 2018-08-02
Fix crash when clicking Polygon2D

mjtorn:
If you have a Polygon2D in the scene and click it, Escoria
would call `get_overlapping_areas()` on it, which does not exist.

Address this by checking the method's existence before calling.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -170,6 +170,7 @@ func clicked(obj, pos, input_event = null):
 		walk_context = {"fast": input_event.doubleclick}
 
 	# If a background is covered by an item, the item "wins"
+	# and the winning item is chosen in item.gd by looking at the overlapping areas
 	if obj is esc_type.BACKGROUND:
 		var overlay = obj.get_child(0)  # Created by background.gd to intercept clicks
 		# Eg. Polygon2D does not have this method
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -171,9 +171,12 @@ func clicked(obj, pos, input_event = null):
 
 	# If a background is covered by an item, the item "wins"
 	if obj is esc_type.BACKGROUND:
-		for area in obj.get_child(0).get_overlapping_areas():
-			if area.has_method("is_clicked") and area.is_clicked():
-				return
+		var overlay = obj.get_child(0)  # Created by background.gd to intercept clicks
+		# Eg. Polygon2D does not have this method
+		if overlay.has_method("get_overlapping_areas"):
+			for area in overlay.get_overlapping_areas():
+				if area.has_method("is_clicked") and area.is_clicked():
+					return
 
 	# Hide the action menu (where available) when performing actions, so it's not eg. open while walking
 	if action_menu:
--- a/device/globals/item.gd
+++ b/device/globals/item.gd
@@ -96,6 +96,19 @@ func input(event):
 	# TODO: Expand this for other input events than mouse
 	if event is InputEventMouseButton || event.is_action("ui_accept"):
 		if event.is_pressed():
+			# See if we're allowed to call game.gd's `clicked()` method.
+			var bg = $"/root/scene/background"
+			var overlay = bg.get_child(0)  # Created by background.gd to intercept clicks
+
+			# Eg. Polygon2D does not have this method
+			if overlay.has_method("get_overlapping_areas"):
+				var overlapping_areas = overlay.get_overlapping_areas()
+
+				# The last overlapping area is the lowest in the node tree, it gets the event
+				# but verify `self` is contested by other areas, because eg. inventory items aren't
+				if self in overlapping_areas and self != overlapping_areas[-1]:
+					return
+
 			clicked = true
 
 			if event.button_index == BUTTON_LEFT:


fleskesvor:
What are the use cases for having a `Polygon2D` as a child of the background?

mjtorn:
As discussed with @StraToN it was the simplest way of making a lighting effect, where a stark light shines through the window. Furthermore we could animate it later.

This is because Godot's 2D doesn't have directional lighting.

fleskesvor:
I see. Does it have to be a child of the background though, or could it be a child of the scene?

Note that I'm not super happy with the way I implemented that logic to begin with, and I found a better way of keeping events from being passed through an Area2D, I'd scrap that logic in a heartbeat.

mjtorn:
There's not really much else to have as a parent, and I really can't see the value of having some generic `Node` placeholder either.

It's not impossible to imagine someone trying a similar trick and going "Buggy piece of shit" and buying Unity, because intuitively there's no reason for this not to work.

I think we're stuck with what we got until TEN LEGION OF SARDA^WCONTRIBUTORS come along and pick up some of the slack.

mjtorn:
@fleskesvor while waiting for ten legion of sardaukar coders, I think I got everything pretty well made.

fleskesvor:
> There's not really much else to have as a parent, and I really can't see the value of having some generic `Node` placeholder either.

My thought was that we'd avoid having to introduce further complexity into the framework by just having it be a child of the "scene" node, but I have no experience with this, so I don't understand why it's easier or only possible to implement as a child of "background" specifically. Hopefully, @StraToN can chime in on this.

fleskesvor:
So you have implemented it as:

```
scene
    └── background
        └── left_light (CollisionPolygon2D)
            ├── animation
            └── area (Area2D)
                └── CollisionPolygon2D
```

What I don't understand is why it's not implemented as:

```
scene
    ├── background
    └── left_light (Area2D)
        ├── animation
        └── CollisionPolygon2D
```

> As are the multiple overlapping `Area2D`s; I think the code used to work only for one `Area2D` over the background. Empirical sightings support this thought.

Yes, because there hasn't been a reason to support more than one `Area2D`. One `Area2D` can also have multiple collision shapes, if necessary, but usually you want the entire screen to be clickable, so a single `RectangleShape2D` should work for most purposes.

mjtorn:
> What I don't understand is why it's not implemented as:

It didn't occur to me. My brain told me they're tightly-enough coupled to the background - as background elements - that I would group them together. I'm in the middle of fifty things, all of them annoying, but testing the animations on those polygons moved down is a high priority now.

I would still advocate for the PR, because it makes for a more elegant node tree and does no harm. I will not, however, let this drag on forever so if all works out, we can leave that out.

> One Area2D can also have multiple collision shapes, if necessary, but usually you want the entire screen to be clickable, so a single RectangleShape2D should work for most purposes.

Sorry, but I don't understand at all. If I have multiple collision shapes on the lander item, one for the lander and one for the door, how does that interact with anything? `item.gd` is bound to the underlying `Area2D` so how would the click handlers know where to go?

With this implementation, the whole screen is clickable still, so there's no regression there.

And it's convenient (and intuitive!) to have the door area over the lander area, but I don't know the proposal from you that I don't understand is equally as convenient.

mjtorn:
@fleskesvor I tried rearranging the tree so that the polygons were parallel to background, the control panel and windows below that so they'd receive the click events, and took out the `if:` check and the player would use the fallback look action when looking at windows.

So the best thing to do is merge.

StraToN:
@mjtorn @fleskesvor I suppose you meant that the nodes branch is using a `Polygon2D` instead of a `CollisionPolygon2D`, right ?

```
scene
    └── background
        └── left_light (**Polygon2D**)
            ├── animation
            └── area (Area2D)
                └── CollisionPolygon2D
```

My first thought was agreeing with @fleskesvor concerning the fact that your left_light node (`Polygon2D`) should avoid being a child of the background node, precisely to avoid side effects. I still agree on this.

But on another hand, we can't forbid a node to have a child in Godot, that's how it is. We can't risk a crash when the user does something Godot allows him to do, imo.

So I see this PR as some kind of security: the regular way (no child for background node) still works, and @mjtorn 's way works as well. I'd explicitly advise to clearly avoid adding children to the background node in the documentation though.

fleskesvor:
> Sorry, but I don't understand at all. If I have multiple collision shapes on the lander item, one for the lander and one for the door, how does that interact with anything? `item.gd` is bound to the underlying `Area2D` so how would the click handlers know where to go?

Sorry, my sentence wasn't very clear. What I meant is that it makes sense for an item to have multiple shapes, eg several pictures hanging on a wall, with distinct shapes, but with the exact same interactions. For a background, however, you usually want to just make the entire screen handle input, so that's why handling more than one area and/or shapes is less useful.

@StraToN Yeah, I meant `Polygon2D` and I agree that it would be better to handle multiple kinds of node structures if it doesn't cost much to do so.

I thought of a couple of alternative implementations for solving this issue, that I also mentioned on IRC:

1. Don't use a loop over `get_overlapping_areas()` to check if an item was already clicked, by instead store a reference to the last input event object, and check if we've already handled the event.
2. Don't give the background an area and shape at all, but instead handle input via `_unhandled_input()`, since I think it's safe to assume that all unhandled input will be background clicks.

I don't have experience with `_unhandled_input()`, but I think it should be easy to implement and should probably have few side-effects.

mjtorn:
@StraToN When I made the polygons not be children of the background, I got the fallback action when clicking. This may be due to something else, though, but still we are in agreement that preventing these types of strange bugs is a thing.

As for @fleskesvor 's suggestions, I considered 1 already. I couldn't get any events to occur in a deterministic fashion, so I gave up. I don't know anything about `_unhandled_input()` either.

I think any and all effort is better spent in merging other PRs.

I will quickly verify - hopefully soon etc - that I can get rid of the second commit altogether by disabling the underlying lander's Area2D entirely, as it's passive. This means we should document that Area2Ds cannot overlap if you want expected behavior.

If all works out, I'll make a new PR with the Polygon2D code and documentation updates.

Please look at the other open PRs either way!

mjtorn:
Please merge #220 so this can be put to bed.

fleskesvor:
Yes, I think some groundwork is needed to better handle overlapping areas in general. These are issues caused by us moving from Control nodes to Area2D and not properly considering the implications of not blocking input. I'm pretty sure this might cause considerable headaches to future users of the framework too, so #221 should be a high priority.

mjtorn:
I'm closing this. It's been referenced elsewhere in case someone wants to see the broken algorithm at adf2894ba5048c86506ed17f30cefcf70b04981d.

PR #214, opened 2018-08-02
Fix item's error about missing terrain

mjtorn:
The error seems benign and temporary, probably related to
how the scene node tree is set up, so just checking `has_node()`
in advance makes it behave better.


mjtorn:
Should not have opened this, sorry, duplicate of #210

PR #215, opened 2018-08-03
Show tooltip, as controlled by AM, only when it has text

mjtorn:
The action-menu code interacts with showing the tooltip,
and if an item doesn't have tooltip text, but the tooltip
is configured to have a background, an empty background
would show up without this.

It's annoying and fortunately fixed now.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -88,9 +88,10 @@ func mouse_enter(obj):
 			vm.hover_begin(obj)
 	else:
 		if current_action != "" && current_tool != null:
-			text = tr(current_action + ".combine_id")
-			text = text.replace("%2", tr(tt))
-			text = text.replace("%1", tr(current_tool.get_tooltip()))
+			if tt:
+				text = tr(current_action + ".combine_id")
+				text = text.replace("%2", tr(tt))
+				text = text.replace("%1", tr(current_tool.get_tooltip()))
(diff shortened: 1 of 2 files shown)


fleskesvor:
Looks good to me. Is it possible to open the action menu while a tooltip is set, resulting in the tooltip not being hidden when it's supposed to as a result of this and the trigger in https://github.com/godotengine/escoria/blob/master/device/globals/action_menu.gd#L42 ?

mjtorn:
I have not seen an issue with tooltips not being hidden when the action menu is opened.

fleskesvor:
Ok. I'm wondering if it might be better to change

    if $"tooltip" and $"tooltip".text:
        $"tooltip".visible = p_visible

to

    if $"tooltip":
        $"tooltip".visible = p_visible and $"tooltip".text

to safe-guard against such scenarios though? Otherwise there's a chance calling `set_tooltip_visible(false)` to hide the tooltip might not be future-proof.

mjtorn:
@fleskesvor agreed with your reasoning, force-pushed.

Of course I haven't had the opportunity to actually test it, but there's no reason it would fail.

Maybe later today or just hit merge on the sucker.

mjtorn:
Hit it, flesk! I noticed no difference.

fleskesvor:
Great, thanks! :+1:

PR #216, opened 2018-08-06
Fix save load states

mjtorn:
I noticed an anomaly in A Lunar Adventure when continuing an old game.

Room 3 features an NPC going off to do his thing, but when loading, he stands at the start position.

This fixes the situation.

### DIFF
--- a/device/demo/game.esc
+++ b/device/demo/game.esc
@@ -1,3 +1,3 @@
-:load
+:start
 
 change_scene res://demo/rooms/test/test.tscn
--- a/device/globals/esc_compile.gd
+++ b/device/globals/esc_compile.gd
@@ -27,7 +27,7 @@ var commands = {
 	"walk_block": { "min_args": 2 },
 	"turn_to": { "min_args": 2 },
 	"set_angle": { "min_args": 2 },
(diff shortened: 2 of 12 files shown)


mjtorn:
FWIW I think this might fix all save games, though the demo of our next game probably won't feature manual saving.

mjtorn:
Pushed an extra commit that doesn't break starting a new game. Introducing a new event keyword is totally worth it!

If there's something trivial wrong with this, I would appreciate someone else squashing it, like if it's wording in the documentation or whatever, to facilitate a quick merge.

Sorry for not catching this all that time ago, it's because of stress.

mjtorn:
Though I did force-push an update for the demo scene ;) Disclaimer: I didn't get around to testing it, but it's used in a real game now, so it has to work at least a little.

mjtorn:
I rebased this on master now. Basically just uses regular angles as well; not doing that would have caused mad conflicts.

mjtorn:
There is always the risk that minor fixing will be required because the test cases aren't formally adequate, but this has been in use for quite a while with us. So I'm merging it, merci beaucoup!

PR #217, opened 2018-08-07
Update Spine documentation to reflect the current state of things

mjtorn:
Because the duration patch has been merged to the Spine module's master and `spine-to-godot-scene.py` has been in its own repository for, like, six weeks.

### DIFF
--- a/docs/spine.md
+++ b/docs/spine.md
@@ -6,14 +6,12 @@ that can be added to the `modules/` directory in your Godot source tree and buil
 with eg. `schedtool -D -e ccache scons platform=x11 tools=yes target=release_debug`
 or whatever you like.
 
-The script as referred to here is `spine-to-godot-scene.py`, which must be run from your `device/`
-directory. Do note it requires `python3-docopt`, so please install it using `pip3` or whatever
-you prefer.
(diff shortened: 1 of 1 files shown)


docs/spine.md
```diff
@@ -6,15 +6,12 @@ that can be added to the `modules/` directory in your Godot source tree and buil
 with eg. `schedtool -D -e ccache scons platform=x11 tools=yes target=release_debug`
 or whatever you like.
 
-The script as referred to here is `spine-to-godot-scene.py`, which must be run from your `device/`
```
fleskesvor: What about something like?

"For compatibility with Godot, run `spine-to-godot-scene.py`, (henceforth referred to as "the script"), found in the [Spine to Godot](https://github.com/mjtorn/spine-to-godot/] project, from your project root (typically the `device/` directory)."

It would both emphasize that "the script" has yet to be introduced (it confused me a little), and highlight your project.

mjtorn:
Something like "henceforth referred to as" made me think of lawyers, but I force-pushed a close approximation of your proposal.

fleskesvor:
Yeah, I agree that it's too formal, but I couldn't think of anything better. I think your changes are good though. :slightly_smiling_face:

PR #218, opened 2018-08-07
Remove inline comments from ESC commands

mjtorn:
It's annoying that depending on the command, inline comments may be ignored in later parsing and sometimes they cause side-effects when the second (or _n_ th) argument.

Just get rid of them up-front.

### DIFF
--- a/device/globals/esc_compile.gd
+++ b/device/globals/esc_compile.gd
@@ -331,6 +331,12 @@ func read_cmd(state, level, errors):
 		return
 	else:
 		params.remove(0)
+
+		# Remove inline comments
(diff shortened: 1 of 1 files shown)


fleskesvor:
This is a great idea. :+1: This should also remove the need to check for comments with `is_comment`, right?

mjtorn:
AFAICT it should not, because the parsing in the commit is very specifically for the comments in a command's arguments. It would never hit the same lines that `is_comment()` would hit; we need both.

fleskesvor:
Yes, that looks accurate. Thanks. :slightly_smiling_face:

PR #219, opened 2018-08-09
Refactor/streamline player's set_state for costumes

mjtorn:
I succumbed to the death of #197 by duplicating and re-creating a ton of animations in favor of the existing API.

The amount of animations and tracks is closing in on bullshit, even though I was able to reuse some things by copying nodes and editing them.

Be that as it may, this should be merged instead of #197 because it retains the existing way, but also makes the code a bit clearer and suppresses a useless error from `get_node(costume)` when the costume doesn't exist.

### DIFF
--- a/device/globals/player.gd
+++ b/device/globals/player.gd
@@ -100,7 +100,12 @@ func anim_get_ph_paths(p_anim):
 	return ret
 
 func play_anim(p_anim, p_notify = null, p_reverse = false, p_flip = null):
-	if p_notify != null && (!has_node("animation") || !get_node("animation").has_animation(p_anim)):
+	if not animation:
+		vm.report_errors("player", ["Animation not found for play_anim"])
+	if not animation.has_animation(p_anim):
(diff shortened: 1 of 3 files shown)


device/globals/player.gd
```diff
@@ -100,7 +100,12 @@ func anim_get_ph_paths(p_anim):
 	return ret
 
 func play_anim(p_anim, p_notify = null, p_reverse = false, p_flip = null):
-	if p_notify != null && (!has_node("animation") || !get_node("animation").has_animation(p_anim)):
+	if not animation:
+		vm.report_errors("player", ["Animation not found for play_anim"])
+	if not animation.has_animation(p_anim):
+		vm.report_errors("player", [animation.name + " does not contain " + p_anim])
+
+	if p_notify != null && (!animation || !animation.has_animation(p_anim)):
```
mjtorn: Before someone points out the redundancy here, I opted to keep the `!animation` checks in place, because not everyone necessarily has configured their projects to crash on `vm.report_errors`.

PR #220, opened 2018-08-14
Discourage background having child nodes

mjtorn:
This should replace #213 as it doesn't try too hard to fix problems that are not relevant right now, but does add a note for the documentation.

I couldn't find a good place to add a documentation note about overlapping Area2Ds not working, too tired also to think, but I think a new GH issue is fine to document that until someone comes along to solve it.

Good to go?


fleskesvor:
Thanks, @mjtorn. :slightly_smiling_face: Yes, this is perfect, and also thanks for opening #221, which should be a high priority for us.

PR #223, opened 2018-08-15
Some angle fixes

mjtorn:
Took a while to receive and integrate diagonal NPCs, but good things come to those who wait. Or maybe it's bugs. Or bug fixes. Anyway....

I noticed some time ago there's some strange skew with the player character's diagonal movements and wanted to see if it's true also for NPCs. It was. The demo scene doesn't really have what it takes to reproduce these problems, but I hope that doesn't prevent a merge.

This fixes what I could discern so far and replaces some less good checks with error-reporting.

### DIFF
--- a/device/globals/interactive.gd
+++ b/device/globals/interactive.gd
@@ -27,9 +27,9 @@ func set_light_on_map(p_light):
 		modulate(Color(1, 1, 1, 1))
 
 func _get_dir(angle):
-	if animations == null:
-		return -1
-	var deg = rad2deg(angle) + 180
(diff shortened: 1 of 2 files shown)


StraToN:
I definitely trust your feedback on this. I already had fixed this, or at least I thought I had succeeded. Maybe I didn't. Anyway, I think this one can be safely merged. Thanks a lot :)

mjtorn:
@StraToN merci! I was sure it was fixed as well, but the demo stuff would really benefit from clearer examples, as this was obviously less than obvious. If you play A Lunar Adventure, you can see things working at least ;)

PR #224, opened 2018-08-31
Implement dialog layer

mjtorn:
This fixes at least three known issues:

  1) Dialog not being clamped to viewport
  1) Dialog not showing when telon is faded out
  1) Dialog size being affected by zoom, obviously

I updated the demo as well, but it still looks ugly. I believe someone should take a look at it separately, like put some time into making all the positionings and alignments and such ok.

At least the demo shows that dialog size is not affected by zoom anymore, and that's the important part.

No documentation updates are required because the `dialog_layer` is in the `game*tscn` files. You break them, your fault.

### DIFF
--- a/device/demo/rooms/room2/room.esc
+++ b/device/demo/rooms/room2/room.esc
@@ -1,0 +1,4 @@
+:ready
+
+say player room2:"ROOM2 IS AMAZING!"
+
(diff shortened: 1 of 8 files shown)


mjtorn:
We have discussed this architecture many times and although the demo looks horrible, these changes make our Iron Sky games visibly better. Therefore I'm going to be a d*ck (quack!) and merge this immediately. Potential bug fixes should not stall the PR process on something that was decided months ago.

PR #225, opened 2018-09-03
Item default action

mjtorn:
This expands upon the object/item action functionality. Now it's possible to set a global default and if this functionality requires a double-click.

My opinion is that it should _always_ require a double-click, the gameplay feels horrible without it, I think this would be a faster merge if I don't change current behavior ;)


mjtorn:
Aaaand because I tested everything and know that it's more ok to have these merged when documented and not changing anything, I'll go ahead and hit it.

Ping @fleskesvor and @StraToN for opinions if it's ok to change this so it always requires a double-click. Please advise.

mjtorn:
And for me it's not a big deal. A Lunar Adventure will probably remain untouched by this and from now on I'll just check the checkbox.

I can't remember which games have the double-click thing. Gemini Rue iirc repeats the previous action, which was really confusing! Maybe Broken Age had default actions.

Now double-clicking on a door (despite having looking as the global default) feels like a good way to play. At least there are alternatives for different tastes now!

PR #226, opened 2018-09-04
Hopefully final clamping fixes

mjtorn:
Another fast-track merge coming in!

I noticed that dialog wasn't clamped correctly when running the development version of Iron Sky: Cold War in full screen, 4k resolution, the room zoomed, and the game resolution as 1080. This would be the default way for the game, so it can't be broken.

I got upset at how broken everything is so I channeled that into really digging into the clamping code and found myself fixing all the problems!

This also renames the clamper function for the action menu as iirc @fleskesvor requested me to do if this code is ever touched again.

If you don't have a game to test this with, the next release of Iron Sky: A Lunar Adventure will demonstrate this very clearly. Currently the tooltip overflows off the right-hand side of the screen and the action menu opens way off. Not sure when I have time for that release, though...

### DIFF
--- a/device/globals/action_menu.gd
+++ b/device/globals/action_menu.gd
@@ -12,14 +12,19 @@ func action_pressed(action):
 func target_visibility_changed():
 	stop()
 
-func check_clamp(click_pos):
+func clamped_position(click_pos):
+	var width = float(ProjectSettings.get("display/window/size/width"))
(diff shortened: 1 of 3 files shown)


mjtorn:
In the third room it overflows, to be more specific. But it does go derp in other cases as well, I'm sure.

PR #228, opened 2018-09-05
Support for separate end credits

mjtorn:
This does not include any new scenes, as someone may want to create them for the demo.


mjtorn:
And because this is so trivial, and works in production, I'll just hit it. We can look at possible course corrections later if required.

PR #229, opened 2018-09-11
Even more angle code

mjtorn:
I accidentally found out that a save game could still contain negative degrees. This was due to getting the angle from `rad2deg()` without the special compensation used elsewhere.

That lead me to refactor the code for both interactives and the player into the VM.

I tested this extensively by:
  * printing a ton of angle data out when walking the player
  *  having an npc walk between whatever items were available
  *  hacking a save game file with a `set_angle` and `turn_to` call for every angle.

The angles defined for both the player and the NPC are as follows:

```
const dir_angles = [0, 45, 90, 135, 180, 225, 270, 315, 360]
```

I also added some robustness in case someone tries to load a savegame with a broken angle value.

On top of that I fixed some setting of states (interactive) or playing animations (player) to make the code behave better.

### DIFF
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -658,6 +658,8 @@ func save():
 			var pos = objects[k].get_position()
 			ret.append("teleport_pos " + k + " " + str(int(pos.x)) + " " + str(int(pos.y)) + "\n")
 			if objects[k].last_deg != null:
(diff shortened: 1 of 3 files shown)


mjtorn:
There were some minor issues, so I force-pushed my way around them. I can't find the time to replay the entire game, but I did play the parts with issues, so I'm confident enough this works that it can be merged.

Any potential problems can be addressed as hotfixes to master, if it comes to that.

PR #232, opened 2018-10-09
Event flags

mjtorn:
As discussed with @StraToN on IRC, it's a bit annoying - especially for fallback events - that the player walks all across the room to say "meh".

This demonstrates how fallbacks, and custom events, can be made "telekinetic" with the `TK` flag.

It also contains a `FAST` flag for fast walking.

Once merged, it would be sweet to see what the community can come up with to make adventure games all the more awesome :)

### DIFF
--- a/device/demo/fallbacks.esc
+++ b/device/demo/fallbacks.esc
@@ -1,20 +1,20 @@
-:look
+:look | TK
 debug "look fallback!"
 say player look_fallback:"I can't look at that."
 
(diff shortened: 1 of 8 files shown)


mjtorn:
I played through the rough version we have for Iron Sky: Cold War's demo with this code and it didn't fail, so I'm merging it now. Any possible remaining bugs will be fixed separately.

PR #234, opened 2018-10-10
Event work

mjtorn:
There were some inexplicable issues with tooltips when trying to implement the `NO_TT` flag.

I had to make `run_event` use signals to get everything straight.

Please refer to the individual commit messages for more information.

### DIFF
--- a/device/globals/game.gd
+++ b/device/globals/game.gd
@@ -390,10 +390,10 @@ func fallback(obj, action, param = null):
 	if typeof(param) != typeof(null):
 		comb = action + " " + param.global_id
 		if comb in fallbacks:
-			vm.run_event(fallbacks[comb])
+			vm.run_event(comb, fallbacks[comb])
 			return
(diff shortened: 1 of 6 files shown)


StraToN:
Very cool ! I like the idea of formalizing Escoria events as a class. It may even be interesting to externalize it it its own script if necessary.

mjtorn:
Nice, thanks!

I was thinking of `NO_TT` (and maybe `NO_HUD` is around the corner if there's a strong-enough use case. or maybe `CUTSCENE` that disables hud and input), but then I started thinking that a clean event object could be useful in many ways...

Like we've talked about testing on IRC, that it would be useful to have something that runs events from the outside. Making them easy to create and inject would be a big win.

I think the code could be further developed to pass instances more around; now it's a bit like constructing them on the go. That will have to wait, because the immediate itch is scratched, but it's something worth keeping in mind.

PR #235, opened 2018-10-11
Megacommit to make all events EscoriaEvent

mjtorn:
This can't be properly split up so that anything works, and most of the changes are restoring the old function signatures.
    
The more important change is that the event table contains actual event instances.


StraToN:
Seems very cool ! It actually makes the code more readable since many functions get 1 less parameter this way.

PR #236, opened 2018-10-11
Split load/run game to enable setup event

mjtorn:
This adds the `:setup` event with requisite changes to the VM.

The idea is that active NPCs, or the player, can be teleported to their correct places before the loaded
scene is ever shown.

This removes the problem of having the player (or NPC) "flash" in the wrong place before appearing in the correct one.


StraToN:
Seems cool :)

mjtorn:
Merci!

PR #239, opened 2018-10-14
Add optional progress meters to resource_queue

mjtorn:
This will hopefully help with the bug when exiting
a room while the resource_queue preloader is running.

Any touching the queue, even asking if it's ready, will
cause a deadlock or some other kind of stale mutex, so
events are (probably) required for anything to have a chance.


mjtorn:
This is clean to go, but take the PR route in case someone wants to comment.

PR #244, opened 2018-10-15
Moved things to clean interactive.gd

santouits:
Made the interactive.gd have the bare minimal to be considered an object by the vm.

1) Moved everything basic for an interactive object to the interactive.gd and everything else to item.gd that should be cleaned next.
2) Made trigger.gd extend item.gd instead for now.

### DIFF
--- a/device/globals/interactive.gd
+++ b/device/globals/interactive.gd
@@ -1,165 +1,33 @@
 extends Node
-export(Script) var animations
 
-var terrain
-var walk_path
-var walk_context
-var moved
-var last_scale = Vector2(1, 1)
-var last_deg = null
-var last_dir = 0
-var animation
-var state = ""
-var walk_destination
-var path_ofs
-var pose_scale = 1
-var task
-var sprites = []
+export var global_id = ""
+export(String, FILE, ".esc") var events_path = ""
+export var active = true setget set_active,get_active
 
-export var speed = 300
-export var scale_on_map = false
-export var light_on_map = false setget set_light_on_map
+var event_table = {}
 
-# This appears to be slightly faster in _process than checking
-# `self is Node2D` or something similar on every loop
-onready var self_has_z_index = self is Node2D
+func run_event(p_ev):
+	vm.emit_signal("run_event", p_ev)
 
-func set_light_on_map(p_light):
-	light_on_map = p_light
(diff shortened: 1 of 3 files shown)


device/globals/interactive.gd
```diff
@@ -1,165 +1,36 @@
...
 
-export var speed = 300
-export var scale_on_map = false
-export var light_on_map = false setget set_light_on_map
+var event_table = {}
 
-# This appears to be slightly faster in _process than checking
-# `self is Node2D` or something similar on every loop
-onready var self_has_z_index = self is Node2D
+func register_to_vm():
```
mjtorn: This function feels a bit redundant... I mean, calling `vm.register_object(global_id, self)` is pretty obvious if it's spelled out in `_ready()`.

The decision on this is @StraToN 's, but I'm not a huge fan of these types of indirections. The bigger value might be in adding debug prints temporarily, but that doesn't necessarily require a wrapper function. Is there another solid reason?

mjtorn:
Generally I like the idea of this, good stuff!

device/globals/interactive.gd
StraToN: I agree with @mjtorn . If `register_to_vm()` is to be called only in `_ready()` function and only contains a one liner, then directly put this one liner in `_ready()` function

santouits:
I removed the function.  And I think I corrected the whitespace line.

mjtorn:
@santouits I've been busy with paid work and pushing out an updated version of [Iron Sky: A Lunar Adventure](https://gamejolt.com/games/ironsky-lunar-adventure/339236) but hopefully I can rebase a game onto this PR and see what happens.

Sorry about the delay. I remember starting off with Escoria and how disheartening it could be to wait for feedback etc, but no one's directly paid to do any of this. Not even paid to make games _per se_. So we have to prioritize to get the bills paid.

Feel free to try Lunar Adventure while waiting ;)

santouits:
Don't worry, take your time, there is no need to hurry, really.
I played your game and I liked the scifi feeling though I guess I missed references because I haven't watched the movie.

mjtorn:
I went ahead and committed 3258ee7.

Good job, thank you, even if we didn't make trigger an item!

PR #246, opened 2018-10-17
Automatic angles

mjtorn:
This has been a desired feature for quite a while, even more so now that we have telekinetic fallbacks.

I believe @fleskesvor wanted this to be properly documented if it gets implemented, and now that we have the beautiful docs/ directory, I believe this condition is satisfied.

Resolving the angles isn't always perfect, and I'm not sure why.

It could be because large `Area2D` style objects can have their coordinates in seemingly unlikely places.

It could also be that the player's idle angles should not be in 45 degrees exactly but prefer showing diagonal idles if they look more natural.

Or maybe it's because your brain tells you you're looking at something 3D but the angles are 2D.

Be that as it may, this is as solid as they come and I'll hit merge on it soon.

Complaints and grievances will be addressed after the fact.


device/globals/item.gd.orig
```diff
@@ -0,0 +1,397 @@
+extends "res://globals/interactive.gd"
```
StraToN: Is this file required ?

PR #249, opened 2018-11-01
Revert faster setup code

mjtorn:
This is because we encountered a bizarre camera alignment
bug when trying to pull off something cool with positioning
and zooming.

There is no obvious reason why this is the code that breaks,
unless it's related to the camera operations not being blocking.

Room changes are still quite fast with `queue_resource`, which
is the proper way to do it. Fixing the thread deadlock bug in
it would probably be the smarter way than trying to figure this
out.

Restored the relevant parts of b246a5761bfa6c80180e58ffbbf4910b014dddfc,
as it was reverted only to make the sequence free from conflicts.

Revert "Fix `not LEAVE_BLACK` behavior for new code"
This reverts commit 4137c1b61e65320274f75632ec87e63c702f4a1c.

Revert "Fake a :setup for scenes without one to keep up the flow of events"
This reverts commit 2625a144c837661b8f3e67feb217c59cb20522b1.

Revert "Yielding on any event causes lag, don't do it"
This reverts commit 05d93233f2644c9ddc08c88548ca1303b83884eb.

Revert "Make it possible to have a scene without events_path"
This reverts commit b246a5761bfa6c80180e58ffbbf4910b014dddfc.

Revert "Kick off :setup before previous scene is cleared"
This reverts commit 9ffda4fd7b4a87e66c44beace4533f3946915449.

### DIFF
--- a/device/globals/global_vm.gd
+++ b/device/globals/global_vm.gd
@@ -423,8 +423,11 @@ func event_done(ev_name):
 	assert ev_name == running_event.ev_name
 	printt("event_done: ", running_event.ev_name, running_event.ev_flags)
 	if ev_name == "setup":
-		if not "LEAVE_BLACK" in running_event.ev_flags:
+		if not "LEAVE_BLACK" in running_event.ev_flags or not "ready" in game:
 			main.telon.cut_to_scene()
+
+		if "ready" in game:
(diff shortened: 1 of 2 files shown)


mjtorn:
Relevant to #238

PR #250, opened 2018-11-14
Implement ! as a way of ending dialog

mjtorn:
The purpose of this is to have a clean way of hiding the HUD during
dialog. The VM gets `in_dialog` in its context and hides the HUD.

When the `!` (alias `end_dialog`) command is called, the context
is reset. The HUD is also restored unless the event flag `NO_HUD`
is set.

### DIFF
--- a/device/globals/dialog_dialog.gd
+++ b/device/globals/dialog_dialog.gd
@@ -31,8 +31,6 @@ func timer_timeout():
 
 # called from global_vm.gd::dialog() function
 func start(params, p_context):
-	vm.set_hud_visible(false)
-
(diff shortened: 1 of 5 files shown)


mjtorn:
PR comment: the VM context is there in case someone gets inventive
and wants to create checking if we've been in a dialog state for suspiciously
long or in another state where it should be impossible or whatever.

Most likely when Escoria becomes a true asset with debug facilities, that's
when it'll be most useful, in simply viewing the VM state!

StraToN:
Sounds good :)
Thanks !

PR #251, opened 2018-11-20
Telon, music and interactive refactor

mjtorn:
Removes dead code and makes `telon` and `bg_music` work without being items.

This does not only clean up the code, but also helps to pave the way to stronger-typed items (which I believe are required in Godot 3.1).

There might even be a performance increase from not having all these things call `_process()` through `item.gd` all the time!


mjtorn:
The force-push was removing `var clicked` from telon. It was left over when I was figuring out which features of `item.gd` would still be required.

mjtorn:
Latest force-push was basically fixing a complaint in telon and bg_music about `tree_exited` already being connected when the game is cleared. And making sure the background music is always played correctly.

PR #252, opened 2018-11-21
Signalification effort

mjtorn:
This is a loose bag of commits I made while refactoring the code for signal support.

Commits I-III are basically the signal part.

Commit IV is a hack with a workaround for a tooltip bug :(

Commit V-VI is drag and drop removal, so no one accidentally believes Escoria supports dragging and dropping.

Commit VII fixes an old bug where an item could be permanently stuck in a clicked state (by basically dragging and dropping off it) and therefore prevent the player from moving. Well, technically the player would only move to the same point or another item.

Commit VIII is a tooltip content bug fix.


device/globals/background.gd
```diff
@@ -4,17 +4,33 @@ export var action = "walk"
...
+			emit_signal("left_click_on_bg", self, pos, event)
 		elif (event.button_index == BUTTON_RIGHT):
-			emit_right_click()
+			emit_signal("right_click_on_bg")
 
 func get_action():
 	return action
 
 func _init():
+	add_user_signal("left_click_on_bg")
```
StraToN: These can simply be defined on top of the script like so:
``` gdscript
signal left_click_on_bg
signal right_click_on_bg
```

device/globals/item.gd
```diff
@@ -479,6 +452,12 @@ func _find_sprites(p = null):
 	for i in range(0, p.get_child_count()):
 		_find_sprites(p.get_child(i))
 
+func _init():
+	add_user_signal("left_click_on_item")
```
StraToN: Same, signals can be defined with `signal` keyword.

device/globals/background.gd
mjtorn: Thanks, didn't know that :) Updated.

device/globals/item.gd
mjtorn: Thanks, didn't know that :) Updated.

PR #253, opened 2018-11-26
Mouse events, state in VM, tooltip refactor

mjtorn:
Sorry about the extreme PR, but I figured it's best I show what I've been up to lately rather than split it into pieces only to find a `squash` here or a `fixup` there to early commits caused things not to work in isolation.

I played through the current state of our _Iron Sky: Cold War_ demo with this, and it did work, though I didn't put mega effort into breaking it.

I have to tweak _Iron Sky: A Lunar Adventure_ a bit more to get it going, but that's not for these changes, but for how the dialog selection was renewed some time ago. Preliminary testing on _A Lunar Adventure_ looks good so far, even though it uses a different HUD style, yay!

@StraToN if it's too much, I can split this into pieces, but let me know if that's required.


mjtorn:
I'll get back to you on the divinity once I'm done patching, and played through _A Lunar Adventure_, but thanks either way, it means a lot :)

PR #254, opened 2018-11-29
Audio management

MKosonen:
If an item has an AudioStreamPlayer node (named "audio"), esc scripts can use it to play audio files. As an example, the box item now plays a sound when it's opened.


mjtorn:
Thank you very much! This works, and inspired me to expand on it a bit, which was made real easy by this code. Instead of going back and forth in the PR, I'm merging it and will commit something more into master.

PR #255, opened 2018-12-15
Camera tween

mjtorn:
Based on some discussion with @StraToN, some example code of his (merci!), and a Youtube video I don't have the link for, it makes sense to use `Tween` nodes instead of the calculated-step-based camera movement/zoom.

This PR reimplements the backend.

### DIFF
--- a/device/globals/camera.gd
+++ b/device/globals/camera.gd
@@ -1,5 +1,7 @@
 extends Camera2D
 
+var default_limits = {}  # This does not change once set
+
 var speed = 0
 var target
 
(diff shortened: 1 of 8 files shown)


mjtorn:
Added a commit for the new feature.

mjtorn:
@StraToN I'm not 100% sure if `camera_push` should set `target` or not. In our current work it certainly is what we want, but you could also argue the opposite position.

To keep things simple, I'm going to assume this is usually the desired behavior, because the camera is easy and smooth to move around.

When the language gets rewritten to use keyword arguments, we can add a `set_target=true` thing or someone who disagrees can open an issue.

mjtorn:
I'll add an option to set the `TransitionType` and not force the target for `camera_push` to be a camera as soon as I can.

This will help with the blocking of the actors, as the camera can be pushed with `LINEAR` for a better shot.

Hopefully a big force-push will come soon!

mjtorn:
Frustratingly this still skates off the camera limits :( I haven't been able to fix it, but I left the limit-setting method in the PR because I like it anyway ;)

mjtorn:
I removed the `can_zoom` variable in that force-push; it wasn't really necessary.

mjtorn:
Since there were no major changes since @StraToN said it's ok on IRC, I'm going to risk merging this.

Some skating can maybe be prevented by moving the camera somewhere else than its final position before zooming in, but that's awful and should have some kind of detection algorithm that moves the camera away from the limits if it's about to hit it.

PR #257, opened 2018-12-21
Scalenode terrain

mjtorn:
Because the gradient approach is hard to deal with and slow in its implementation, have an alternative to use lerping instead.

### DIFF
--- a/device/globals/terrain_scalenodes.gd
+++ b/device/globals/terrain_scalenodes.gd
@@ -1,0 +1,84 @@
+tool
+
+extends "terrain_base.gd"
+
+export(int, "None", "Lightmap") var debug_mode = 1 setget debug_mode_updated
+
+var _texture_dirty = false
+var texture
+
+var scale_nodes = []
+
+onready var scale_min = $"scale_min"
+onready var scale_max = $"scale_max"
+
+func debug_mode_updated(p_mode):
+	debug_mode = p_mode
+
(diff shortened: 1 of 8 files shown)


mjtorn:
I think the main thing missing is documentation, but plz @StraToN check this out. I wrote it in my typical deadline-induced panic so there might be problems I just don't see.

At least our player moves smoothly and it should perform at least twice as good as the gradient scalemap thing.

device/globals/terrain_scalenodes.gd
```diff
@@ -0,0 +1,89 @@
...
+
+	update()
+
+static func sort_by_y(a, b):
+	return a.global_position.y < b.global_position.y
+
+# Return a "scale range" immediately based on the interpolated scale size
+func get_terrain(pos):
+	# printt("Called", pos)
+	var distant_color = Color(1, 1, 0, 1)
```
mjtorn: Please disregard this. I will force-push it out later.

mjtorn:
Force-pushed removal of debug print

StraToN:
I had developed a similar approach in my pre-Escoria demo, so I definitely agree with the principle. It simply misses documentation, other than that it's cool, thanks!

mjtorn:
Thanks @StraToN ! I'm merging now with some documentation, so it doesn't get in the way of the next big thing :)

PR #265, opened 2019-04-05
Replace telon's input-catching with acceptable inputs

mjtorn:
This fixes the bug where the tooltip (following the mouse),
when moved during dialog, would show the old tooltip text.

The reason for that was how telon blocked not just click events
but also the item hover events.

Because we actually care only about skipping dialog, make it
a state and use the main input handler to deal with it.


mjtorn:
@StraToN there should probably be demos for tooltip-follows-mouse as well as what currently exists.

Do you foresee problems with this? Seems to work here, though I haven't played through a game with it yet.

mjtorn:
@StraToN btw I think a new interface in ESC for this should be like `accept_input ONLY_SKIP`. And add a `NONE`. So I'll get back to this soon. Opinions on the interface? The change in general?

StraToN:
@fire What other type of actions could be inputed by the player over the dialog apart from skipping it?
If any other can be thought of, then `accept_input` sounds ok. Else, `accept_skip [YES/NO]` would probably be clearer. But again, only if no other kind of action exists.

mjtorn:
@StraToN `InputEventMouseMotion` which is required for tooltips-following-mouse to work.

The telon implementation for disabling input disabled everything but clicks to telon, meaning no item could be hovered during dialog, leading to the bug where we see the wrong tooltip.

Also the telon implementation for `cut_scene telon disable_input` and `cut_scene telon enable_input` work on a `Viewport` level, meaning that even if we had only `accepted_inputs.ACCEPT_INPUT_SKIP` instead of telon, any cutscene which disables input would *also* display the wrong tooltip.

So far in my experimentation, this commit is pure gold. It does not mess with `Viewport` or control nodes and allows the tooltip to work under all circumstances. It also denies opening the menu in `accept_input NONE` and `accept_input SKIP`, as well as elegantly disables savegames in those conditions so we narrow the risk of saving a messed-up state.

StraToN:
@mjtorn Ah, you meant it as a direct information for the node itself (ie. mouse pass/skip...), not in a functional way (ie. dialog skip). My bad, sorry for this misunderstanding. Well, seems ok then, since it can be interesting to have different behaviors for each dialog.

Issue #267, opened 2019-05-01
Get multiple errors on Godot 3.1 and 3.1.1


Alyks:
Would you know when an official support for Godot 3.1.1 will be available? I hope there is a project to port it to recent version. I've tried Escoria 3.0 on Godot 3.1 and 3.1.1 but there's multiple errors on files before i can play with it and impossible to run a preview of the Daïza game from IDE.  I understand that the latest version of Escoria is only tested on Godot 3.0 and i've already seen this thread **[#237](https://github.com/godotengine/escoria/issues/237)** about similar issue on Godot 3.1.

mjtorn:
I had to revert our current development back to 3.0, because the Spine module caused a segfault every time the game was quit. This is why the 3.0 branch is more active than 3.1/master, as @StraToN cherry-picks the changes when he can.

The Daïza game is completely unsupported at the moment, as no one has the time or money (you could pay me or my company if you want; let me know and we can set something up) to bring it up to speed. IIRC it also has some worst practices, like using `Control` nodes for game elements.

Using `Control` should be banned entirely. There are corner-cases, like having an always-open non-blocking inventory, which currently pretty much has to be `Control`-based, making it impossible to animate an inventory item.

Likewise it would be nice to see if the action menu could be `Node2D`-based, because with a tooltip that follows the mouse, you will see the tooltip after selecting an action (until you move the mouse), because `Control` captures mouse movement and therefore doesn't see that the mouse exited the item while moving to the action button.

Either way, you might want to look at the demo rooms provided in the 3.1 branch.

You can also cherry-pick commits from the 3.0 branch and if they work for you, make PRs to pull them into 3.1. I doubt there'd be much bureaucracy around the matter if you see that everything works.

I should probably drop a quick comment on the issue you mentioned as well...

And, as always, unfortunately too busy struggling with our own studio's business to do pro-bono work for old tutorials. I wish it wasn't so, but incomes and expenses need to be balanced, and paid contract gigs to update them would help a lot.

Alyks:
I completely understand.  Thank you for your explanation. I was a bit lost and now this help me make decisions on my current projects.

jeancallisti:
Hello! For the record, for the last few years my (tiny) team and I have been developing a full-length point-n-click game with AGS (still in progress) and I'm considering scaling up to a better engine, like Godot-Escoria. If that can be an incentive to make Escoria fully compatible with Godot 3.1.1... this is just a shot in the dark.

[...]

PR #268, opened 2019-05-03
for 3.1 branch - rename local variables that shadow globals

santouits:
fix the errors when running the main scene with godot 3.1.1 in #267


mjtorn:
o_O I could swear I would have caught those issues when doing the 3.1 port. Unless 3.1.1 is stricter or something.

IMO this is a no-brainer merge, what do you think, @StraToN ?

StraToN:
Indeed. Thanks @santouits !

PR #281, opened 2020-03-25
Show off screenshots in the README

Beuc:
Since the asset lib screenshots are stored in the repo, we can show them off in github as well ^_^


StraToN:
Nice idea indeed :)
Thanks!

PR #332, opened 2021-07-28
Fix tooltip_node wrong references

### LINKED ISSUE #302: Fix missing tooltip_node reference in ui_mouse_icons
Because of the latest work, a reference to tooltip_node got lost in the mouse icons UI.

StraToN:
Fixes #302

I had to move the tooltip_node reference from ESCGame to ESCMain because of the loading order induced.
I am not sure this is right though so I'm open to suggestions.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_tooltip.gd
+++ b/addons/escoria-core/game/core-scripts/esc_tooltip.gd
@@ -1,46 +1,69 @@
+# A tooltip displaying <verb> <item1> [<item2>] 
+
 tool
 extends RichTextLabel
 class_name ESCTooltip
 
-func get_class():
-	return "ESCTooltip"
-
-
 # Infinitive verb
 var current_action: String
+
 # Target item/hotspot
 var current_target: String setget set_target
+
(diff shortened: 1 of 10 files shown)


dploeger:
What was your problem with the loading order? I simply corrected the reference in my Safehouse fork and it worked. 🤔

StraToN:
When the room loads, the ESCTooltip scene _ready() command runs and tries to set 
```
escoria.main.current_scene.game.tooltip_node = self
```
But escoria.main.current_scene is null at this moment of the execution so it cannot hold the reference.

dploeger:
Ahhh! So that's why it was lost after the compiler was rewritten.

How about we use a signal? Something like "scene_ready" or something and listen to it in the Tooltip? The process responsible for loading the scene will emit it once its ready.

StraToN:
We could do that indeed. I'll come up with a correction commit a bit later

dploeger (approved):

dploeger:
Great work! Thanks!

StraToN:
Thanks! :)

PR #356, opened 2021-08-04
Add default behaviors in ESCGame click handlers on BG and items.

### LINKED ISSUE #301: Call left mouse button function in game.gd by default
Original author: @dploeger 

The mouse handler functions in ESCGame should call the left click mouse handler function by default, so that for example if all mouse buttons on the background lead to the player walking to the point, only one function has to be implemented and all other functions in esc_game.gd call that.

Also, functions like `left_click_on_item` should call the default function:

```
func left_click_on_item(item_global_id : String, event : InputEvent) -> void:
	escoria.do("item_left_click", [item_global_id, event])

func right_click_on_item(item_global_id : String, event : InputEvent) -> void:
	escoria.do("item_right_click", [item_global_id, event])

func left_double_click_on_item(item_global_id : String, event : InputEvent) -> void:
	escoria.do("item_left_click", [item_global_id, event]) 
```

StraToN:
Fixes #301 

This PR adds default behaviors in these functions:
- left_click_on_bg()
- right_click_on_bg()
- left_double_click_on_bg()
- left_click_on_item()
- right_click_on_item()
- left_double_click_on_item()

So that if these functions are *not* overridden in user's game.gd script, the default behavior will be used.
NB: if these function are overridden in user's game.gd script, these default behaviors will not be called (unless explicitly called using parent script call by adding a dot before the function: `.left_click_on_bg()`)

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_game.gd
+++ b/addons/escoria-core/game/core-scripts/esc_game.gd
@@ -50,7 +50,7 @@ func _draw():
 #
 # - position: Position clicked
 func left_click_on_bg(position: Vector2) -> void:
-	pass
+	escoria.do("walk", ["player", position])
 
 
 # Called when the player right clicks on the background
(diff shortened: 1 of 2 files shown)


dploeger (approved):

StraToN:
Thanks!

PR #357, opened 2021-08-04
Add set_angle 'immediate' parameter.

### LINKED ISSUE #340: Implement immediate parameter in ESCMovable#set_angle
The immediate parameter is currently not implemented in ESCMovable#set_angle, so that only the direct angle will be shown and not the intermediate animations between the current and the target angle.

StraToN:
Fixes #340 

Modified fade_in transition in main.gd to happen between :setup and :ready so that :ready don't start before fade_in is finished.

### DIFF
--- a/addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
+++ b/addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
@@ -59,7 +59,7 @@ onready var task = MovableTask.NONE
 # the destination position was reached
 func _ready() -> void:
 	parent.add_user_signal("arrived")
-
+	
 
 # Main processing loop
 #
@@ -410,37 +410,22 @@ func set_angle(deg: int, immediate = true) -> void:
 		last_deg = deg
 		var target_dir = _get_dir_deg(deg, parent.animations)
 		
-		if target_dir - current_dir > 1:
-			var way_to_turn = get_shortest_way_to_dir(current_dir, target_dir)
-		
-			var dir = current_dir
-			while dir != target_dir:
-				dir += way_to_turn
-				if dir >= parent.animations.dir_angles.size():
-					dir = 0
-				if dir < 0:
-					dir = parent.animations.dir_angles.size() - 1
-				
-#				if parent.animation_sprite.animation != \
-#						parent.animations.idles[last_dir].animation:
-				parent.animation_sprite.play(
-					parent.animations.idles[dir].animation
-				)
-				yield(parent.animation_sprite, "animation_finished")
-				pose_scale = -1 if parent.animations.idles[dir].mirrored else 1
+		var way_to_turn = get_shortest_way_to_dir(current_dir, target_dir)
+	
+		var dir = current_dir
+		while dir != target_dir:
+			dir += way_to_turn
+			if dir >= parent.animations.dir_angles.size():
+				dir = 0
+			if dir < 0:
+				dir = parent.animations.dir_angles.size() - 1
 			
-				
+			parent.animation_sprite.play(
+				parent.animations.idles[dir].animation
+			)
+			yield(parent.animation_sprite, "animation_finished")
+			pose_scale = -1 if parent.animations.idles[dir].mirrored else 1
 			
-		else:
-			# The character may have a state animation from before, which would be
-			# resumed, so we immediately force the correct idle animation
-			if parent.animation_sprite.animation != \
-					parent.animations.idles[last_dir].animation:
-				parent.animation_sprite.play(
-					parent.animations.idles[last_dir].animation
-				)
-			pose_scale = -1 if parent.animations.idles[last_dir].mirrored else 1
-		
 	update_terrain()
 
 
@@ -448,7 +433,7 @@ func set_angle(deg: int, immediate = true) -> void:
 func _get_angle() -> int:
 	return parent.animations.dir_angles[last_dir].animation
 
-	
+
 # Return the shortest way to turn from a direction to another. Returned way is
 # either: 
 # -1 (shortest way is to turn anti-clockwise) 
@@ -467,13 +452,13 @@ func get_shortest_way_to_dir(current_dir: int, target_dir: int) -> int:
 	if current_dir < 0 or current_dir > parent.animations.dir_angles.size() - 1:
 		escoria.logger.report_errors(
 			"esc_movable.gd:get_shortest_way_to_dir()",
-			["Invalid direction (current_dir) " + str(current_dir)]
+			["Invalid direction (current_dir) %s" % str(current_dir)]
 		)
 	
 	if target_dir < 0 or target_dir > parent.animations.dir_angles.size() - 1:
 		escoria.logger.report_errors(
 			"esc_movable.gd:get_shortest_way_to_dir()",
-			["Invalid direction (target_dir) " + str(target_dir)]
+			["Invalid direction (target_dir) %s " % str(target_dir)]
 		)
 	
 	if current_dir == target_dir:
@@ -481,7 +466,7 @@ func get_shortest_way_to_dir(current_dir: int, target_dir: int) -> int:
 	
 	var internal = false
 	if max(current_dir, target_dir) - min(current_dir, target_dir) \
-			< parent.animations.dir_angles.size()/2:
+			< parent.animations.dir_angles.size() / 2:
 		internal = true
 	else:
 		internal = false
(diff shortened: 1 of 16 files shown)


addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
```diff
@@ -379,27 +383,111 @@ func is_angle_in_interval(angle: float, direction_angle: ESCDirectionAngle) -> b
...
+			parent.animation_sprite.play(
+				parent.animations.idles[last_dir].animation
+			)
+		pose_scale = -1 if parent.animations.idles[last_dir].mirrored else 1
+	else:
+		var current_dir = last_dir
+		last_deg = deg
+		var target_dir = _get_dir_deg(deg, parent.animations)
+		
+		if target_dir - current_dir > 1:
```
dploeger, line 413: What if the target_dir is smaller than current_dir? Then this would be < 1...

Or am I not getting something here?

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
```diff
@@ -379,27 +383,111 @@ func is_angle_in_interval(angle: float, direction_angle: ESCDirectionAngle) -> b
...
+		
+			var dir = current_dir
+			while dir != target_dir:
+				dir += way_to_turn
+				if dir >= parent.animations.dir_angles.size():
+					dir = 0
+				if dir < 0:
+					dir = parent.animations.dir_angles.size() - 1
+				
+#				if parent.animation_sprite.animation != \
```
dploeger, line 424: Please remove this commented code

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
```diff
@@ -379,27 +383,111 @@ func is_angle_in_interval(angle: float, direction_angle: ESCDirectionAngle) -> b
...
+#						parent.animations.idles[last_dir].animation:
+				parent.animation_sprite.play(
+					parent.animations.idles[dir].animation
+				)
+				yield(parent.animation_sprite, "animation_finished")
+				pose_scale = -1 if parent.animations.idles[dir].mirrored else 1
+
```
dploeger, line 431: Please remove the extra empty lines here.

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
```diff
@@ -379,27 +383,111 @@ func is_angle_in_interval(angle: float, direction_angle: ESCDirectionAngle) -> b
...
+#						parent.animations.idles[last_dir].animation:
+				parent.animation_sprite.play(
+					parent.animations.idles[dir].animation
+				)
+				yield(parent.animation_sprite, "animation_finished")
+				pose_scale = -1 if parent.animations.idles[dir].mirrored else 1
+			
+				
+			
+		else:
```
dploeger, line 434: This is duplicated. Better change the condition in line 396 so that immediate changes and 1-step direction changes use the same code.

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
```diff
@@ -379,27 +383,111 @@ func is_angle_in_interval(angle: float, direction_angle: ESCDirectionAngle) -> b
...
+# - target_dir: integer corresponding to the target direction as defined in
+# the attached ESCAnimationResource.directions.
+#
+# *Returns*
+# Integer: -1 (anti-clockwise), 1 (clockwise) or 0 (no movement needed).
+func get_shortest_way_to_dir(current_dir: int, target_dir: int) -> int:
+	if current_dir < 0 or current_dir > parent.animations.dir_angles.size() - 1:
+		escoria.logger.report_errors(
+			"esc_movable.gd:get_shortest_way_to_dir()",
+			["Invalid direction (current_dir) " + str(current_dir)]
```
dploeger, line 470: Use GDScript format strings here instead of +.

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
```diff
@@ -379,27 +383,111 @@ func is_angle_in_interval(angle: float, direction_angle: ESCDirectionAngle) -> b
...
+		)
+	
+	if target_dir < 0 or target_dir > parent.animations.dir_angles.size() - 1:
+		escoria.logger.report_errors(
+			"esc_movable.gd:get_shortest_way_to_dir()",
+			["Invalid direction (target_dir) " + str(target_dir)]
```
dploeger, line 476: Use GDScript format strings here instead of +.

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
```diff
@@ -379,27 +383,111 @@ func is_angle_in_interval(angle: float, direction_angle: ESCDirectionAngle) -> b
...
+		)
+	
+	if current_dir == target_dir:
+		return 0
+	
+	var internal = false
```
dploeger, line 482: What does that variable do? I don't understand its purpose.

dploeger (changes requested):
Awesome work and good algorithm! 👍🏽

addons/escoria-core/game/core-scripts/esc/commands/change_scene.gd
```diff
@@ -92,15 +92,18 @@ func run(command_params: Array) -> int:
...
+			
 			if script.events.has("setup"):
 				escoria.event_manager.queue_event(script.events["setup"])
 				var rc = yield(escoria.event_manager, "event_finished")
 				while rc[1] != "setup":
 					rc = yield(escoria.event_manager, "event_finished")
 				if rc[0] != ESCExecution.RC_OK:
 					return rc[0]
-
+
```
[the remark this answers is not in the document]
StraToN: That is because the :ready events are started before the transition is finished, so the player misses the beginning of the actions because they are hidden behind the transition.

Starting the transition (and waiting for it to finish) before :ready starts avoids the problem.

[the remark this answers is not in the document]
dploeger: Maybe we should add examples to the [event documentation](https://github.com/godotengine/escoria/blob/main/docs/esc.md#events) about :setup and :ready. Based on your description I don't get why ready should only be called the first time the player visits a scene.

mjtorn:
Although I'm effectively off Escoria for business reasons, I've been skimming
the emails occasionally.

I think I'm responsible for the `:setup` and `:ready` distinction, and the
intent wasn't to have anything trigger only for the first time. A global
variable can be used for that.

The big issue was how `telon` kept going so fast, that moving NPCs from
their *original initial* position to where they're *now* supposed to be was
done *after* the room had become visible.

So I added options to do things like keep the room black (eg. lights
being out) and such.

Of course as much as I still love Godot + Escoria, and appreciate your
work in it from a distance, you're free to redefine the semantics in
the rewrite. To me it made sense that you have a BTS `:setup` and then
action when `:ready`.

dploeger:
Thanks for your contribution here, @mjtorn. That definition is much more coherent than the one I got from the previous code! I was actually only guessing there 😀 

@StraToN based on this, do you want to remove the code that stores "readied" scenes and checks in this pr as well and correct the esc documentation part?

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
StraToN: Indeed, this test has no use. Same for the else block after it. I'll fix this.

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
StraToN: If internal=true, then the angle rotation finishes in the [0;7] interval and does not need to pass down 0 (back to 7) or upper than 7 (back to 0).

StraToN:
@mjtorn Thanks for the precision here! That's very helpful.
 
@dploeger 

> @StraToN based on this, do you want to remove the code that stores "readied" scenes and checks in this pr as well and correct the esc documentation part?

I'd rather do this in another dedicated PR, to avoid unclear commits.

addons/escoria-core/game/core-scripts/esc/commands/change_scene.gd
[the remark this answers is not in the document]
dploeger: Based on the conversation in the comments, this can be resolved now.

addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
dploeger: Ah okay. That makes sense then. Thanks

dploeger (approved):

PR #358, opened 2021-08-07
:setup action is not managed only once anymore

StraToN:
After this comment from @mjtorn clarifying the difference between :setup and :ready actions, a change has to be done on these actions.

As per this comment, it appears that :setup is not meant to be run once while :ready is meant to be run each time the scene is loaded. Indeed, this can be easily done with a global variable anyway.

~I was unable to find a clear use case for :setup event so this draft PR deletes it so I'd like to discuss this here first.~

Would it be correct to consider :setup and :ready as "steps" of one event happening when the scene is loaded? If this is correct, what is the exact separation of both steps?

----

Also adds various fixes (worker animations missing, etc).

Still missing: documentation update


addons/escoria-core/game/escoria.gd
```diff
@@ -315,7 +315,7 @@ func _ev_left_click_on_item(obj, event, default_action = false):
 			"arrived"
 		)
 		
-		self.logger.info("Context arrived: %s" % var2str(context))
+		self.logger.info("Context arrived: %s" % context)
```
dploeger, line 318: Ah, does context return a readable representation when called in a string context? Nice.

game/rooms/room03/esc/room03_bridge.esc
```diff
@@ -1,6 +1,13 @@
 # THIS ROOMS FEATURES AN OPEN BRIDGE WITH A ***BROKEN*** BUTTON TO USE IN ORDER TO OPEN
 
-:setup
+
```
dploeger, line 3: Hm. Isn't this part of the preparation for the scene? Doesn't that make the bridge flip up or down visibly to the player? (Applicable to all other changed esc files as well)

dploeger (changes requested):

addons/escoria-core/game/escoria.gd
StraToN: In fact, this was causing a crash due to an infinite loop, related to `_get_property_list()`. Not sure why. But yes, `Dictionary` is readable directly when printed.

game/rooms/room03/esc/room03_bridge.esc
StraToN: Indeed this is not correct. Fixing.

dploeger (approved):

StraToN:
Thanks!

PR #359, opened 2021-08-07
Removed hardcoded global_id "player"

### LINKED ISSUE #303: Fix hardcoded global id "player" throughout the framework
In an ESCScene, the player node (although referenced in ESCScene) has to have the global id "Player", which is counterintuitive and makes it problematic to use multiple switchable players.

I saw several other locations in the code, where this might be the case.

We should generalize it. For example, just refer to the player reference of ESCScene. Switching players would then just be changing that reference through code.

StraToN:
Fixes #303 

Also:
Removed hardcoded node name "start_position" (to do so, added a "is_start_location" property to `ESCLocation`)

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_room.gd
+++ b/addons/escoria-core/game/core-scripts/esc_room.gd
@@ -69,14 +69,12 @@ func _ready():
 		)
 		game.get_node("camera").set_target(player)
 	
-	if has_node("player_start"):
-		escoria.object_manager.register_object(
-			ESCObject.new(
-				$player_start.name,
-				$player_start
-			),
-			true
-		)
+	for n in get_children():
+		if n is ESCLocation and n.is_start_location:
+			escoria.object_manager.register_object(
+				ESCObject.new(n.name, n),
+				true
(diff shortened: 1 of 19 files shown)


addons/escoria-core/game/core-scripts/esc_room.gd
```diff
@@ -69,14 +69,12 @@ func _ready():
...
 	
-	if has_node("player_start"):
-		escoria.object_manager.register_object(
-			ESCObject.new(
-				$player_start.name,
-				$player_start
-			),
-			true
-		)
+	for n in get_children():
```
dploeger, line 72: Oooh. Nice! This is great to reduce ESC command use. It doesn't work for scenes with different exits, but it still makes things easier. 👍🏽

dploeger (approved):
Great work! Thanks

addons/escoria-core/game/core-scripts/esc_room.gd
StraToN: Thanks :)

PR #369, opened 2021-08-12
fix: FIxed non-editor call to _set_animation_player

dploeger:
This fixes calls to the setter for animation_player when the player is running the game, which resulted in a crash.


StraToN (approved):

StraToN:
Thanks!

PR #371, opened 2021-08-23
Moved and split escoria._ev_left_click_on_item() function into ESCController

### LINKED ISSUE #346: Overhaul escoria#_ev_left_click_on_item method
As a new Escoria developer I'd like to quickly understand all structures in the Escoria project. The method `_ev_left_click_on_item` is currently responsible for all actions required when the player clicks on an object. This includes working out the action and tool to take, walking the player character to the interaction point and the carrying out the action.

This is way too complex and needs refactoring. Also, a new developer would not expect that workflow in a method called `_ev_left_click_on_item`, so we need to move that code out of there as well.

(breakout issue of #314)

StraToN:
Added a boolean in ESCWalkContext to held whether interact should happen after destination is reached.

Fixes #346


addons/escoria-core/game/core-scripts/esc_controller.gd
```diff
@@ -0,0 +1,244 @@
...
+	if player_global_pos == destination_position:
+		# Manage exits
+		if obj.node.is_exit and escoria.action_manager.current_action in ["", "walk"]:
+			escoria.action_manager.activate("exit_scene", obj)
+		else:
+			# Manage movements towards object before activating it
+			if escoria.action_manager.current_action in ["", "walk"]:
+				if destination_position != clicked_position \
+						and not escoria.inventory_manager.inventory_has(obj.global_id):
+					escoria.action_manager.activate("arrived", obj)
```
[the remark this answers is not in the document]
dploeger: It triggers the event `:action` on the target object, okay. But for what type of object is the destination_position not the clicked_position?

dploeger (changes requested):
*Very* well done and *very* good documented. Thanks! 🎁 

I only have smaller remarks.

addons/escoria-core/game/core-scripts/esc_controller.gd
```diff
@@ -0,0 +1,244 @@
...
+# another object.
+#
+# #### Parameters
+#
+# - moving_obj_id: global id of the object that needs to move
+# - destination: Position2D or String of the object the moving object has head
+# to
+# - is_fast: if true, the walk is performed at fast speed (defined in the moving
+# object.
+func perform_walk(moving_obj_id: String, destination, is_fast: bool = false):
```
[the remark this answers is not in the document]
StraToN: Ah, I had done that in the first place then reverted because I needed both the id AND the ESCObject. But after multiple changes, I didn't see that I could set this back again. Thanks!

addons/escoria-core/game/core-scripts/esc_controller.gd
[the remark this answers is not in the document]
StraToN: Indeed this position check has no use. I removed it and it did work well. That was probably an old stuff of the first rewrite.

addons/escoria-core/game/core-scripts/esc_controller.gd
```diff
@@ -0,0 +1,244 @@
...
+# another object.
+#
+# #### Parameters
+#
+# - moving_obj_id: global id of the object that needs to move
+# - destination: Position2D or String of the object the moving object has head
+# to
+# - is_fast: if true, the walk is performed at fast speed (defined in the moving
+# object.
+func perform_walk(moving_obj: ESCObject, destination, is_fast: bool = false):
```
dploeger, line 19: You missed changing this to use Position2D or ESCObjects for the destination. Maybe that wasn't clear enough in my last review.

dploeger (changes requested):

addons/escoria-core/game/core-scripts/esc_controller.gd
StraToN: I think I missed it, sorry about this. Not sure how to achieve though: I can't use an ESCObject to hold either an ESCItem or a Vector2 because the node data has to be a Node, so I'd need to create a Position2D node just to hold the Vector2 data. I don't think it is worth...

dploeger: No, I meant support an Position2D or an ESCObject, but no string. This way you can just test for one of the two classes (either using `is` or `has_method` or something) and don't need to check the object in the object manager etc.

StraToN: Done

dploeger (approved):
Thanks a lot.

PR #384, opened 2021-09-04
Issue 335

dploeger:
This introduces a player being "selectable" and thus reacts like any ESCItem.

For testing, I implemented a look action on the player.

I also had to fix clearing the hover stack of ESCInputManager when switching scenes, because the player would be removed from that and a position would be null afterwards.


StraToN (approved):

StraToN:
Seems perfect. I initially thought we'd require an item action use on self to test this, but at the moment I don't see a reason why this would fail. Not urgent to add though, this can wait for another PR.

Thanks a lot!

PR #387, opened 2021-09-07
Small fixes and issue 386

### LINKED ISSUE #386: Looping idle animation causes player to never 'arrive'
In ESCMovable L288, we have started the idling animation and yield for it to finish before emitting the "arrived" signal.
We should not yield for the idle animation to finish.

dploeger:
This introduces the new "set_animations" command from my stream yesterday, fixes #386 and removes the @STUB keyword from already implemented ESC commands.


StraToN (approved):

StraToN:
Thanks!

PR #389, opened 2021-09-09
Fix when no speech key is used and translate speech key if used

StraToN:


dploeger (approved):
Thank you for cleaning up after me. 😆  🎁

StraToN:
Sorry that I missed that issue during the review :/ 
Thanks!

PR #391, opened 2021-09-10
Fixed escoria.current_state when dialog line finishes

StraToN:
This was causing ESC execution to hang after a say command because the state was at DIALOG and was not changed after the line is said.

- Also fixed connect() errors concerning the player being ready'd 2 times in certain cases, so trying to connect signals twice. That is a lot of checks!


dploeger (approved):
Great! Thanks. 🎁

PR #398, opened 2021-09-23
Changed dialog cursor

StraToN:
+ some modifications in android exports presets
+ commented out jump command in worker.esc as not implemented


dploeger (approved):
Thanks! 🎁

PR #403, opened 2021-09-26
feat: Use wait parameter instead of immediate for turn_to and set_angle

### LINKED ISSUE #402: Make the time between two angles in set_angle configurable
Currently, when using `set_angle` with "immediate" set to false, the `set_angle` method of ESCMovable waits one animation play until it turns around more.

This means one full walkcycle or one full idle animation for each angle, which isn't really practical.

It should better have a configured wait time per angle with a 0.1 second default, so turning around is actually visible.

dploeger:
Fixes #402


StraToN (approved):

StraToN:
Thanks!

PR #404, opened 2021-09-26
fix: Fixes for when the player scene isn't used in a room.

### LINKED ISSUE #393: Test ESCRooms with no "player_scene" parameter scene defined
It is expected that if no ESCPlayer is defined as parameter of ESCRoom, then any action requested by the player should apply immediately without waiting for the ESCPlayer to "arrive" at clicked position or ESCItem's interaction position.

This is supposed to allow Myst-like rooms: gamedev defines ESCItems as interactive, clickable hotspots that will trigger actions immediately, if and only if no `ESCRoom.player_scene` parameter is defined.

Currently, `ESCController.perform_inputevent_on_object()` method makes the player scene (using its reference from `escoria.main.current_scene.player`) walk to the determined position. If no ESCPlayer scene is defined, `escoria.main.current_scene.player` should be `null`, leading to null exception.

dploeger:
Fixes #393

### DIFF
--- a/game/rooms/room13/esc/hint.esc
+++ b/game/rooms/room13/esc/hint.esc
@@ -1,0 +1,2 @@
+:look
+	say player "I can just click the exits apparently."
--- a/addons/escoria-core/game/core-scripts/esc_controller.gd
+++ b/addons/escoria-core/game/core-scripts/esc_controller.gd
@@ -105,112 +105,111 @@ func perform_inputevent_on_object(
 	# (because object is inactive for example)
 	var dont_interact = false
 	
-	var destination_position: Vector2 = escoria.main.current_scene.player.\
-			global_position
-	
-	# If clicked object not in inventory, player walks towards it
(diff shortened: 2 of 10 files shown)


game/rooms/room13/esc/hint.esc
```diff
@@ -0,0 +1,2 @@
+:look
+	say player "I can just click the exits apparently."
```
StraToN, line 2: This tabulation will probably make the command not to be run. Remove the tab please ;)

StraToN (changes requested):

game/rooms/room13/esc/hint.esc
dploeger: No. That's actually not a problem. The regexp for commands support arbitrary whitespace at the start of a command.

StraToN (approved):

StraToN:
Thanks a lot!

PR #405, opened 2021-09-26
feat: This enables user-defined transition and uses a tween to animate the transitions

### LINKED ISSUE #344: Make transition shader material configurable
As a game developer using Escoria I'd like to configure how the scene transition is played. I may even want to change that dynamically.

The transition is using a shader in the color rect of transition.tscn. This shader is using a grayscale gradient which is configured as a texture in the shader params. This texture should be configurable in the project settings and changeable using code.

dploeger:
fixes #344

This can maybe clash with #404 as I was changing the ESC script of room12 there. I've completely changed room 12 in this, so THIS is the real deal. Maybe merge #404 first.


StraToN (approved):

StraToN:
This is great! That should ease transitions for users. Thanks!!

PR #406, opened 2021-09-30
Enhancement: load game scene once to allow run scene directly

StraToN:


StraToN:
Almost done.
- Crash on focusing item in the inventory after exiting room

addons/escoria-core/game/core-scripts/esc_room.gd
```diff
@@ -78,7 +87,24 @@ func _ready():
 	
 	if global_id.empty():
 		global_id = name
-
+		
+	if run_events and esc_script:
```
dploeger, line 91: Why is this called here and in change_room as well?

addons/escoria-core/game/core-scripts/esc_room.gd
```diff
@@ -123,3 +149,17 @@ func set_camera_limits(p_camera_limits: Array) -> void:
...
+
+# Runs the script event from the script attached, if any
+#
+# #### Parameters
+#
+# - event_name: the name of the event to run
+func run_script_event(event_name: String):
+	if !esc_script:
+		return
+	var script = escoria.esc_compiler.load_esc_file(esc_script)
```
dploeger, line 162: This should be stored in an instance variable so that the script isn't recompiled with every call to run_script_event. Only if that instance variable is empty, the script should be compiled.

addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
```diff
@@ -99,6 +99,9 @@ func unregister_object(object: ESCObject) -> void:
 	if not escoria.inventory_manager.inventory_has(object.global_id) \
 			and not object.global_id in RESERVED_OBJECTS:
 		objects.erase(object.global_id)
+	else:
```
dploeger, line 102: Probably doesn't really matter, but maybe add the "not object.global_id in RESERVED_OBJECTS" as an if clause here, so that *only* inventory objects get re-instanced.

addons/escoria-core/game/core-scripts/esc_room.gd
```diff
@@ -55,7 +55,16 @@ func _ready():
 	if Engine.is_editor_hint():
 		return
 	
+	var run_events = false
 	game = $game
+	if game == null:
+		game = escoria.game_scene
```
dploeger, line 61: Why is this logic here and in the change_scene command as well?

dploeger (changes requested):

addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
StraToN: Good point!

addons/escoria-core/game/core-scripts/esc_room.gd
StraToN: The `run_events` variable determines whether the events are run here or not. The variable is true if the game scene was not loaded. If it was already loaded, that means the scene is loaded from change_scene command.

The idea was to differentiate the events execution when the room is started from change_scene or where the room is directly run.

addons/escoria-core/game/main_scene.gd
```diff
@@ -5,10 +5,10 @@ extends Node
 
 # Start the main menu
 func _ready():
-	var main_menu_path = ProjectSettings.get_setting(
-		"escoria/ui/main_menu_scene"
-	)
-	var main_menu_scene = load(main_menu_path).instance()
-	escoria.call_deferred("add_child", main_menu_scene)
-	escoria.main_menu_instance = main_menu_scene
+	if escoria.main_menu_instance == null:
```
dploeger, line 8: Why is this not in escoria#_ready?

dploeger (changes requested):
Only a small remark

addons/escoria-core/game/main_scene.gd
StraToN: Because the main menu needs to be instanced AFTER settings have been loaded so that the language is correct.

dploeger (approved):
Awesome! That really helps a lot. Thanks. 🎁

PR #408, opened 2021-10-01
Move settings load and main menu loading in _init method

StraToN:
I managed to put all scenes loadings in escoria.gd into _init() method. 
I also added a few checks so that a message is displayed when some mandatory Escoria Settings are not set.


dploeger (approved):
Great! 🎁

PR #418, opened 2021-10-21
Fix a wrong validation check in camera_set_limits

StraToN:
Fixes godot-escoria/escoria-demo-game#1129
Also fix a type to use the enum, so that Inspector shows the string value of the selection for ESCRoom 's CameraLimits

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -85,4 +85,7 @@ func _on_event_finished(return_code: int, event: ESCEvent) -> void:
 func interrupt_running_event():
 	if _running_event == null:
 		return
+	for event in events_queue:
+		event.interrupt()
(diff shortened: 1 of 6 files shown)


addons/escoria-core/game/core-scripts/esc_room.gd
```diff
@@ -26,7 +26,7 @@ export(PackedScene) var player_scene
 export(Array, Rect2) var camera_limits: Array = [Rect2()] setget set_camera_limits
 
 # The editor debug display mode
-export(int) var editor_debug_mode = EditorRoomDebugDisplay.NONE setget set_editor_debug_mode
+export(EditorRoomDebugDisplay) var editor_debug_mode = EditorRoomDebugDisplay.NONE setget set_editor_debug_mode
```
dploeger, line 29: Does that actually work? I remember using an enum in an export and it didn't work, but I can't remember what the cause was.

PR #426, opened 2021-10-30
docs: Reorganized docs to the docs repo

dploeger:
This removes all the old documentation from the repo and instead uses the workflow to update the API docs in the docs repo.

Also see the upcoming docs PR about the missing information that were in the docs here. (namely the history and authors)


dploeger:
See godot-escoria/escoria-docs#21

StraToN (approved):

StraToN:
Thanks a lot!!

PR #430, opened 2021-11-10
Fix: Make float dialog follow character

StraToN:
Now when a say() is trigged, the floating dialog player will follow the character it's attached to, positioning the dialog box at its "dialog_position" node position.

Can be tested in room10 using "Slide" button

### DIFF
--- a/addons/escoria-dialog-simple/types/floating.gd
+++ b/addons/escoria-dialog-simple/types/floating.gd
@@ -112,5 +112,4 @@ func _on_dialog_finished():
 	# Make the speaking item animation stop talking, if it is still alive
 	if is_instance_valid(_current_character) and _current_character != null:
 		_current_character.stop_talking()
-	set_process(false)
 	emit_signal("say_finished")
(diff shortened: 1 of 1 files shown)


addons/escoria-dialog-simple/types/floating.gd
```diff
@@ -102,4 +112,5 @@ func _on_dialog_finished():
 	# Make the speaking item animation stop talking, if it is still alive
 	if is_instance_valid(_current_character) and _current_character != null:
 		_current_character.stop_talking()
+	set_process(false)
```
dploeger, line 115: This isn't really required I think, because the dialog manager removes the node from the scene tree anyway.

See https://github.com/godot-escoria/escoria-demo-game/blob/develop/addons/escoria-dialog-simple/esc_dialog_simple.gd#L55

dploeger (changes requested):

addons/escoria-dialog-simple/types/floating.gd
StraToN: Done

dploeger (approved):
Great! Thanks. 🎁

PR #433, opened 2021-11-11
Makes ESCDialog and ESCDialogOption instantiable without an ESC string

dploeger:
… string and other dialog optimizations


StraToN (approved):

StraToN:
Thanks a lot!

PR #434, opened 2021-11-12
feat: Optimized ESCCamera

dploeger:
This optimizes the ESCCamera class in various aspects.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
@@ -39,6 +39,6 @@ func run(command_params: Array) -> int:
 		.push(
 			escoria.object_manager.get_object(command_params[0]).node,
 			command_params[1],
-			command_params[2]
+			Tween.new().get("TRANS_%s" % command_params[2])
 		)
 	return ESCExecution.RC_OK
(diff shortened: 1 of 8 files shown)


addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
```diff
@@ -39,6 +39,6 @@ func run(command_params: Array) -> int:
 		.push(
 			escoria.object_manager.get_object(command_params[0]).node,
 			command_params[1],
-			command_params[2]
+			Tween.new().get("TRANS_%s" % command_params[2])
```
StraToN, line 42: If I understand correctly, the idea here is to get the right enum value from Tween, using command_params[2]. Does this really require creating a Tween object?

StraToN (changes requested):

addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
dploeger: I used that code like it was before. Is there another way to get the Enum value from a string?

StraToN: I thought there was... After a search, it seems there is not. Ok for me then.

StraToN (approved):

StraToN:
Thanks! 🥳

PR #438, opened 2021-11-13
Added a check on unknown ESC command at compilation time

StraToN:
If an unknown command is detected on compile ESC, fail.
This was not triggering any error in ESCRoom because ESCRoom looks for `:setup` and `:ready` events only. Any other element would not be considered.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_compiler.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_compiler.gd
@@ -219,8 +219,9 @@ func _compile(lines: Array) -> Array:
 			returned.append(dialog_option)
 		elif command_regex.search(line):
 			var command = ESCCommand.new(line)
-			escoria.logger.trace("Line is the command %s" % command.name)
-			returned.append(command)
+			if command.is_valid():
+				escoria.logger.trace("Line is the command %s" % command.name)
+				returned.append(command)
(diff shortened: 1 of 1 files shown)


addons/escoria-core/game/core-scripts/esc/esc_compiler.gd
```diff
@@ -219,8 +219,9 @@ func _compile(lines: Array) -> Array:
 			returned.append(dialog_option)
 		elif command_regex.search(line):
 			var command = ESCCommand.new(line)
-			escoria.logger.trace("Line is the command %s" % command.name)
-			returned.append(command)
+			if command.is_valid():
```
dploeger, line 222: I'm not 100% sure, that this works. `is_valid` runs the conditions of the command that check the game state. Are you sure, that compile time is the right time for this to happen?

dploeger (changes requested):

addons/escoria-core/game/core-scripts/esc/esc_compiler.gd
dploeger: Okay. As discussed on Discord, this is working indeed.

dploeger (approved):
Great. Thanks. 🎁

StraToN:
Thanks!

PR #443, opened 2021-11-15
fix: Resets the current verb after the action has finished

dploeger:
Simplemouse only sets the verb if it is changed and thus it is set to nothing when an action finished and resets the verb


StraToN (approved):

StraToN:
Very nice fix!

PR #456, opened 2021-11-22
Check that ending animation is not looping before resetting it

StraToN:
Fixes godot-escoria/escoria-issues#60


StraToN:
Thanks!

PR #464, opened 2021-11-25
Reverted setting ESC_LAST_SCENE too early in ESCRoom

StraToN:
Recent change in ESC_LAST_SCENE so that it was set too early resulted in this variable containing the current room id after the room was loaded, instead of the last one.

This "incorrect" change was necessary though for the transitions management between change_scene and ESCRoom.
As a side issue, I had to make ESCRoom manage unpause if in certain cases (happening when accessing room14 using room selector: in this case ESC_LAST_SCENE is null ; using 4th button for the first time in this case will have to deal with this).

Problem was that, in this case, change_scene does not play automatic transition, so the ESCRoom does, and it will now have to hide main/pause menus.


dploeger (approved):
Great! Thanks!

PR #465, opened 2021-11-26
fix: Fixes smaller things

dploeger:
Fix player sprite not showing correct idle animation after animations change in room switch
Remove some more error outputs (godot-escoria/escoria-issues#64)
Fix verbs menu button selection when input mode is not all


StraToN:
Looks good to me. Thanks!

PR #471, opened 2021-11-29
Fix a bug where player angle was wrongly tested around 360

StraToN:


StraToN:
@dploeger Thank you so much, you just nailed it!

PR #474, opened 2021-11-30
feat: add esc_current_scene reserved global

BHSDuncan:
New reserved global to hold current scene global ID to allow for flexibility in ESC scripts.
Also refactor reserved globals' handling for rooms with addition of ESCRoomManager and the elimination of string literals across multiple files. Allows for less chance of accidental errors as opposed to using string literals everywhere. As well, this encapsulates room-specific reserved globals and will serve as the basis for more room-specific behaviour management.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
@@ -13,6 +13,7 @@ signal global_changed(global, old_value, new_value)
 export(Dictionary) var _globals = {}
 
 
+# Registry of globals that are to be reserved for internal use only.
 var _reserved_globals: Dictionary = {}
 
 
@@ -42,7 +43,9 @@ func register_reserved_global(key: String, value = null) -> void:
 		)
 	_reserved_globals[key] = value
 	_globals[key] = value
-	emit_signal("global_changed", key, _globals[key], value)
+	
+	if value != null:
+		emit_signal("global_changed", key, _globals[key], value)
 	
 
 # Get the current value of a global
--- a/addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
@@ -5,30 +5,15 @@ extends Resource
 class_name ESCGlobalsManager
 
 
-const GLOBAL_ANIMATION_RESOURCES = "ANIMATION_RESOURCES"
-
-
 # Emitted when a global is changed
 signal global_changed(global, old_value, new_value)
-
-
-# A list of reserved globals which can not be overridden
-const RESERVED_GLOBALS = [
-	"ESC_LAST_SCENE",		# Contains the global_id of previous room
-	"FORCE_LAST_SCENE_NULL",	# If true, ESC_LAST_SCENE is not considered for 
-							# automatic transitions
-	"ANIMATION_RESOURCES"
-]
 
 
 # The globals registry
 export(Dictionary) var _globals = {}
 
 
-
-func _init():
-	set_global("ESC_LAST_SCENE", "", true)
-	set_global("FORCE_LAST_SCENE_NULL", false, true)
+var _reserved_globals: Dictionary = {}
 
 
 # Check if a global was registered
@@ -40,6 +25,25 @@ func _init():
 func has(key: String) -> bool:
 	return _globals.has(key)
 
+
+# Registers a global as being reserved and initializes it.
+#
+# #### Parameters
+#
+# - key: The key of the global to register
+# - value: The initial value (optional)
+func register_reserved_global(key: String, value = null) -> void:
+	if key in _reserved_globals:
+		escoria.logger.report_errors(
+			"ESCGlobalsManager.register_reserved_global: Can not override reserved global",
+			[
+				"Global key %s is already registered as reserved" % key
+			]
+		)
+	_reserved_globals[key] = value
+	_globals[key] = value
+	emit_signal("global_changed", key, _globals[key], value)
+	
 
 # Get the current value of a global
 #
@@ -77,7 +81,7 @@ func filter(pattern: String) -> Dictionary:
 # - key: The key of the global to modify
 # - value: The new value
 func set_global(key: String, value, ignore_reserved: bool = false) -> void:
-	if key in RESERVED_GLOBALS and not ignore_reserved:
+	if key in _reserved_globals and not ignore_reserved:
 		escoria.logger.report_errors(
 			"ESCGlobalsManager.set_global: Can not override reserved global",
 			[
--- a/addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
@@ -1,0 +1,32 @@
+extends Object
+class_name ESCRoomManager
+
+
+# Reserved globals which can not be overridden; prefixed with "GLOBAL_"
+#
+# Contains the global_id of previous room
+const GLOBAL_LAST_SCENE = "ESC_LAST_SCENE"
+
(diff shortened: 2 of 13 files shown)


addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
```diff
@@ -5,30 +5,15 @@ extends Resource
...
-
 # The globals registry
 export(Dictionary) var _globals = {}
 
 
-
-func _init():
-	set_global("ESC_LAST_SCENE", "", true)
-	set_global("FORCE_LAST_SCENE_NULL", false, true)
+var _reserved_globals: Dictionary = {}
```
StraToN, line 16: Please add a comment describing the attribute.

addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
```diff
@@ -0,0 +1,32 @@
+extends Object
+class_name ESCRoomManager
```
StraToN, line 2: I am not so fond of the name of this class: ESCRoomManager seems to imply this class manages rooms, but this is not true as it's just an enum entries holder. Would `ESCReservedGlobals` fit instead?

addons/escoria-core/game/escoria.gd
```diff
@@ -124,6 +128,7 @@ func _init():
 func _ready():
 	settings = save_manager.load_settings()
 	apply_settings(settings)
+	room_manager.register_reserved_globals()
```
StraToN, line 131: This could be done in room_manager's `_init()` function (to create) since it is being done right after it was instanced.

StraToN (changes requested):

addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
```diff
@@ -41,6 +26,25 @@ func has(key: String) -> bool:
...
+	if key in _reserved_globals:
+		escoria.logger.report_errors(
+			"ESCGlobalsManager.register_reserved_global: Can not override reserved global",
+			[
+				"Global key %s is already registered as reserved" % key
+			]
+		)
+	_reserved_globals[key] = value
+	_globals[key] = value
+	emit_signal("global_changed", key, _globals[key], value)
```
dploeger, line 45: I'd say, you should only emit the signal, if the value is not null.

addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
dploeger: Actually, my idea was that this manager would also change the room for example. Also, why's the GLOBAL_ANIMATION_RESOURCES here? That doesn't have to do anything with rooms if I'm not mistaken.

dploeger: Ah, I see. You saw it in ESCRoom and put it here. It's actually the global cache of the current animation resource of the player, so it doesn't really fit here. Hm... maybe ESCPlayer? I'm not 100% sure.

addons/escoria-core/ui_library/tools/room_select/room_select.gd
```diff
@@ -54,7 +54,9 @@ func _on_button_pressed():
 	# automatic transitions. 
 	# If FORCE_LAST_SCENE_NULL is True when change_scene starts:
 	# - ESC_LAST_SCENE is set to empty
-	escoria.globals_manager.set_global("FORCE_LAST_SCENE_NULL", true, true)
+	escoria.globals_manager.set_global( \
```
dploeger, line 57: I'd better format that like

```python
escoria.globals_manager.set_global(
  escoria.room_manager.GLOBAL_FORCE_LAST_SCENE_NULL,
  true,
  true
)
```

dploeger (changes requested):
Good first PR. :) Some remarks though.

addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
StraToN: > Actually, my idea was that this manager would also change the room for example

Can you give more precisions on this?

addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
BHSDuncan: Will do.

addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
BHSDuncan: > 

Yeah, you got it. It's also only ever used in ESCRoom, it seems. But I'm happy to put it wherever you deem best.

BHSDuncan: > Ah, I see. You saw it in ESCRoom and put it here. It's actually the global cache of the current animation resource of the > player, so it doesn't really fit here. Hm... maybe ESCPlayer? I'm not 100% sure.

Yeah, you got it. It's also only ever used in ESCRoom, it seems. But I'm happy to put it wherever you deem best.

addons/escoria-core/game/escoria.gd
BHSDuncan: Are you sure? `ESCRoomManager` depends on/makes use of `ESCGlobalsManager`, which would mean we'd have to instantiate the two in order in the `_init` method which could produce issues if, for some reason, someone decides to re-order/refactor the method. Not to mention that `input_manager.register_core()` is also called in `ready`.

Let me know.

addons/escoria-core/ui_library/tools/room_select/room_select.gd
BHSDuncan: Will do.

addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
dploeger: Isn't it also used in the set_animations command? 🤔

BHSDuncan: Actually, good catch. I didn't see it because it's actually referenced by a one-off constant in `globals_manager`, so I'll need to fix that anyway.

But like I said: I'll put the definition in whichever class you prefer!

addons/escoria-core/game/escoria.gd
StraToN: Actually this would just be what you're doing here: 
L113 instanciates `room_manager`, and in `escoria._ready()` the function `room_manager.register_reserved_globals()` is called, so this happens every time Escoria is run. Since `register_reserved_globals()` is called nowhere else, I assume that can automated a bit.

StraToN: Ok I re-read your answer and your justification is satisfying :) Thanks!

addons/escoria-core/ui_library/tools/room_select/room_select.gd
dploeger: Ah, you don't need the \ in the first line. Also, please put the closing bracket into the next line all to the left aligned with the command itself.

dploeger (changes requested):
Ah, I guess you still need to fill ESCRoomManager with the business logic of change scene and everything that is not related to just one room from ESCRoom and ESCChangeSceneCommand

dploeger (approved):
Awesome. Thanks. I'll approve this like we talked on Discord. Need to wait for Julian's comment, though.

dploeger:
Oh, I forgot: Great work. 🎁  :)

BHSDuncan:
> Oh, I forgot: Great work. 🎁 :)

Thank you! Hopefully I can always produce good work!

StraToN (approved):

StraToN:
All good for me. Good remarks and great addition, great work! Thanks a lot!

PR #475, opened 2021-12-01
feat: Simplified checking for angles

dploeger:


StraToN:
I cannot tell why I couldn't come up with something simpler as you did. Thanks!

PR #478, opened 2021-12-06
fix: Support Godot 3.4

dploeger:


StraToN (approved):

StraToN:
To give a few more context about this fix, the "issue" is that in Godot <3.4, it was possible to define a specific Resource type for an exported variable, allowing the user to load the right Resource file directly instead of setting the global "Resource" type that allows using any resource (ie. inheriting Resource), including non-related or stock Godot Resources, not just Escoria's.

From Godot 3.4, this is not possible anymore.
We opened this bug: https://github.com/godotengine/godot/issues/51219
The answer was that this hack was based on a bug in Godot, which was fixed and thus invalidated the hack.

Long story short: the Escoria resources to use here are explicitly indicated in Escoria documentation. Dear users, please ensure you are setting the right resources for your animations.

StraToN:
Thanks!

PR #481, opened 2021-12-13
Fix input being blocked when item is not interactive #91

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/91


dploeger (approved):
Great. Thanks! 🎁

PR #483, opened 2021-12-31
Fix initial value of globals in signals emissions

StraToN:
Fixes a crash when setting globals for the first time.


dploeger (approved):
Great! 🎁

StraToN:
Thanks!

PR #485, opened 2022-01-17
Fix transitions not happening if the ESCRoom has no ESC script

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/95


dploeger (approved):
Great! Thanks.

PR #486, opened 2022-01-18
Set a default implementation for ESCGame._on_event_done()

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/97


dploeger (approved):
Great. Thanks! 🎁

PR #488, opened 2022-02-04
fix: set a proper guard in case of no speaking animations

BHSDuncan:
… the upper limit check for the speaking index


dploeger (approved):
Great! Thanks. 🎁

StraToN (approved):

StraToN:
Oh indeed, that was all wrong! Thanks a lot!

PR #489, opened 2022-02-11
Fix: use Escoria plugin version from plugin.cfg in savegames

StraToN:
Save manager used escoria.ESCORIA_VERSION constant value, which is now incorrect. We now use plugin.cfg's version instead.


balloonpopper:
I can confirm this fixes the bug with the infinite loop, however I did find a edge case that isn't handled. I manually deleted save games 1-3 but still had save game 4 in the folder.
This results in an error on line 38 of load_game.gd
Invalid get index '1' (on base: 'Dictionary').

StraToN:
Good point. I'll add a check to manage this case.

.gitignore
```diff
@@ -1,5 +1,5 @@
 bin
 *.import
 .import
-*.translation
-.godot
\ No newline at end of file
+game/translations/*.translation
```
dploeger, line 4: What's the point of this in the context of this PR? If it doesn't make sense, please remove the change.

dploeger (changes requested):
Just a very small remark.

BHSDuncan (approved):
I'm good with things and agree w Dennis.

.gitignore
StraToN: Done

BHSDuncan (approved):

PR #495, opened 2022-02-15
feat: Debug set state animations

balloonpopper:
Print a debug message when a set state command is run to indicate whether a matching animation was found or not.


dploeger (approved):
Great! Thanks. 🎁

StraToN (approved):

PR #497, opened 2022-02-16
Fix crash when save folder is empty and manage corrupted savegame file

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/121
Fixes https://github.com/godot-escoria/escoria-issues/issues/122

PR #499, opened 2022-02-17
feat: Removed obsolete code. Fixes godot-escoria/escoria-issues#99

dploeger:
I haven't tested that yet, because develop is currently unstable. So I might check this again after #497 

But I don't see, why this shouldn't work as the code was bogus basically. 😄


BHSDuncan (approved):
Looks ok in principle. :)

StraToN (approved):

StraToN:
LGTM.

PR #500, opened 2022-02-19
fix: animations doesnt have a speaks parameter when it is null

balloonpopper:
fix: This is to fix #496. The current code doesn't work as "animations" doesn't have a "speaks" parameter when "animations" is null.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_item.gd
+++ b/addons/escoria-core/game/core-scripts/esc_item.gd
@@ -453,7 +453,7 @@ func turn_to(object: Node, wait: float = 0.0):
 # Play the talking animation
 func start_talking():
 	# Only start the speaking animation if we actually have them setup
-	if animations.speaks != null \
+	if animations != null \
 			and animations.speaks.size() > 0 \
 			and get_animation_player() \
 			and _movable.last_dir >= 0 \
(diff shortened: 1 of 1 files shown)


balloonpopper:
FYI @BHSDuncan @dploeger @StraToN  (as I can't tag you as reviewers)

BHSDuncan:
@balloonpopper says, "If you go to an ESCItem that hasn't had its "Animations" set and you go to create one, the resource it asks you to create is an ESCAnimationsResource, which contains an empty Speaks array."

So this check works for me.

dploeger:
> FYI @BHSDuncan @dploeger @StraToN  (as I can't tag you as reviewers)

This is done automatically now. 😊

dploeger:
Didn't @StraToN do the same change in one of the last prs? 🤔 Or was that with some other animation?

balloonpopper:
@dploeger : @BHSDuncan did a change to animations but it needed the above correction. Is that the change you're thinking of?

dploeger:
Ah maybe.

dploeger:
I was thinking if we'd need a check for `animations.speak!=null` as well, but if animations is set, we expect an animations resource there.

dploeger (approved):
Thanks! 🎁

BHSDuncan:
@dploeger can you merge this? I'm not authorized.

PR #502, opened 2022-02-19
fix: enable transition to ESCRoom even if esc_script is not set

bolinfest:
When creating a new `ESCRoom`, there was a bug where it was not possible
to transition into the room if the room's `esc_script` was not set.
Apparently, `_perform_script_events()` would bail out early when
`room.esc_script` is falsy, which seems to be why the transition does
not happen.

Perhaps this defensive check was necessary at one time, but that does
not appear to be the case today because the only code that reads
`room.esc_script` in this file is within the `_run_script_event()`
function, which does its own defensive check for `room.esc_script`. As
such, it appears that we can let the rest of `_perform_script_events()`
do its thing even when `room.esc_script` is falsy (i.e., `""`).

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
@@ -267,7 +267,7 @@ func init_room(room: ESCRoom) -> void:
 #
 # - room: The ESCRoom to be initialized for use.
 func _perform_script_events(room: ESCRoom):
-	if room.esc_script and escoria.event_manager.is_channel_free(escoria.event_manager.CHANNEL_FRONT) \
+	if escoria.event_manager.is_channel_free(escoria.event_manager.CHANNEL_FRONT) \
 			or (
 				not escoria.event_manager.is_channel_free(escoria.event_manager.CHANNEL_FRONT) and \
 				not escoria.event_manager.get_running_event(


StraToN:
This may be related to https://github.com/godot-escoria/escoria-issues/issues/95 (fixed)

StraToN:
This check was added by @BHSDuncan in https://github.com/godot-escoria/escoria-demo-game/commit/99dc1e01106ce651d1b78ae574cca9d0202c48b1 (#485). Maybe he can give us some details on the reason why the check was re-introduced?

BHSDuncan:
> This check was added by @BHSDuncan in [99dc1e0](https://github.com/godot-escoria/escoria-demo-game/commit/99dc1e01106ce651d1b78ae574cca9d0202c48b1) (#485). Maybe he can give us some details on the reason why the check was re-introduced?

I didn't actually reintroduce it: I factored it to a spot that (at the time) made more sense since where the places it was originally in could be reduced.

The ticket that commit was a part involved introducing no new functionality, and if code was moved around, it would've been for the purpose of maintaining the same functionality (no more, no less) but with what was supposed to be cleaner/more efficient code.

TL;DR: It wasn't really "re-introduced" per se; it was just factored, so if the check can safely be removed, then by all means.

dploeger (approved):
Fine by me. Thanks! 🎁

BHSDuncan (approved):

bolinfest:
We’re there any outstanding concerns with this one?

StraToN:
Thanks a lot !

PR #504, opened 2022-02-20
refactor: remove redundant check for camera node

balloonpopper:
@StraToN @dploeger @dploeger FYI 
I couldn't find any reason you needed to set camera_node in ESCItem to get push commands to work. As far as I can tell it's not necessary so I've removed it. If I'm wrong and it ends up that it's needed, the camera node setting in ESCItem (what it's for and what you set it to) needs to be documented.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_item.gd
+++ b/addons/escoria-core/game/core-scripts/esc_item.gd
@@ -489,7 +489,8 @@ func update_idle():
 # Return the camera position if a camera_position_node exists or the
 # global position of the player
 func get_camera_node():
-	if camera_node and get_node(camera_node):
+	if has_node(camera_node):
+		escoria.logger.debug("Camera node found - directing camera to the camera_node on " + global_id)
 		return get_node(camera_node)
 	return self
 


dploeger (approved):
Wayy cleaner. Thanks! 🎁

balloonpopper:
Can the camera_node variable be removed entirely from ESCItem then? I'm not sure what it was put there for.

dploeger:
> Can the camera_node variable be removed entirely from ESCItem then? I'm not sure what its purpose is - it looks deprecated to me, the only function that uses it is the one I've edited, but I'm not sure why it was originally put there.

No, it can be used together with `camera_set_target` if you don't want to use the global position of the item, but of a secondary node.

dploeger:
(at least that's the plan :D )

balloonpopper:
Happy to be overruled, but I'd suggest removing it and re-adding it as part of the proposed feature to keep the code/interface cleaner. If the camera_set_target change is planned for 4.0 though, maybe leave it and add an issue to track the work?

dploeger:
No sorry, I wasn't specific enough. It is already possible with the current code. It is not planned.

I was saying "that's the plan" because I hope, that that it's implemented currently. You might have a look.

balloonpopper:
As far as I can see it's not implemented, a search of the specific file returns no results.
```
/mnt/e/Godot Projects/escoria-demo-game$ grep -i camera_node ./addons/escoria-core/game/core-scripts/esc/commands/camera_set_target.gd
/mnt/e/Godot Projects/escoria-demo-game$
```

And a search of all files shows the same. The esc_item reference is the one we're updating here, the others don't point to the camera_node entity that is currently manually set.
```
/mnt/e/Godot Projects/escoria-demo-game$ grep -ir camera_node

addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd:#   has a child node called `camera_node`, its location will be used. If not,
addons/escoria-core/game/core-scripts/esc_item.gd:export(NodePath) var camera_node
addons/escoria-core/game/core-scripts/esc_item.gd:func get_camera_node():
addons/escoria-core/game/core-scripts/esc_item.gd:      if camera_node and get_node(camera_node):
addons/escoria-core/game/core-scripts/esc_item.gd:              return get_node(camera_node)
addons/escoria-core/game/scenes/camera_player/esc_camera.gd:    elif p_target.has_method("get_camera_node"):
addons/escoria-core/game/scenes/camera_player/esc_camera.gd:            if "global_position" in p_target.get_camera_nod
():
addons/escoria-core/game/scenes/camera_player/esc_camera.gd:                    _target = p_target.get_camera_node().global_position
addons/escoria-core/game/scenes/camera_player/esc_camera.gd:            if "zoom" in p_target.get_camera_node():
addons/escoria-core/game/scenes/camera_player/esc_camera.gd:                    _zoom_target = p_target.get_camera_node().zoom
```

dploeger:
No, it's actually used in esc_camera!

p_target is the target for the set_target command and it can be of various type: A location, a vector2d or an escitem. So it calls get_camera_node on the esc item and sets the target to the position of that node instead of the escitem itself.

Or am I misreading that?

balloonpopper:
How that reads to me is that it looks for a "get_camera_node" method - not the node_path that's specified in ESCitem. As ESCitem has the function, it resolves the function and gets back whatever the ESCitem's "camera_node" node is. It could be changed to return the nodepath of camera_node if one's set, but that's not how the function is currently written.
(Assuming I'm reading it correctly 😄 )

dploeger:
> How that reads to me is that it looks for a "get_camera_node" method - not the node_path that's specified in ESCitem. As ESCitem has the function, it resolves the function and gets back whatever the ESCitem's "camera_node" node is. It could be changed to return the nodepath of camera_node if one's set, but that's not how the function is currently written. (Assuming I'm reading it correctly 😄 )

Exactly, so the camera_node variable still is required (to set the path to the node which position should be used).

dploeger:
> Not as written in this pr though.
> 
> if camera_node and get_node(camera_node): if has_node(camera_node): return get_node(camera_node)
> 
> This will return the node called "camera_node", not the contents of a nodepath variable.

? But the function in esc_camera.gd requires the node and not the nodepath?

balloonpopper:
Let me play around with it and get back to you.

BHSDuncan (approved):

StraToN:
Thanks! :partying_face:

PR #505, opened 2022-02-22
feat: more useful error messages

balloonpopper:
The current error messages are not as developer friendly as they could be. This addresses that as per #108 
It does not however fix the other fault mentioned in that issue where the filename is not mentioned as part of the error message.

Before:
 2022-2-22T173259 (E)	Errors in file Argument type did not match descriptor for command "camera_shift"
Argument 1  is of type 4. Expected 2,3

After:
 2022-2-22T173259 (E)	Errors in file Argument type did not match descriptor for command "camera_shift"
Argument 1 ("-100") is of type TYPE_STRING. Expected [TYPE_INT|TYPE_REAL]

FYI @dploeger @StraToN @BHSDuncan

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/types/esc_command_argument_descriptor.gd
+++ b/addons/escoria-core/game/core-scripts/esc/types/esc_command_argument_descriptor.gd
@@ -2,6 +2,15 @@
 extends Object
 class_name ESCCommandArgumentDescriptor
 
+# As the get_type command was deprecated with Godot 2.x w we need a way to determine
+# variable types. Ideally these wouldn't be hardcoded but there's no GDScript 3.x command to
+# turn a type back to its name.
+const GODOT_TYPE_LIST = ["nil", "bool", "int", "real",  "string", \
+	"vector2", "rect2", "vector3",  "matrix32", "plane", "quat", \
+	"aabb", "matrix3",  "transform", "color", "image", "node_path", \
+	"rid", "object", "input_event", "dictionary", "array", \
+	"raw_array", "int_array", "real_array", "string_array", \
+	"vector2_array", "vector3_array", "color_array", "max"]
 
 # Number of arguments the command expects
 var min_args: int = 0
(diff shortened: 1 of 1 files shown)


dploeger (approved):
And there it is already" Thanks a lot! Very helpful. 🎁

addons/escoria-core/game/core-scripts/esc/types/esc_command_argument_descriptor.gd
```diff
@@ -2,6 +2,15 @@
 extends Object
 class_name ESCCommandArgumentDescriptor
 
+# As the get_type command was deprecated with Godot 2.x w we need a way to determine
+# variable types. Ideally these wouldn't be hardcoded but there's no GDScript 3.x command to
+# turn a type back to its name.
+var godot_type_list = ["TYPE_NIL", "TYPE_BOOL", "TYPE_INT", "TYPE_REAL",  "TYPE_STRING", \
```
BHSDuncan, line 8: Minor nitpick/comment/not married to it, but, unless this list can be added to dynamically, maybe this should be marked as `const`? I realize that arrays in GDScript are passed by reference anyway, but for the sake of intent (a la pseudo-private prefixes), maybe this makes some sense. Other than that, this PR looks good.

BHSDuncan (changes requested):

addons/escoria-core/game/core-scripts/esc/types/esc_command_argument_descriptor.gd
BHSDuncan: If you can CONSTANT_CASE the name here, I think we're good!

balloonpopper: Made a couple of changes here in addition to CONSTANT_CASE. I've replaced the "|" with " or ", removed "TYPE_" off all the types,  and lower-cased all the types to make it more readable. New error example:
Argument 0 ("-300") is of type string . Expected [ int or real ]

balloonpopper:
Made a couple of changes here in addition to CONSTANT_CASE. I've replaced the "|" with " or ", removed "TYPE_" off all the types, and lower-cased all the types to make it more readable. New error example:
```
Argument 0 ("-300") is of type string . Expected [ int or real ]
```

BHSDuncan (approved):

BHSDuncan:
Works for me! Thanks!

balloonpopper:
One more tiny update. I didn't like the additional space after the "type string ." before the "." - now fixed.
Argument 0 ("-300") is of type string. Expected [ int or real ]

Issue #126, opened 2018-04-18
Create inventory toggle button


fleskesvor:
The inventory should have a button which can be used to open and close the inventory, as mentioned by @StraToN in #123.

It should be possible to configure whether or not to toggle the inventory by clicking the button or having it open the inventory by hovering the mouse. The latter case can also be used to open the inventory with a button without a texture if a developer wants a mouse aware area at the bottom of the screen, which opens the inventory when the cursor enters this area. In that case, the inventory should probably also close when the cursor leaves, unless an item has been selected.

lavaduderDev:
What is this in hud.gd? lines 10-19
```
func inv_toggle():
	#get_node("inventory").toggle()
	pass


func _on_inv_toggle_vis_chaged():
	if (get_node("inv_toggle").is_hidden()):
		get_node("buttons").hide()
	else:
		get_node("buttons").show()
```
Where in the world is the node "buttons" coming from? It's not in the base scene. Was there a node for inventory called buttons at one point? It seems like unneeded code.

fleskesvor:
My guess is it's the remains of some of the same stuff that was removed in #70. It can most likely be safely removed.

StraToN:
This is part of UI design, which has to be developed by gamedev. Escoria will provide API calls to fill the inventory and send according signals when clicking on inventory items, but showing/hiding the inventory is not Escoria's task.

PR #506, opened 2022-02-22
documentation: Fixes to command descriptions

balloonpopper:
Migration of https://github.com/godot-escoria/escoria-docs/pull/44

Outstanding questions below please (@StraToN @dploeger @BHSDuncan ) - once I've got clarification on the below I can make some further updates.

I've taken a stab at updating the ESC language commands documentation as there were things I thought could be worded better. Among other things, this fixes #124 - camera page links (#126 - camera limit documentation still needs a bit more general limits discussion, but I thought a future PR by someone for that would be better).

Some other pages as listed below I'm happy to update but I need your input on please.

HideMenu Command : hide_menu menu_type [enable_automatic_transition]
What is the difference between automatic_transition = true and = false? I don't understand what "automatic_transition" does.

QueueEvent :
I don't understand what channel is for, and the block documentation doesn't make sense to me either.

QueueResource:
As a user, I don't know what a "resource cache" is, or what a "resource" is, so I can't update the documentation to help other new users unfortunately.

RangGlobal:
It says both that the max_value is inclusive and exclusive, I need to know which it is to correct the documentation.

Repeat :
I'd like to give an example of where to use this to make it more obvious for users when this would come in handy.
Presumably if the event has a series of commands and you put a repeat at the end, the series repeats, but how do you break out of the loop? Is it possible?
Or do you only use this for something like animating a guard who walks to the left, stops, walks to the right, stops etc? How would I get him to stop so I could talk to him?

SchedEvent:
This implies that two events can't run at the same time as it queues the event if another is already running. Does this only apply to scheduled events - i.e. you can't run 2 scheduled events at the same time (the second queues), but you can run a scheduled event while normal scene events are in progress?

SetAngle:
The wait time isn't clear. "Number of seconds to wait for each animation occurring between the
current angle of object and the angle specified. "
What are the animations it plays on each rotation? The idle animation?

SetAnimations:
What's an animation resource? What format is the "animations" parameter (and how do I create one of these? - is it just an AnimatedSprite?)

SetInteractive:
What exactly does it mean to be interactive? How does a non-interactive object differ from a sprite that I put in my scene?

SetState:
It says that it "Can be used to change the appearance of an item or player character. See https://docs.escoria-framework.org/states for details."
The states link says nothing about changing the appearance of an item/player. If someone can give me a brief overview of how that works I'll update the documentation.

ShowMenu :
same as HideMenu - no idea what automatic_transition does

Stop:
What's the scope? If it's inside an if block, is it only the if block that is affected or does it terminate the event?

StopSnd :
Says "Stops the given sound player's stream" - not "stops the specified sound/speed/music file that is currently playing".
Does this mean you can only run 1 x _sound file at a time, 1 x _speech file at a time and 1 x _music file at at time?
This limitation (if it exists) is not listed in PlaySnd.

Transition
Can a user provide a transition? Are there requirements for the transition they must follow if they want to write their own?
Is a transition "in" just playing the shader backwards? (Or is the "out" just playing the shader backwards?)

turn_to:
Which animation(s) does it use for each turn? Is it playing the "idle" animation (and will it just play the first frame or the whole idle animation?)

walk_block:
Does this move the ESCPlayer/ESCItem to the coordinate of target? If the target has a ESCLocation does it go to the target or the ESCLocation?

Walk:
Speed as above

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
@@ -1,6 +1,8 @@
 # `camera_push target [time] [type]`
 #
-# Pushes the camera to point at a specific `target`.
+# Pushes (moves) the camera so it points at a specific `target`. If the camera 
+# was following a target (like the player) previously, it will no longer follow
+# this target. 
 #
 # **Parameters**
 #
@@ -13,7 +15,7 @@
 # Supported transitions include the names of the values used
 # in the "TransitionType" enum of the "Tween" type (without the "TRANS_" prefix):
 #
-# https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations
+# See https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations
 #
 # For more details see: https://docs.escoria-framework.org/camera
 #
@@ -31,7 +33,7 @@ func configure() -> ESCCommandArgumentDescriptor:
 	)
 
 
-# Validate wether the given arguments match the command descriptor
+# Validate whether the given arguments match the command descriptor
 func validate(arguments: Array):
 	if not escoria.object_manager.objects.has(arguments[0]):
 		escoria.logger.report_errors(
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_set_pos.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_set_pos.gd
@@ -1,6 +1,6 @@
 # `camera_set_pos time x y`
 #
-# Moves the camera to the given position.
+# Moves the camera to the given absolute position over a time period.
 #
 # **Parameters**
 #
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_set_target.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_set_target.gd
@@ -1,10 +1,13 @@
 # `camera_set_target time object`
 #
-# Configures the camera to follow the specified target `object`
+# Configures the camera to follow the specified target `object` as it moves
+# around the current room. The transition to focus on the `object` will happen
+# over a time period.
 #
 # **Parameters**
 #
-# - *time*: Number of seconds the transition should take
+# - *time*: Number of seconds the transition should take to move the camera 
+#   to follow `object`
 # - *object*: Global ID of the target object
 #
 # For more details see: https://docs.escoria-framework.org/camera
@@ -23,7 +26,7 @@ func configure() -> ESCCommandArgumentDescriptor:
 	)
 
 
-# Validate wether the given arguments match the command descriptor
+# Validate whether the given arguments match the command descriptor
 func validate(arguments: Array):
 	if not escoria.object_manager.objects.has(arguments[1]):
 		escoria.logger.report_errors(
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom.gd
@@ -1,8 +1,10 @@
 # `camera_set_zoom magnitude [time]`
 #
-# Zooms the camera in/out to the desired `magnitude`. Values larger than 1 zoom
-# the camera out while smaller values zoom in, relative to the default value
-# of 1.
+# Zooms the camera in/out to the desired `magnitude`. Values larger than '1' zoom
+# the camera out while smaller values zoom in. These values are relative to the 
+# default zoom value of '1', not the current value. As such, while using a value 
+# of '0.5' would double the size of the graphics, running the same command again
+# would result in no change. The zoom will happen over the given time period.
 #
 # **Parameters**
 #
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_height.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_height.gd
@@ -1,6 +1,6 @@
 # `camera_set_zoom_height pixels [time]`
 #
-# Zooms the camera in/out so it occupies the given height in pixels
+# Zooms the camera in/out so it occupies the given height in pixels.
 #
 # **Parameters**
 #
@@ -24,7 +24,7 @@ func configure() -> ESCCommandArgumentDescriptor:
 	)
 
 
-# Validate wether the given arguments match the command descriptor
+# Validate whether the given arguments match the command descriptor
 func validate(arguments: Array):
 	if arguments[0] < 0:
 		escoria.logger.report_errors(
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_shift.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_shift.gd
@@ -1,6 +1,7 @@
 # `camera_shift x y [time] [type]`
 #
-# Shifts the camera by the given horizontal and vertical amounts.
+# Shifts the camera by the given horizontal and vertical amounts relative to the 
+# current location.
 #
 # **Parameters**
 #
--- a/addons/escoria-core/game/core-scripts/esc/commands/custom.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/custom.gd
@@ -1,13 +1,15 @@
 # `custom object node func_name [params]`
 #
-# Calls the given Godot function on a (child) node of a registered `ESCitem`.
+# 
+# Executes the specified Godot function. This function must be in a script
+# attached to a child node of a registered `ESCitem`.
 #
 # **Parameters**
 #
 # - *object*: Global ID of the target `ESCItem`
 # - *node*: Name of the child node of the target `ESCItem`
 # - *func_name*: Name of the function to be called
-# - *params*: Any primitive, non-array arguments for the function. Multiple
+# - params: Any arguments to be passed to the function (array and object parameters are not supported). Multiple
 #   parameters can be passed by using comma-separated values inside a string
 #
 # @ESC
@@ -24,7 +26,7 @@ func configure() -> ESCCommandArgumentDescriptor:
 	)
 
 
-# Validate wether the given arguments match the command descriptor
+# Validate whether the given arguments match the command descriptor
 func validate(arguments: Array):
 	if not escoria.object_manager.objects.has(arguments[0]):
 		escoria.logger.report_errors(
--- a/addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
@@ -2,10 +2,13 @@
 #
 # Hides either the main menu or the pause menu. The enable_automatic_transition
 # parameter can be used to specify if Escoria manages the graphical transition
-# or not. If set to false, you can perform your own transition as part of the
-# room's `setup` event (if you want to change the look of the transition used
-# from the default for example) using the `transition` command.
-#
+# for you or not. 
+# Setting `enable_automatic_transition` to false allows you to manage the
+# transition effect for your room as it transitions in and out. Place a 
+# `transition` command in the room's `setup` event to manage the look of the 
+# transition in, and in the room's `exit_scene` event to manage the look of the
+# transition out.
+# 
 # **Parameters**
 #
 # - *menu_type*: Which menu to hide. Can be either `main` or `pause` (default: `main`)
--- a/addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
@@ -1,10 +1,14 @@
 # `hide_menu menu_type [enable_automatic_transition]`
 #
-# Hides either the main menu or the pause menu.
+# Hides either the main menu or the pause menu. The enable_automatic_transition
+# parameter can be used to specify if Escoria manages the graphical transition
+# or not. If set to false, you can perform your own transition as part of the
+# room's `setup` event (if you want to change the look of the transition used
+# from the default for example) using the `transition` command.
 #
 # **Parameters**
 #
-# - *menu_type*: Type of menu to hide. Can be either `main` or `pause` (default: `main`)
+# - *menu_type*: Which menu to hide. Can be either `main` or `pause` (default: `main`)
 # - *enable_automatic_transition*: Whether to automatically transition from the menu (default: `false`)
 #
 # @ESC
@@ -21,7 +25,7 @@ func configure() -> ESCCommandArgumentDescriptor:
 	)
 
 
-# Validate wether the given arguments match the command descriptor
+# Validate whether the given arguments match the command descriptor
 func validate(arguments: Array):
 	if not arguments[0] in ["main", "pause"]:
 		escoria.logger.report_errors(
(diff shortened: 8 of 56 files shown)


addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
```diff
@@ -13,9 +15,9 @@
 # Supported transitions include the names of the values used
 # in the "TransitionType" enum of the "Tween" type (without the "TRANS_" prefix):
 #
-# https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations
+# See https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations
 #
-# For more details see: https://docs.escoria-framework.org/camera
+# For more details see: https://docs.escoria-framework.org/en/devel/getting_started/camera.html
```
balloonpopper, line 20: @dploeger (Comment from @BHSDuncan on original PR) : will this tie us into this particular link always using the devel branch of the docs? What if we eventually deploy from a different branch? Or am I way off?

There are a few references pointing to this development branch link across this PR that may need to be corrected based on your response.

addons/escoria-core/game/core-scripts/esc/commands/custom.gd
```diff
@@ -1,13 +1,15 @@
...
+# Executes the specified Godot function. This function must be in a script
+# attached to a child node of a registered `ESCitem`.
 #
 # **Parameters**
 #
 # - *object*: Global ID of the target `ESCItem`
 # - *node*: Name of the child node of the target `ESCItem`
 # - *func_name*: Name of the function to be called
-# - *params*: Any primitive, non-array arguments for the function. Multiple
+# - *params*: Any non-array arguments to be passed to the function. Multiple
```
balloonpopper, line 12: @dploeger : Comment from @BHSDuncan on this change - Do we allow for objects to be used as params @dploeger ? I thought we only allowed non-array primitive types?

addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
```diff
@@ -1,10 +1,14 @@
 # `hide_menu menu_type [enable_automatic_transition]`
 #
-# Hides either the main menu or the pause menu.
+# Hides either the main menu or the pause menu. The enable_automatic_transition
+# parameter can be used to specify if Escoria manages the graphical transition
+# or not. If set to false, you can perform your own transition as part of the
+# room's `setup` event (if you want to change the look of the transition used
+# from the default for example) using the `transition` command.
```
balloonpopper, line 7: @StraToN - Comment from @BHSDuncan on my earlier PR : "I actually think you do it as part of both when exiting a room and when entering a room, so not just setup but also the exiting event in the previous room. @StraToN those correspond to transitioning out and in, no?"

addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
dploeger: Yes, the way here is to go with redirects. /camera should be a redirect to the develop page currently, but apparently is missing.

That's a thing we can configure on the read the docs admin page.

Can you kindly filter out the revelant links and list here? I'll add the redirects then.

balloonpopper: @dploeger Here's all the links listed based off a grep in every .gd file

https://docs.escoria-framework.org/camera
https://docs.escoria-framework.org/en/devel/getting_started/camera.html
https://docs.escoria-framework.org/states for details.
https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-match)
https://docs.godotengine.org/en/stable/classes/class_tween.html#enumerations
https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations

addons/escoria-core/game/core-scripts/esc/commands/custom.gd
dploeger: Yes. Agreed. No objects.

addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
dploeger: - [ ] https://docs.escoria-framework.org/camera
- [ ] https://docs.escoria-framework.org/en/devel/getting_started/camera.html
- [ ] https://docs.escoria-framework.org/states
- [ ] https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-match
- [ ] https://docs.godotengine.org/en/stable/classes/class_tween.html#enumerations
- [ ] https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations)

dploeger: Okay, that's not much. You can use https://docs.escoria-framework.org/camera for https://docs.escoria-framework.org/en/devel/getting_started/camera.html.

I just added the redirect.

The godot docs are outside of our jurisdiction, so I'd just keep the pages under docs.godotengine.org as is.

balloonpopper, line 20:
```suggestion
# For more details see: https://docs.escoria-framework.org/camera
```

addons/escoria-core/game/core-scripts/esc/commands/camera_set_pos.gd
```diff
@@ -1,14 +1,14 @@
...
+# Moves the camera to the given absolute position over a time period.
 #
 # **Parameters**
 #
 # - *speed*: Number of seconds the transition should take
 # - *x*: Target X coordinate
 # - "y*: Target Y coordinate
 #
-# For more details see: https://docs.escoria-framework.org/camera
+# For more details see: https://docs.escoria-framework.org/en/devel/getting_started/camera.html
```
balloonpopper, line 11:
```suggestion
# For more details see: https://docs.escoria-framework.org/camera
```

addons/escoria-core/game/core-scripts/esc/commands/camera_set_target.gd
```diff
@@ -1,13 +1,16 @@
...
 #
 # **Parameters**
 #
-# - *speed*: Number of seconds the transition should take
+# - *speed*: Number of seconds the transition should take to move the camera 
+#   to follow `object`
 # - *object*: Global ID of the target object
 #
-# For more details see: https://docs.escoria-framework.org/camera
+# For more details see: https://docs.escoria-framework.org/en/devel/getting_started/camera.html
```
balloonpopper, line 13:
```suggestion
# For more details see: https://docs.escoria-framework.org/camera
```

addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom.gd
```diff
@@ -1,16 +1,18 @@
...
+# would result in no change. The zoom will happen over the given time period.
 #
 # **Parameters**
 #
 # - *magnitude*: Magnitude of zoom
 # - *time*: Number of seconds the transition should take, with a value of `0`
 #   meaning the zoom should happen instantly (default: `0`)
 #
-# For more details see: https://docs.escoria-framework.org/camera
+# For more details see: https://docs.escoria-framework.org/en/devel/getting_started/camera.html
```
balloonpopper, line 15:
```suggestion
# For more details see: https://docs.escoria-framework.org/camera
```

addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_height.gd
```diff
@@ -1,14 +1,14 @@
...
+# Zooms the camera in/out so it occupies the given height in pixels.
 #
 # **Parameters**
 #
 # - *pixels*: Target height in pixels
 # - *time*: Number of seconds the transition should take, with a value of `0`
 #   meaning the zoom should happen instantly (default: `0`)
 #
-# For more details see: https://docs.escoria-framework.org/camera
+# For more details see: https://docs.escoria-framework.org/en/devel/getting_started/camera.html
```
balloonpopper, line 11:
```suggestion
# For more details see: https://docs.escoria-framework.org/camera
```

addons/escoria-core/game/core-scripts/esc/commands/camera_shift.gd
```diff
@@ -15,7 +16,7 @@
 #
 # https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations
 #
-# For more details see: https://docs.escoria-framework.org/camera
+# For more details see: https://docs.escoria-framework.org/en/devel/getting_started/camera.html
```
balloonpopper, line 19:
```suggestion
# For more details see: https://docs.escoria-framework.org/camera
```

addons/escoria-core/game/core-scripts/esc/commands/custom.gd
balloonpopper: So how would you like this reworded please @dploeger ?

addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
balloonpopper: Thanks @dploeger - I've corrected all the camera references accordingly (I'll squash the commits at a later point)

addons/escoria-core/game/core-scripts/esc/commands/custom.gd
dploeger: Hmmmm. Maybe like this?

```
# - *params*: Arguments to the function (array and object parameters are not supported). (...)
```

dploeger: "Any primitive, non-array argument" would include object, right? Or is an object non-primitive? Maybe. It's really nitpicky I'd say. :)

BHSDuncan: primitive = non-object, non-array (usually native/intrinsic data types to the compiler, typically int, float, double, etc., and usually string, but I guess not always).

All good, though.

balloonpopper: That was the reason I removed the word "primative". I didn't think the term was accessible enough to the audience that they would be sure of the meaning. As the developers are confirming here exactly what it entails, it shows me it's not the right word to use - if it's not 100% immediately clear to you, then the users are going to be totally lost.

If you would like to leave the intent, can I suggest a variant of @dploeger's version from 9 hours ago? 
# - *params*:  Any arguments to be passed to the function (array and object parameters are not supported).

BHSDuncan: I'm good with that.

I take your point; I sometimes forget that not everyone who's going to use this is a software engineer or someone who just inherently/intuitively knows what "primitive" means/represents.

balloonpopper, line 12:
```suggestion
# - params: Any arguments to be passed to the function (array and object parameters are not supported). Multiple
```

BHSDuncan:
Can't remember if we came to a decision on this, but we should for consistency's sake:

Are we going with `dialog` or 'dialogue` throughout the docs? We really should pick one and stick with it.

@dploeger @StraToN

balloonpopper:
I wasn't sure so I left it at as-is. I'll update all of them for consistency based on the feedback (there's currently 2 x "dialogue" and the rest are "dialog"  😬 )

BHSDuncan:
> I wasn't sure so I left it at as-is. I'll update all of them for consistency based on the feedback (there's currently 2 x "dialogue" and the rest are "dialog" 😬 )

Yeah no worries. This is something I'd meant to address awhile ago but it kinda slipped through the cracks. Appreciate all your work!

StraToN:
> Can't remember if we came to a decision on this, but we should for consistency's sake:
> 
> Are we going with `dialog` or 'dialogue` throughout the docs? We really should pick one and stick with it.
> 
> @dploeger @StraToN

After checking around for a confirmation of my assumption of the usage of both words, here is what I think we should use:

- "dialog" refers to GUI element, such as "popup dialog" or "dialog box"
- "dialogue" refers to the conversational meaning

This mean that a "dialog tree" is not correct, we should use "dialogue tree" in this case, and so on. WDYT?

addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
StraToN: Yes, that's exactly it. 
If you disable automatic transitions, then you have to manage the transition out as you exit the scene (so in the :exit event, for an ESCExit) *and* you also have to manage the transition in the new room (at the end of the :setup event of the new room, usually)

StraToN (approved):

addons/escoria-core/game/core-scripts/esc/commands/turn_to.gd
```diff
@@ -2,6 +2,12 @@
 #
 # Turns `object` to face another object.
 #
+# Unlike movement commands, `turn_to` will not automatically reference an
+# ESClocation that is a child of an ESCitem. 
+# To turn towards a ESClocation that is a child of an ESCitem, give that 
+# ESClocation a `Global ID`, and set the `object_to_face` parameter to this 
+# `Global ID`.
```
BHSDuncan, line 9:
```suggestion
# Unlike movement commands, `turn_to` will not automatically reference an
# `ESCLocation` that is a child of an `ESCItem.` 
# To turn towards an `ESCLocation` that is a child of an `ESCItem`, give the
# `ESCLocation` a `Global ID` and use this value as the `object_to_face`
# parameter.
```

addons/escoria-core/game/core-scripts/esc_location.gd
```diff
@@ -1,5 +1,8 @@
 # A simple node extending Position2D with a global ID so that it can be
-# referenced in ESC Scripts.
+# referenced in ESC Scripts. Movement-based commands like walk_to_pos will
+# automatically use an ESClocation that is a child of the destination node.
+# Commands like turn_to which are not movement-based will ignore child
+# ESClocations and refer to the parent node.
```
BHSDuncan, line 5:
```suggestion
# referenced in ESC Scripts. Movement-based commands like `walk_to_pos` will
# automatically use an `ESCLocation` that is a child of the destination node.
# Commands like `turn_to`--which are not movement-based--will ignore child
# `ESCLocation`s and refer to the parent node.
```

addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
StraToN: @balloonpopper Does this answer your question? May I mark the discussion as resolved?

balloonpopper: @StraToN - I'll need to update the text as I only mentioned transition in. 
If I'm reading the code correctly, it appears the ":exit" event is actually ":exit_scene". Can you please confirm and then I'll update the documentation to add
"Transitions when leaving a room can be managed manually as part of the `exit_scene` event."

StraToN (approved):

addons/escoria-core/game/core-scripts/esc_location.gd
balloonpopper: I'm not sure how "--" gets translated by the document generator, but I'll take your word for it that that's how it should look @BHSDuncan 😄

addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
StraToN: Correct, the actual event name is `exit_scene`, not `exit`.

BHSDuncan (approved):

addons/escoria-core/game/core-scripts/esc/commands/hide_menu.gd
balloonpopper: I'll leave updating the documentation for hide_menu for now. I want to clarify when you'd expect to use it (see [issue 179](https://github.com/godot-escoria/escoria-issues/issues/179) ) and will modify the text accordingly.

StraToN:
Thanks!

Issue #139, opened 2018-04-26
Fix lockit.gd


mjtorn:
According to https://fr.flossmanuals.net/creating-point-and-click-games-with-escoria/i18n/ translation magic happens with the `lockit.gd` script.

> After it runs, you will get a file called "strings.csv" with all the strings on your esc files, separated by file, and with some context information. You'll also have all the .esc.lockit files generated.

```
$ Godot_v3.0.2-stable_x11.64 -script ../tools/lockit.gd
OpenGL ES 3.0 Renderer: Gallium 0.4 on AMD POLARIS10 (DRM 3.19.0 / 4.14.14-ck1-ryzen, LLVM 3.9.1)
ERROR: load_source_code: Condition ' err ' is true. returned: err
   At: modules/gdscript/gdscript.cpp:792.
ERROR: load: Condition ' err != OK ' is true. returned: RES()
   At: modules/gdscript/gdscript.cpp:1856.
ERROR: _load: Failed loading resource: res://tools/lockit.gd
   At: core/io/resource_loader.cpp:186.
ERROR: start: Failed loading scene: res://tools/lockit.gd
   At: main/main.cpp:1672.
WARNING: cleanup: ObjectDB Instances still exist!
   At: core/object.cpp:1989.
```

It exits with 0.

```
mjt@lechuck:~/src/git_checkouts/escoria (devel *$%<>)$ find . -name strings.csv
mjt@lechuck:~/src/git_checkouts/escoria (devel *$%<>)$ find . -name \*lockit
```

I tested adding some debug prints but nothing really led me to understand wtf is up with this.

#59 should probably be related to fixing this, as well as #138

fleskesvor:
The lockit.gd script is fixed in #140. I will see if there are some other useful features that can be added to it, and then merge it, if no one objects.

PR #507, opened 2022-02-24
fix: parameter validation function cant print necessary arrays on error

balloonpopper:
While looking into [issue 139](https://github.com/godot-escoria/escoria-issues/issues/139) I came across a bug. The validate command can't print the error message it's supposed to. The "%s" that is supposed to print the "arguments" array fails. The error message given is "not enough arguments for format string in operator '%'.

Test this by just changing the arguments.size comparison (line 58, esc_command_argument_descriptor.gd) from
```
if arguments.size() < self.min_args:
```
to
```
if arguments.size() < 50:
```
image

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/types/esc_command_argument_descriptor.gd
+++ b/addons/escoria-core/game/core-scripts/esc/types/esc_command_argument_descriptor.gd
@@ -59,9 +59,7 @@ func validate(command: String, arguments: Array) -> bool:
 		escoria.logger.report_errors(
 			"Invalid command arguments for command %s" % command,
 			[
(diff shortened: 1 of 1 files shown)


balloonpopper:
FYI @dploeger @StraToN @BHSDuncan

dploeger (approved):
Great! Thanks a lot. Does that completely fix #139? 🎁

balloonpopper:
Unfortunately not @dploeger . The number of parameters check (139) is still an outstanding issue. Once 139 is fixed though, this pr fixes the output it will produce. 139 is still broken because a value or null is passed for every parameter, so the too few parameter check never works

StraToN:
Thanks!

PR #510, opened 2022-02-24
chore: add override.cfg to .gitignore

bolinfest:
For example, this makes it easier to do a bunch of testing
to `escoria-ui-simplemouse` without making a temporary change
to `project.godot` because you can create `override.cfg` with
the following:

```
[escoria]
ui/game_scene="res://addons/escoria-ui-simplemouse/game.tscn"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ bin
 .import
 *.translation
 .godot
+
+# Local overrides to project.godot go in override.cfg.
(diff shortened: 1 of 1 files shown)


StraToN (approved):

StraToN:
I think this is a good change indeed. Thanks!

BHSDuncan (approved):

PR #514, opened 2022-02-27
feat: New graphics for room 7

balloonpopper:
New graphics for room 7.

PR #516, opened 2022-03-01
fix: ESCItem that is a trigger should not capture input events

bolinfest:
Fixes https://github.com/godot-escoria/escoria-issues/issues/147

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_item.gd
+++ b/addons/escoria-core/game/core-scripts/esc_item.gd
@@ -270,6 +270,14 @@ func _ready():
 #
 # - event: Triggered event
 func _unhandled_input(event: InputEvent) -> void:
+	# If this is a trigger, then escoria.inputs_manager is not wired up to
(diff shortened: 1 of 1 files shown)


bolinfest:
If you trace through the code, in `ESCItem._ready()`, signals such as `mouse_left_clicked_item` are wired up *only* when `!is_trigger`. Therefore, in `ESCItem._unhandled_input()`, there is no reason to exercise the `emit_signal` calls when `is_trigger` is `true`. At least in this case, it also appears that the `get_tree().set_input_as_handled()` calls are working *against* us, so we should just bail out altogether.

balloonpopper:
Confirmed that this fixes the issue in the new version of room 7 (https://github.com/godot-escoria/escoria-demo-game/pull/514)

BHSDuncan (approved):
Looks good to me, though the other guys will be able to better test this.

dploeger (approved):
Looks good to me as well. Thanks a lot. 🎁

PR #520, opened 2022-03-02
fix: room3 logic issues

balloonpopper:
Fixed various logic issues where the button / bridge weren't in the correct status based on the rooms you'd visited and what you'd already done in room 3. Also added lots more comments.


StraToN (approved):

StraToN:
Indeed that makes many commands clearer with the examples provided.
I especially appreciate the example usage of `custom` ESC function. Thanks !

PR #542, opened 2022-03-12
fix: Reset audio bus layout

dploeger:
Apparently Godot stores the settings of the audio bus layout in the editor locally somewhere. We set "default audio bus layout" to the resource file in this commit and when starting up the auditor, Godot loads the bus layout setting and applies it to the resource file.

This is just to reset the layout to 0db for each channel. If any dev should've changed the layout in their editor, they should reset their layout as well.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,7 @@ bin
 
 # Local overrides to project.godot go in override.cfg.
 /override.cfg
+
+# Disable accidental changes to the default bus layout. If a change to this
+# file is required, use git add --force 
+addons/escoria-core/default_bus_layout.tres
--- a/addons/escoria-core/default_bus_layout.tres
+++ b/addons/escoria-core/default_bus_layout.tres
@@ -1,7 +1,6 @@
 [gd_resource type="AudioBusLayout" format=2]
 
 [resource]
-bus/0/volume_db = -8.89811
 bus/1/name = "SFX"
 bus/1/solo = false
 bus/1/mute = false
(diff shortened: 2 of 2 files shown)


BHSDuncan (approved):

balloonpopper (changes requested):
Can you add it to gitignore as part of the change? I can't see any reason to have it commit if a developer makes a local change.

dploeger:
> Can you add it to gitignore as part of the change? I can't see any reason to have it commit if a developer makes a local change.

Hmm. We need this file. But putting it into .gitignore now would keep it in the history but just wouldn't record any new changes to it, right? And if you specifically want to record a change, you can force it. 

Never did it like this, but looks okay to me. I'd like your opinions on this, @BHSDuncan @StraToN

BHSDuncan:
I'm ok with how this PR handles it. It might seem unlikely, but if a developer decides to set his/her own defaults, it might be better not to have it in the .gitignore file.

balloonpopper:
My Godot editor settings aren't considered part of this repository - when I change my Godot editor's font or other options they don't become a suggested commit - why should my local sound settings be any different?
My thought : If the developer wants to commit their audio settings then let them update their own version of the .gitignore file - but I think the one attached to the demo repository should ignore local user-specific settings.

dploeger:
As we have decided yesterday, added the file to the gitignore with a comment on how to ignore it there.

StraToN (approved):

StraToN:
Thanks a lot !! :partying_face:

PR #547, opened 2022-03-19
feat: intro cutscene

balloonpopper:
image

It's designed for the game's 1280 x 900 resolution, but I've included 1080p files in case it gets upgraded in future.


dploeger:
Pretty awesome! Thanks.

We have two problems here, though, although I don't know if/how to fix them here:

The intro should be skippable. As I hit ESC, the menu came up and when I selected "continue game", the UI reappeared.

I guess it's because the UI is shown regardless whether we're in NO_UI mode. @balloonpopper Could you maybe check out, if you can find a fix for that here?

I'm not really sure if we have some method of skipping cutscenes currently. @StraToN @BHSDuncan do you know if we do?

BHSDuncan:
I'm unaware of any. I don't think we currently have any way of mapping input to being able to skip any ESC commands.

StraToN (approved):

BHSDuncan (approved):

PR #549, opened 2022-03-21
prefer .get() to .has() to eliminate extra lookup

bolinfest:


StraToN (approved):

BHSDuncan:
I think this is going to become moot once the "instant transition" ticket/PR is finished as I've overhauled the object manager and this method has changed a lot. I'd suggested we just close this PR, if that's ok.

Also, while I'm sure this is known, it bears repeating that "has" lookups can save a lot of time in the case where an object being retrieved is especially large. Although I'm not sure if this would ever happen here, it might be best to still do the lookup (if this wasn't moot :) ). But anyway, this will all have changed with the PR mentioned above.

PR #550, opened 2022-03-21
fix: exit loop as soon as command is found

bolinfest:


BHSDuncan:
This is a debate as old as the computer science hills, and I could honestly go either way on it.

@dploeger since @StraToN is ok with it, if you're ok with it, too, then I'll happily sign off. I have no strong preference either way.

dploeger (approved):

PR #552, opened 2022-03-23
fix: Added animations for the worker character

balloonpopper:
Added new animations for the worker. Note that fixing this showed that there's another bug ([201](https://github.com/godot-escoria/escoria-issues/issues/201)) that needs fixing in relation to talking animations. This means that demo levels where the player talks to the NPC will now be broken as before the NPC only faced one direction.

Fixes https://github.com/godot-escoria/escoria-issues/issues/173


StraToN (approved):

BHSDuncan (approved):

dploeger (approved):
Wonderful! Thanks. 🎁

StraToN:
Thanks a lot!!

PR #532, opened 2022-03-08
feat: room6 New graphics

balloonpopper:
This change is 95% complete. There's an outstanding issue
See https://github.com/godot-escoria/escoria-issues/issues/177

I'll need to remove the hack (room6.esc lines 14 and 15) that immediately unlock the second door (to allow the bug to be investigated) before the room is production-ready.

image

PR #554, opened 2022-03-24
feat: Save objects states in Obj Manager to keep them btwn rooms

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/193

Unblocks https://github.com/godot-escoria/escoria-demo-game/pull/532

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
@@ -163,8 +163,12 @@ func register_object(object: ESCObject, room: ESCRoom = null, force: bool = fals
 					]
 			]
 		)
-
 		return
+	# Object exists in room, set it to is last state (if different from 
+	# "default")
+	elif _object_exists_in_room(object, room_key):
+		# Object is already known, set its state to last known state
+		object.set_state(get_object(object.global_id).state)
 
 	# If the object is already connected, disconnect it for the case of
 	# forcing the registration, since we don't know if this object will be
@@ -205,6 +209,23 @@ func register_object(object: ESCObject, room: ESCRoom = null, force: bool = fals
 
 	var objects: Dictionary = _get_room_objects_objects(room_key)
 	objects[object.global_id] = object
+	
+	# If object state is not STATE_DEFAULT, save it in manager's object states 
+	if object.state != ESCObject.STATE_DEFAULT:
+		if get_object(object.global_id) == null:
+			escoria.logger.report_errors(
+			"ESCObjectManager:register_object()",
+			[
+				"Object with global id %s in room (%s, %s) not found in Object Manager" %
+					[
+						object.global_id,
+						room_key.room_global_id,
+						room_key.room_instance_id
+					]
+			]
+		)
+		else:
+			get_object(object.global_id).state = object.state
 
 	# If this is the first object for the room, that means we have a brand new
 	# room and it needs to be setup and tracked.
@@ -346,8 +367,9 @@ func unregister_object(object: ESCObject, room_key: ESCRoomObjectsKey) -> void:
 		if object.node != null:
 			object.node = object.node.duplicate()
 			register_object(object, null, true)
-
-	room_objects.erase(object.global_id)
+	
+	if object.state == ESCObject.STATE_DEFAULT:
+		room_objects.erase(object.global_id)
 
 	# If this room is truly empty, it's time to do away with it.
 	if room_objects.size() == 0:
@@ -438,8 +460,7 @@ func _is_current_room(container: ESCRoomObjects) -> bool:
 # **Returns** True iff container represents the the object manager entry specified
 # by room_key.
 func _compare_container_to_key(container: ESCRoomObjects, room_key: ESCRoomObjectsKey) -> bool:
-	return container.room_global_id == room_key.room_global_id \
-		and container.room_instance_id == room_key.room_instance_id
+	return container.room_global_id == room_key.room_global_id
 
 
 # Checks whether an entry in the object manager array corresponds to the passed in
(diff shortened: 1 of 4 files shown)


BHSDuncan:
How are we going to merge this? If we put it into `develop` before we merge the "instant transition" PR, it's going to cause merge conflicts, especially given how very much different the object and room managers have become.

How do we want to handle this?

dploeger (dismissed):
Thanks! 🎁

dploeger:
Sorry, didn't read your comment, @BHSDuncan. Maybe you could incorporate this change in your branch then? It isn't quite big.

StraToN:
@BHSDuncan Agree with @dploeger . I'm fine if you cherry-pick this commit into #515 and we close this one.

StraToN:
Status: waiting for #515 to be merged. Will have to be rebased after that.

StraToN:
To test:
- checkout [branch "room6" from @balloonpopper ](https://github.com/balloonpopper/escoria-demo-game/tree/room6)
- rebase this branch on origin/develop
- If you want to see the issue, test (apply scenario below)
- apply this PR onto the resulting branch
- run

Scenario:
1. Access room06 -> notice that right door is closed
2. exit room06 to room05
3. Get back to room06
-> if PR not applied, notice that right door is now open
-> else, notice that right door is still closed

balloonpopper:
The state was preserved in my testing - the door stays "closed" with the fix applied.
Other issues were found in the transition between these rooms (as dicsussed in the channel) but they seem unrelated to this PR.

addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
```diff
@@ -207,6 +214,10 @@ func register_object(object: ESCObject, room: ESCRoom = null, force: bool = fals
 
 	var objects: Dictionary = _get_room_objects_objects(room_key)
 	objects[object.global_id] = object
+	
+	# If object state is not STATE_DEFAULT, save it in manager's object states 
+	if object.state != ESCObject.STATE_DEFAULT:
+		objects_states[object.global_id] = object.state
```
BHSDuncan, line 220: In the future, wouldn't this prevent us from being able to load multiple rooms (for whatever reason, e.g. pre-loading) that contain the same object at the same time? Would it make sense to include these states as part of the larger "room->object" structure for the object manager? Or is this just not a concern?

BHSDuncan:
> The state was preserved in my testing - the door stays "closed" with the fix applied. Other issues were found in the transition between these rooms (as dicsussed in the channel) but they seem unrelated to this PR.

Can you also maybe check if it's preserved across saving and loading? It should be based on what I see, but best to test. :)

balloonpopper (approved):

balloonpopper:
> Can you also maybe check if it's preserved across saving and loading? It should be based on what I see, but best to test. :)

I was asked to hold off on save testing while @StraToN does some more work on the save system.

balloonpopper (approved):

addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
```diff
@@ -165,8 +165,11 @@ func register_object(object: ESCObject, room: ESCRoom = null, force: bool = fals
 					]
 			]
 		)
-
 		return
+	# Object exists in room, set it to is last state (if different from "default"
```
BHSDuncan, line 169:
```suggestion
	# Object exists in room, set it to is last state (if different from "default")
```

addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
```diff
@@ -207,6 +210,10 @@ func register_object(object: ESCObject, room: ESCRoom = null, force: bool = fals
 
 	var objects: Dictionary = _get_room_objects_objects(room_key)
 	objects[object.global_id] = object
+	
+	# If object state is not STATE_DEFAULT, save it in manager's object states 
+	if object.state != ESCObject.STATE_DEFAULT:
+		get_object(object.global_id).state = object.state
```
BHSDuncan, line 216: There needs to be a null check based on what `get_object` returns, since returning null is a possibility.

addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
```diff
@@ -439,8 +447,7 @@ func _is_current_room(container: ESCRoomObjects) -> bool:
 # **Returns** True iff container represents the the object manager entry specified
 # by room_key.
 func _compare_container_to_key(container: ESCRoomObjects, room_key: ESCRoomObjectsKey) -> bool:
-	return container.room_global_id == room_key.room_global_id \
-		and container.room_instance_id == room_key.room_instance_id
+	return container.room_global_id == room_key.room_global_id
```
BHSDuncan, line 450: I'm curious about this change. I remember making this method to check this very specific combination in case we ever have two rooms setup but that are different instances. What benefit do we gain by removing that part of the check? I don't think it does any harm to leave it as it was.

BHSDuncan (changes requested):

addons/escoria-core/game/core-scripts/esc/esc_object_manager.gd
StraToN: This is because when I save the ESCItem state in the Dictionary, the key is composed of `global_id` and `room_instance_id`. However, when I exit the room to enter another one, then go back to the first room, the `room_instance_id` saved as key is different from the new `room_instance_id` of the room I just went back.

I can revert this change, but then I'll have to create a new test function that doesn't test the instance id.

BHSDuncan: In other words, since the room_object container is no longer deleted when you leave a room if an object's state has changed, room equality is now an issue. Got it. Will take a look.

BHSDuncan (approved):

PR #556, opened 2022-03-26
Fix: Avoid globals starting with 'i/'

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/153

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
@@ -92,15 +92,7 @@ func set_global(key: String, value, ignore_reserved: bool = false) -> void:
 				"Global key %s is reserved and can not be overridden" % key
 			]
 		)
-	
-	if key.count("/") > 1 and key.substr(1, 1) == "/":
-		escoria.logger.report_errors(
-			"ESCGlobalsManager.set_global: Global name error",
-			[
-				"Global key %s has more than 1 occurrence of '/'" % key
-			]
-		)
-		
+
 	emit_signal(
 		"global_changed",
 		key,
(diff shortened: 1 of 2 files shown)


addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
```diff
@@ -92,6 +92,15 @@ func set_global(key: String, value, ignore_reserved: bool = false) -> void:
 				"Global key %s is reserved and can not be overridden" % key
 			]
 		)
+	
+	if key.count("/") >= 2 and key.substr(1, 1) == "/":
```
BHSDuncan, line 96:
```suggestion
	if key.count("/") > 1 and key.begins_with("/"):
```

BHSDuncan (changes requested):
More of me being pedantic. :)

addons/escoria-core/game/core-scripts/esc/esc_globals_manager.gd
StraToN: You suggestion is incorrect: the "/" character can only be located at second position in the string, not as the first character.

dploeger:
@StraToN  I'm confused. Wouldn't it make more sense to check in `inventory_add` whether the given global_Id doesn't have a "/" in its name?

StraToN:
@dploeger That would also make sense but in this PR I wanted above all to forbid a global having 2 / characters in its name. Since inventory_add calls set_global, I intended to catch all cases there.

dploeger:
Shouldn't have a global **no** / in the name at all? Even i/ is only a hack to work with inventory items. 🤔

StraToN:
We could indeed modify the global manager to separate globals from inventory globals instead. Technically, nothing blocks the use of / characters in globals. And I'm just thinking that could actually be useful to keep, so that gamedevs can kind of "classify" their globals.

Which would advocate for a change in inventory_add instead as you mentioned. 

I'll change that tomorrow morning.

BHSDuncan (approved):

balloonpopper (approved):

PR #558, opened 2022-03-29
RFC: RoomScript

bolinfest:
This is a demo of something I have been working on, tentatively called RoomScript.

The overarching goal is to define a room in a single file, more in the style of
**Delores: A Thimbleweed Park mini-adventure**:
https://github.com/grumpygamer/DeloresDev/blob/master/Scripts/Rooms/MainStreet.dinky

Today, Escoria requires a lot of names:
- Godot "name" of a Node
- Escoria `global_id`
- Path to `.esc` file

RoomScript eliminates the individual `.esc` files. It still supports the `global_id` field,
but it often turns out to be unnecessary, as many scripts can use the existing Node
name rather than the `global_id`.

The `RSCRoom` (which is a subclass of `ESCRoom`) maintains its global id,
but there must be a matching `.room` file in the same directory. Therefore,
instead of having to declare the path to the `.room` file as a property of `RSCRoom`,
we exploit the convention and rely on this logic instead:

```gdscript
static func find_script_file_for_room(room: RSCRoom) -> String:
	# For now, we enforce the convention that the name and Escoria global_id
	# should match, which must also be the name of the subfolder under rooms/,
	# as well as the name of the `.room` file. (If the rooms/ folder becomes
	# too large, we can revisit this.)
	var room_id = room.global_id
	var room_dir = zero_pad_room_id(room_id)
	var res = "res://game/rooms/%s/%s.room" % [room_dir, room_id]
	escoria.logger.debug("room path: %s" % res)
	return res
```

Within the `.room` file, there are simple declarations/blocks like `event setup {}`
and `item button {}` that contain ESC. Again, this is done in favor of the invidual `.esc` files.
In practice, many items support only one verb, like `look` or `use`, so creating a new
`.esc` file for each can feel like unnecessary overhead. This is also one less field you
need to enter for each item in the Godot editor.

As you can see, the end result is in `game/rooms/room02/room2.room` and all the
`.esc` files for `room02` have been eliminated at the top of this stack, yet everything
still works as it did before.

I certainly don't expect this PR to be accepted as-is. As you can see, the logic in
`rsc_item.gd` and `rsc_room.gd` is pretty hacky to make this work. Likewise, I had
to create parallel versions of a number of built-in commands in
`addons/escoria-roomscript/commands/` in order to support the local names instead
of the global ids. My hope is that we could evolve the framework to support this
style of development, even if it is still an addon.


dploeger:
This is pretty awesome. Just to be sure: You're not proposing this as an official plugin with a conversion of the given rooms but rather as a showcase of your own plugin?

I really like what you did here, but I don't see it as an official plugin because we have to maintain ESC and that's already hard. 😉 

I'd love to see it as your own Escoria plugin (like the way I did the Escoria ink bridge), so that the "Escoria ecosystem" grows with it.

bolinfest:
@StraToN I think the biggest thing I need is the ability for a different way to resolve identifiers to objects, particulary in ESC. For example, the way I had to rewrite all the commands in this PR so I can intercept the calls and run `RSCRoom.rewrite_command_params()` should be done in one place somewhere else so the original commands can be used as-is. A somewhat "natural" place to move that would be inside `ESCObjectManager`, but the current design of `ESCObjectManager` is not particularly friendly to inheritance, nor is there a clean way to assign a different `globals_manager` on startup in `addons/escoria-core/game/escoria.gd` right now. These are all things that could be changed, but I'm not sure what the project would be most open to.

Also, the name resolution I add in RoomScript relies heavily on this bit:

```gdscript
static func _resolve_item_id(item_id: String) -> String:
	var room = find_room_for_item(escoria.main.current_scene)
	if room:
		var global_id = room.resolve_item_id(item_id)
		if global_id:
			return global_id
	return item_id
```

That is, it goes through `escoria.main.current_scene` and assumes that it is set and the `RSCRoom` should be ready and available in the tree when `_resolve_item_id()` is called. I haven't looked at the work that is currently in flight for #515 yet, though I worry something in there could break some of the current assumptions in my code.

PR #560, opened 2022-04-02
RFC: match statement

bolinfest:
As an initial attempt, I first made a small update to initially set a global in `room02_bridge.esc`:

```
:setup
# TODO: only set this if not yet set?
set_global r2_look_dialog_advance 0
```

which then makes the `:look` command in `button.esc` a bit cleaner when using the new `match` statement:

```
:look
>> r2_look_dialog_advance
	0 =>
		say player "That button must activate the bridge."
		inc_global r2_look_dialog_advance 1
	1 =>
		say player "I already said that this button must activate the bridge."
		inc_global r2_look_dialog_advance 1
	2 =>
		set_angle player 180
		say player "Listen closely."
		say player "This"
		say player "button"
		say player "must"
		say player "activate"
		say player "the bridge."
		inc_global r2_look_dialog_advance 1
	3 =>
		say player "I give up."
		say player "<sob>"
```

For the moment, to avoid introducing a new keyword (which would take away a possible command name), `>>` is used to indicate the start of a match statement. As you can see, the condition arms use a `=>` syntax like Rust (and a number of other languages that supports pattern matching).

Arguably, using a match statement avoids repeating the comparison valuable, eliminates the need for `stop` in many cases, and makes it easier to see that all cases are covered.


bolinfest:
Admittedly, this is not "real" pattern matching. It's more of a glorified `switch` statement, but it's a start. Honestly, I didn't realize how flexible `match` is in GDScript today: https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html?highlight=match#match.

addons/escoria-core/game/core-scripts/esc/esc_compiler.gd
```diff
@@ -150,6 +153,47 @@ func _compile(lines: Array) -> Array:
...
+					lines.push_front(next_line)
+					break
+
+			if match_lines.size() > 0:
+				escoria.logger.trace(
+					"Compiling the next %d lines into the match expr" % \
+						match_lines.size()
+				)
+				var pairs = match_expr.parse(match_lines)
+				escoria.logger.trace("%d arms found for match" % pairs.size())
```
balloonpopper, line 183: Can we rename "arm" to something like "condition_option" please? It's not very intuitive.

balloonpopper (changes requested):
Assuming this gets the green light, please update the documentation as part of the PR.

balloonpopper:
Also, is there any way to add conditional logic to this? i.e. [issue 220](https://github.com/godot-escoria/escoria-issues/issues/220)

StraToN:
Very good improvement to the language. I have nothing particular to add about it, apart that obviously documentation is required in the code and some TODOs need to be adresse. 

Otherwise, great addition, thanks!

addons/escoria-core/game/core-scripts/esc/esc_compiler.gd
dploeger: Or "branch".

dploeger (changes requested):

StraToN:
Closing for now, but we'll try to deal with this proposal in the new syntax @BHSDuncan

PR #563, opened 2022-04-04
feat: Added print command to print output at any debug level

balloonpopper:
The debug command only works when the log level is set to "DEBUG". This command allows for debug message printing at any log level.
Alternatively, we could make debug print at all debug levels.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/print.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/print.gd
@@ -1,0 +1,26 @@
+# `print string`
+#
+# Prints a message to the log. Use this for debugging game state.
+#
+# **Parameters**
+#
+# - *string*: The string to log
+#
+# @ESC
+extends ESCBaseCommand
+class_name PrintCommand
+
+
+# Return the descriptor of the arguments of this command
+func configure() -> ESCCommandArgumentDescriptor:
+	return ESCCommandArgumentDescriptor.new(
+		1,
+		[TYPE_STRING],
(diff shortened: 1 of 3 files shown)


addons/escoria-core/game/core-scripts/log/esc_logger.gd
```diff
@@ -54,6 +54,15 @@ func _init():
...
 	)
 
+# Print a message
+#
+# #### Parameters
+#
+# * string: Text to log
+# * args: Additional information
+func print(string: String):
+	_log("(?)\t" + string)
```
StraToN, line 64: `_log()` function does write in the logfile, which is precisely what you want to avoid :)

StraToN (changes requested):

addons/escoria-core/game/core-scripts/esc/commands/print.gd
```diff
@@ -0,0 +1,26 @@
+# `print string`
```
StraToN, line 1: Doesn't `debug` ESC command do the same thing?

StraToN (changes requested):

addons/escoria-core/game/core-scripts/esc/commands/print.gd
balloonpopper: debug only works if you're in debug log level. I think a better solution is to not use this command but instead allow print to work for all log levels (info/warning etc). I'll remove this PR if you agree.

PR #564, opened 2022-04-04
feat: Added print command to print output at any debug level

balloonpopper:
The current debug command is restricted to "log level = debug".  This means that you can't use it to do your own debugging if you use a different log level.
This PR adds a print command. It needs to be a separate command to debug as debug is used by the engine to print all the internal engine state. As a result, if debug was allowed to print at any log level, you would end up with engine debugging messages even at "info" level.


BHSDuncan (approved):

StraToN:
Thanks!

PR #566, opened 2022-04-07
feat: adjustable text speed

balloonpopper:
Adjustable text display speed.
Fixes https://github.com/godot-escoria/escoria-issues/issues/194

### DIFF
--- a/addons/escoria-core/game/core-scripts/save_data/esc_save_manager.gd
+++ b/addons/escoria-core/game/core-scripts/save_data/esc_save_manager.gd
@@ -412,6 +412,7 @@ func save_settings():
 	settings_res.fullscreen = escoria.settings.fullscreen
 	settings_res.skip_dialog = escoria.settings.skip_dialog
 	settings_res.custom_settings = escoria.settings.custom_settings
+	settings_res.text_speed = escoria.settings.text_speed
 
 	var directory: Directory = Directory.new()
 	if not directory.dir_exists(settings_folder):
--- a/addons/escoria-core/game/core-scripts/save_data/esc_savesettings.gd
+++ b/addons/escoria-core/game/core-scripts/save_data/esc_savesettings.gd
@@ -46,3 +46,7 @@ export var skip_dialog: bool = true
 
 # Dictionary containing all user-defined settings.
 export var custom_settings: Dictionary
+
+# Speech text speed
+export var text_speed: float = ProjectSettings.get_setting(
+	"escoria/dialog_simple/text_speed_per_character")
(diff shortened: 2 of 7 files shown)


StraToN (approved):

dploeger:
Hm. I'm not sure about putting this into core. Don't get me wrong, this setting makes a lot of sense in terms of a11y, but it's very UI+Dialog Manager-specific (and why stop at text_per_character and not provide settings to the other settings in the stock dialog manager).

So I at least propose to use custom_settings to store this value and not a hardcoded setting.

@StraToN @BHSDuncan what's your take on this?

balloonpopper:
I don't mind moving the text speed settings to custom_settings in theory. The only custom setting I can see in the code ("a_custom_setting") is a very synthetic one so I don't have a good example to follow. 
Do you have any pointers about how the change to a custom setting would modify the current codebase? My main concern is I can't see how it can be kept out of core when it needs changes to addons/escoria-core/ui_library/menus/options/options.gd without moving all menus to being non-core.

dploeger:
Hm... You're right with the part from the ui library there... Now I'm biased. 😆 

Let's think about something more sophisticated here perhaps. Something like: The dialog manager exposes new options to the game, that need to be incorporated into the options screen. Maybe we have a list of "custom options" in escoria core, that the dialog manager could add theirs to and the stock options screen would pick them up?

Or does that go too far? It all boils down to what we want to have (and first and foremost: maintain) in the stock and ui library. Calling @StraToN  and @BHSDuncan  here again. :)

BHSDuncan:
I don't know if it goes TOO far...I mean, it might make sense to have external UI implementations be able to register arbitrary settings related to their respective implementations with core. Or somehow otherwise combine the two. Definitely warrants further discussion.

StraToN:
What about externalizing UIs that are currently in `addons/escoria-core/ui_library` in its own plugin? They're here for demonstration purpose.

It is also true that escoria-demo will not run if that plugin is not enabled, so that's virtually creating dependancies between core and other plugins.

About custom settings, what about having each plugin manage their own subcategory of settings?
That is, instead of setting them among existing subcategories defined by Core (`escoria/core/`, `escoria/ui/` and such), we could define a rule that forces each plugin to have its own set of settings sorted in `escoria/<subcategory name>/`.

StraToN:
@balloonpopper If possible I'd advise to have the UI-specific parameters that should appear in Project Settings to be managed by the plugin upon activation. That is: text speed should not appear in Project Settings if that UI plugin is disabled. WDYT?

StraToN:
After discussion with OP @balloonpopper , we agreed to close that PR since it needs to be remade after the big plugin remake, that broke this PR.
Ideally, no core code should be altered by this feature.
If specific settings are to be added, they should belong to the UI plugin, not core plugin.

PR #570, opened 2022-04-12
PR to cleanup known outstanding issues caused by "instant transition" code updates

BHSDuncan:


addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -124,7 +127,13 @@ func _process(delta: float) -> void:
 			if event_flags & ESCEvent.FLAG_NO_SAVE:
 				escoria.save_manager.save_enabled = false
 
-			_running_events[channel_name].run()
+			var rc = _running_events[channel_name].run()
```
StraToN, line 130: Won't this be an issue for events happening while the player has control? For example, activate an alarm so that the guard turns back to the player for a few seconds, during while the player has to perform a specific action and remain unseen?

balloonpopper:
* Changed the "Node" nodes in room 6 to "Node2D" as you suggested. When I leave room 5 I get a transition out of room 5 twice then transition in to room 6. That corrected the half drawn room issue - I'll need to go fix a few rooms. 

* Removed L62-92 on my room6 branch and the double transition disappears. There's a noticable lag that I don't recall seeing when we initially tested the "final" version of instant transitions. It may be the "Parsing line wait 0.1" which is doing it as that seems to line up roughly with the spot in the debug where I think it occurs. I'm not sure why it seems to be more noticeable now than before.

* In the transition it parses mark.esc 3 times but I've raised that as a new bug as that's not a transition thing. 
Tootle — Today at 11:03 PM
* There's still a frame or two lag where the right door appears open before the closed door is drawn when starting room 6 directly.

* Noticable delay from room 3 to room 2

* Bug exiting room 4 to room 5 - room 4 is being drawn twice. Now it does have a Node rather than a Node2D node in it, but it's still broken when that's corrected.

* Bug exiting room 7 to room 6 (or 7 to 8) - The camera resets to 0,0 before it switches to room 6

* Bug when jumping direct to room 7 - the words above the 4 buttons seem to appear before the rest of the room

* Big lag room 8 to room 7 (but I'm pretty sure that's just printing with debug mode as room 7 is one the most script intensive)

* From 8 to 7, the camera jumps to the start_location for a frame (showing the left hand door) instead of at the right hand door

* Jumping to 9, the bottles in the cupboards are drawn before everything else.

I'll stop there as I think there's a few common errors that are repeating. All the above is with 62-92 commented out.

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
BHSDuncan: As noted in the Discord channel: Yeah, I hadn't considered the other channels. I'll try implementing a "per channel yield" check and we can test that out.

BHSDuncan:
I wonder if the delays are just part of that "wait" command. We can discuss together.

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
BHSDuncan: This has been put in. I'll mark this as resolved but please do test.

balloonpopper:
As discussed in the Discord channel, the bugs mentioned above are pretty much all ironed out. In room9 there is a very small (i.e. ~1/60th of a second) space of time where you see the cupboards change from one animation state to another. It's hard to spot - even when I made 43 copies of the open door animation to exaggerate the effect, so I'm happy to say it's within acceptable limits and approve the change as such. 
Camera resets, draw-order issues, double transition outs, and transition-based-lag have all been fixed based on my testing.

balloonpopper (approved):

StraToN (approved):

PR #572, opened 2022-04-15
refactor: remove deprecated debug command

balloonpopper:
- Removed "debug" as it's been replaced by print (which works at all debug levels)
- Fixed issue where print didn't interpolate the globals in the text in all instances
- Updated the debug console to use print over debug. Compared output in the debug window when using print vs debug and the 11 debug output lines are identical (repacing everywhere that says "debug" with "print")
- Deduplicated the replace_globals function

Fixes : https://github.com/godot-escoria/escoria-issues/issues/221

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/print.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/print.gd
@@ -23,5 +23,6 @@ func configure() -> ESCCommandArgumentDescriptor:
 
 # Run the command
 func run(command_params: Array) -> int:
-	print(command_params[0])
+	# Replace the names of any globals in "{ }" with their value
+	print(escoria.logger.replace_globals(command_params[0]))
 	return ESCExecution.RC_OK
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -21,7 +21,7 @@ signal background_event_finished(return_code, event_name, channel_name)
 
 
 # Pre-defined ESC events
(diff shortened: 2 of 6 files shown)


addons/escoria-core/game/core-scripts/esc/commands/print.gd
```diff
@@ -23,5 +23,6 @@ func configure() -> ESCCommandArgumentDescriptor:
 
 # Run the command
 func run(command_params: Array) -> int:
-	print(command_params[0])
+	# Replace the names of any globals in "{ }" with their value
+	print(escoria.logger._replace_globals(command_params[0]))
```
BHSDuncan, line 27: If you're going to use `_replace_globals`, then it'll either need to be renamed to follow a public method convention, or you'll have to make a public method that defers to `_replace_globals` in the same class.

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -22,6 +22,7 @@ signal background_event_finished(return_code, event_name, channel_name)
 
 # Pre-defined ESC events
 const EVENT_DEBUG = "debug"
```
BHSDuncan, line 24: If we're removing `debug` all together, can we remove this constant? Or is it used elsewhere?

BHSDuncan (changes requested):

BHSDuncan (approved):

balloonpopper:
Fixes https://github.com/godot-escoria/escoria-issues/issues/204
Fixes https://github.com/godot-escoria/escoria-issues/issues/221

StraToN (approved):

PR #576, opened 2022-04-19
fix: Enforce rooms to be located at (0,0)

balloonpopper:
@BHSDuncan has already closed this issue as the bug was a small room misconfiguration, however on reflection, I thought it would be good to fix as it is easy to detect and flag to the developer. If we can do it in the demo levels, they can do it in the real world.
If the character is at the edge of a terrain, and you try to make them walk further, the walk_stop command will reset their location by the room offset amount (ignoring any terrains and repositioning them by an addition room's pixel offset amount each time you click).

https://github.com/godot-escoria/escoria-issues/issues/218

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_room_manager.gd
@@ -178,6 +178,17 @@ func init_room(room: ESCRoom) -> void:
 		if escoria.main.current_scene == null:
 			escoria.main.set_scene(room)
 
+	# If the room node isn't at (0,0), the walk_stop function will offset the 
+	# player by the same number of pixels when they're at the terrain edge and
+	# move them when it shouldn't.
+	if room.position != Vector2(0,0):
+		escoria.logger.report_errors(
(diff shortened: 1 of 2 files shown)


BHSDuncan:
Hmm, is it possible that a developer could have a good reason for not having the room located at (0, 0)? Yes, it looks like it would affect how walking works, but playing devil's advocate, do we want to impose a constraint that might prevent something in the future?

Personally I'm ok with doing so, although maybe we should make it a warning and not so much an error.

StraToN (approved):

StraToN:
> Hmm, is it possible that a developer could have a good reason for not having the room located at (0, 0)? Yes, it looks like it would affect how walking works, but playing devil's advocate, do we want to impose a constraint that might prevent something in the future?
> 
> Personally I'm ok with doing so, although maybe we should make it a warning and not so much an error.

The ESCRoom node itself should be in (0,0). Its content can be wherever gamedev wants, even if some content is suppose to be on the left of the room.

balloonpopper:
@BHSDuncan Can you please approve?

BHSDuncan (approved):

PR #578, opened 2022-04-19
Fix: removed calls to inexisting "escoria" singleton after plugin disbld

StraToN:
This fix removes all direct calls to "escoria" singleton that blocks re-enabling the plugin after it was disabled (because the singleton was removed).
Replaced by using `get_node("escoria")` instead.

Also added implementions of `_enable_plugin()` and `_disable_plugin()` virtual functions from EditorPlugin class.

### DIFF
--- a/addons/escoria-core/plugin.gd
+++ b/addons/escoria-core/plugin.gd
@@ -2,6 +2,23 @@
 tool
 extends EditorPlugin
 
+# The path to node 'escoria'.
+const ESCORIA: String = "/root/escoria"
+
+
+# Returns the 'escoria' singleton/autoload.
+#
+# *Returns*
+# The escoria singleton.
+func _get_escoria():
+	var escoria_singleton = get_node(ESCORIA)
+	if escoria_singleton == null:
+		printerr(
+			"Escoria could not load the 'escoria' singleton/autoload.\n",
+			"Please try to disable and re-enable 'Escoria' plugin."
+		)
+	return escoria_singleton
+
 
 # Virtual function called when plugin is enabled.
 func _enable_plugin():
@@ -11,8 +28,8 @@ func _enable_plugin():
 	)
 
 	# Add input actions in InputMap
-	InputMap.add_action(get_node("escoria").inputs_manager.SWITCH_ACTION_VERB)
-	InputMap.add_action(get_node("escoria").inputs_manager.ESC_SHOW_DEBUG_PROMPT)
+	InputMap.add_action(_get_escoria().inputs_manager.SWITCH_ACTION_VERB)
+	InputMap.add_action(_get_escoria().inputs_manager.ESC_SHOW_DEBUG_PROMPT)
 
 	# Define standard settings
 	ProjectSettings.set_setting(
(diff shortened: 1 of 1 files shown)


addons/escoria-core/plugin.gd
```diff
@@ -3,16 +3,16 @@ tool
...
+func _enable_plugin():
 	add_autoload_singleton(
 		"escoria",
 		"res://addons/escoria-core/game/escoria.tscn"
 	)
 
 	# Add input actions in InputMap
-	InputMap.add_action(escoria.inputs_manager.SWITCH_ACTION_VERB)
-	InputMap.add_action(escoria.inputs_manager.ESC_SHOW_DEBUG_PROMPT)
+	InputMap.add_action(get_node("escoria").inputs_manager.SWITCH_ACTION_VERB)
```
jeancallisti, line 14: Suggestion : You might want to even create a wrapper for get_node("escoria") that acts as some sort of factory. Something like Escoria.get_escoria(). 

If the node is not present, display a message in the console that says to the user that Escoria core is missing (or disabled) and that it's not the intended behaviour. And what they should do to fix it (re-enable plugin or whatnot).

jeancallisti: @BHSDuncan but if you can't refer to the singleton instance by a hard-coded name ("escoria") then how can you ever refer to it? By getting the (hopefully) only root node's child that has the Escoria class name?

addons/escoria-core/plugin.gd
```diff
@@ -2,6 +2,23 @@
...
+const ESCORIA: String = "/root/escoria"
+
+
+# Returns the 'escoria' singleton/autoload.
+#
+# *Returns*
+# The escoria singleton.
+func _get_escoria():
+	var escoria_singleton = get_node(ESCORIA)
+	if escoria_singleton == null:
```
BHSDuncan, line 15:
```suggestion
	if not is_instance_valid(escoria_singleton):
```

BHSDuncan: Also, I think it would make sense to make this into a true singleton instead of calling `get_node` with each call. I.e. make an instance variable (I realize we don't really have static variables in GDScript), check for null, populate it if it's null, and if it can't be loaded, throw an error. Then return the instance var.

BHSDuncan (changes requested):

addons/escoria-core/plugin.gd
BHSDuncan: What are you talking about? My point is that using that string EVERYWHERE is an issue, and to reduce it down to a single location. I'm effectively agreeing with what you're suggesting.

BHSDuncan (approved):

balloonpopper (approved):
Confirmed I can disable and reenable the core plugin and everything continues to work.
Can you please confirm @StraToN whether this bug is related?
https://github.com/godot-escoria/escoria-issues/issues/252

StraToN:
> Confirmed I can disable and reenable the core plugin and everything continues to work. Can you please confirm @StraToN whether this bug is related? [godot-escoria/escoria-issues#252](https://github.com/godot-escoria/escoria-issues/issues/252)

Responded on https://github.com/godot-escoria/escoria-issues/issues/252: not related

PR #580, opened 2022-04-19
fix: incorrect dialog player type

BHSDuncan:
An underlying bug also existed where multiple `say` commands in the same event caused an issue if one of the subsequent `say` commands in said event used an invalid dialog player type. It looks like re-using the same player as such would throw an error with trying to connect the same signal (despite being a 'oneshot'), but this could be either because of a race condition or that the manager was being marked for deletion and not processed in a timely fashion.

### DIFF
--- a/addons/escoria-core/game/scenes/dialogs/esc_dialog_player.gd
+++ b/addons/escoria-core/game/scenes/dialogs/esc_dialog_player.gd
@@ -105,6 +105,8 @@ func say(character: String, type: String, text: String) -> void:
 			var _manager: ESCDialogManager = load(_manager_class).new()
 			if _manager.has_type(type):
 				_dialog_manager = _manager
+			else:
+				_dialog_manager = null
 
 	if _dialog_manager == null:
 		escoria.logger.report_errors(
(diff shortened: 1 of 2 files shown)


balloonpopper (approved):

BHSDuncan:
This one was interesting but it bear closer scrutiny, especially with the change from a "oneshot" connection to a normal one: See the commit message above, but it seemed like checking for and disconnecting/connecting another oneshot signal inside the same event wasn't behaving correctly, and I can't really see a downfall with doing it this way.

(I've also left in the original incorrect `avatar` argument for now, but we should probably correct it. I just didn't want its correction being conflated with the bug corrected in this PR.)

@StraToN @dploeger WDYT?

balloonpopper (approved):
Confirmed with both the old and new versions of the game/rooms/room01/esc/wall_item_popupdialog.esc script that this fixes the issue.

2022-4-20T125619 (E)    Errors in file esc_dialog_player.gd:say
No dialog manager supports the type avatar_dialog_player

StraToN (approved):

StraToN:
Thanks!

PR #582, opened 2022-04-20
Fix: disable ESCGame inputs until signal room_ready is emitted

StraToN:
Note: this bug happens on Godot 3.4.4, using simple-mouse plugin.

Fixes a crash happening when game scene was receiving inputs while changing the scene.

### DIFF
--- a/addons/escoria-core/game/escoria.gd
+++ b/addons/escoria-core/game/escoria.gd
@@ -391,8 +391,9 @@ func _handle_direct_scene_run() -> void:
 
 
 # Used by game.gd to determine whether the game scene is ready to take inputs
-# from _input() function. To do so, the current_scene must be set, game scene 
-# must be set and game scene must've been notified that room is ready.
+# from _input() function. To do so, the current_scene must be set, the game
+# scene must be set, and the game scene must've been notified that room
+# is ready.
 #
 # *Returns*
 # true if game scene is ready for inputs
--- a/addons/escoria-core/game/escoria.gd
+++ b/addons/escoria-core/game/escoria.gd
@@ -391,8 +391,8 @@ func _handle_direct_scene_run() -> void:
 
 
 # Used by game.gd to determine whether the game scene is ready to take inputs
-# from _input() function. To do so, the current_scene must be set, the game
-# scene must be set, and the game scene must've been notified that room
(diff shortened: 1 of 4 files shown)


BHSDuncan:
Looks ok, but I'd still want to know what was introduced in Godot 3.4.4 that causes this.

@balloonpopper can you please regression test this against previous versions of Godot?

balloonpopper (approved):
The issue we were seeing was that moving the mouse in the Godot window while the game was starting would break the game with a tooltip error. This happened only in simplemouse and not in 9-verb.

The updated fix has been tested against 9 verb and simplemouse in 3.4.4, 3.3 and 3.2
Tested with starting a full game and loading directly into a scene.
Moving the mouse as the game starts doesn't break the game under this change.

addons/escoria-core/game/escoria.gd
```diff
@@ -381,3 +381,13 @@ func _handle_direct_scene_run() -> void:
 	if current_scene_root is ESCRoom:
 		escoria.object_manager.set_current_room(current_scene_root)
 
+
+# Used by game.gd to determine whether the game scene is ready to take inputs
+# from _input() function. To do so, the current_scene must be set, game scene 
+# must be set and game scene must've been notified that room is ready.
```
balloonpopper, line 387:
```suggestion
# from _input() function. To do so, the current_scene must be set, the game
# scene must be set, and the game scene must've been notified that room
# is ready.
```

addons/escoria-core/game/escoria.gd
```diff
@@ -383,8 +383,9 @@ func _handle_direct_scene_run() -> void:
 
 
 # Used by game.gd to determine whether the game scene is ready to take inputs
-# from _input() function. To do so, the current_scene must be set, game scene 
-# must be set and game scene must've been notified that room is ready.
+# from _input() function. To do so, the current_scene must be set, the game
+# scene must be set, and the game scene must've been notified that room
```
BHSDuncan, line 387:
```suggestion
# from the _input() function. To do so, the current_scene must be set, the game
# scene must be set, and the game scene must've been notified that the room
```

BHSDuncan (changes requested):

BHSDuncan (approved):

PR #584, opened 2022-04-21
fix: manage clearing of current_action when action is interrupted

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/222
Fixes https://github.com/godot-escoria/escoria-issues/issues/117
Fixes https://github.com/godot-escoria/escoria-issues/issues/248

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -281,9 +281,6 @@ func _on_event_finished(finished_statement: ESCStatement, return_code: int, chan
 	if event_flags & ESCEvent.FLAG_NO_SAVE:
 		escoria.save_manager.save_enabled = true
 
-	if return_code == ESCExecution.RC_CANCEL:
-		return_code = ESCExecution.RC_OK
-
 	_running_events[channel_name] = null
 	_channels_state[channel_name] = true
 
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -280,6 +280,11 @@ func _on_event_finished(finished_statement: ESCStatement, return_code: int, chan
 
 	if event_flags & ESCEvent.FLAG_NO_SAVE:
 		escoria.save_manager.save_enabled = true
+	
+	# If the return code was RC_CANCEL due to an event finishing with "stop" command for example
+	# we convert it to RC_OK so that other processed waiting for RC_OK can carry on.
+	if return_code == ESCExecution.RC_CANCEL:
+		return_code = ESCExecution.RC_OK
 
 	_running_events[channel_name] = null
 	_channels_state[channel_name] = true
--- a/addons/escoria-core/game/core-scripts/esc/commands/accept_input.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/accept_input.gd
@@ -59,3 +59,9 @@ func run(command_params: Array) -> int:
 
 	escoria.inputs_manager.input_mode = mode
 	return ESCExecution.RC_OK
+
+
+# Function called when the command is interrupted.
+func interrupt():
+	# Do nothing
+	pass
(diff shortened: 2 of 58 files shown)


addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -281,9 +281,6 @@ func _on_event_finished(finished_statement: ESCStatement, return_code: int, chan
 	if event_flags & ESCEvent.FLAG_NO_SAVE:
 		escoria.save_manager.save_enabled = true
 
-	if return_code == ESCExecution.RC_CANCEL:
```
BHSDuncan, line 284: @dploeger can you maybe tell us why this block was here originally? It seems to just swallow a cancelled execution for some reason.

BHSDuncan:
@dploeger can you maybe tell us why the block at L284 of the event manager was there originally? It seems to just swallow a cancelled execution for some reason.

BHSDuncan (approved):

dploeger:
> @dploeger can you maybe tell us why the block at L284 of the event manager was there originally? It seems to just swallow a cancelled execution for some reason.

I think so. The RC_CANCEL is returned from the stop command, which cancels the current event but not in a failed way. So for the event handling the stop command I returned a successful code. 

At least that was my thinking behind it.

balloonpopper (changes requested):
From my testing, switching scenes is currently broken with this patch.
Additionally,  the animation sometimes jumps (or the the walk cycle is reset?) when you interrupt the current action which would be good to correct.

balloonpopper:
I've tried lots of different things to break the issue-222 branch (including room changes, interruptions at different times, bringing in an inventory item and using "pick up" to interrupt it). I've found 2 issues.
1) Wait doesn't get interrupted as mentioned above
2) I brought the "worker" in from a different room and made him walk around. If I interrupt the player, the worker's "walk" command is not interrupted.  This might be a problem, or might be expected behavior as I don't think we've discussed this use-case.

StraToN:
@balloonpopper

> From my testing, switching scenes is currently broken with this patch.
> Additionally, the animation sometimes jumps (or the the walk cycle is reset?) when you interrupt the current action which would be good to correct.

 I may have fixed the issues with latest commit. Thanks for retesting!

balloonpopper (changes requested):
Marking this as "Request changes" temporarily until I find out if you want to deal with the below bugs as separate issues.

* The animation still breaks when you interrupt the walking (I think it resets the walking animation frame? Also seems to move Mark a few pixels to the right so the walking isn't smooth)
* As mentioned on https://github.com/godot-escoria/escoria-demo-game/pull/588
(I don't know if you want to look at the issue I mentioned on 588 as part of this fix?) If you interrupt Mark walking by right clicking picture 2, then while he's walking to picture 2, right click picture 1 - Mark will walk to picture 1 and do both picture 1 and 2's talk actions.

* Confirming 117 is now fixed.
* Confirming 248 is now fixed.

BHSDuncan (approved):
Nothing really jumps out at me as being wrong with the code. And also, I love the Command pattern for the reasons in this PR. :) Looks good! Good work!

StraToN:
@balloonpopper 

> Marking this as "Request changes" temporarily until I find out if you want to deal with the below bugs as separate issues.
> 
>     * The animation still breaks when you interrupt the walking (I think it resets the walking animation frame? Also seems to move Mark a few pixels to the right so the walking isn't smooth)

I think it's not related to this issue. I advise creating a separate issue for this one.

>     * As mentioned on #588
>       (I don't know if you want to look at the issue I mentioned on 588 as part of this fix?) If you interrupt Mark walking by right clicking picture 2, then while he's walking to picture 2, right click picture 1 - Mark will walk to picture 1 and do both picture 1 and 2's talk actions.

While this looks related to interruptions, that is also a different issue - or at least, it can be managed as a separate issue.

balloonpopper:
Raised https://github.com/godot-escoria/escoria-issues/issues/259
and
https://github.com/godot-escoria/escoria-issues/issues/258

balloonpopper (approved):

PR #585, opened 2022-04-24
fix: allows for transitions to be disabled during :init event

BHSDuncan:
Also does the following:

- Eliminates UI flicker during game startup
- Eliminates menu disappearing resulting in flicker of previous scene when transitioning to first scene during `:newgame`

These should be tested against saving and loading games when the system is back in working order.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -282,8 +282,8 @@ func _on_event_finished(finished_statement: ESCStatement, return_code: int, chan
 		escoria.save_manager.save_enabled = true
 
 	if return_code == ESCExecution.RC_CANCEL:
-		return_code = ESCExecution.RC_OK
-
+	  return_code = ESCExecution.RC_OK
+	
 	_running_events[channel_name] = null
 	_channels_state[channel_name] = true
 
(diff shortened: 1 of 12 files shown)


balloonpopper (approved):

balloonpopper:
Tested and it's working as expected for me across both the intro and the first couple of screens.

balloonpopper (commented):
@BHSDuncan - I now get into the intro cutscene and then intothe menu without a transition. 
I can't see a way now though to have a cutscene to the menu if I want one - is there a way?

BHSDuncan:
> @BHSDuncan - I now get into the intro cutscene and then intothe menu without a transition. I can't see a way now though to have a cutscene to the menu if I want one - is there a way?

I think one of the rooms (14?) demonstrates that you can have a transition before going into a menu. So, yes, I think it's possible. If you can't see a way, let's discuss on Discord when we're able.

Otherwise, I'd think that using a `transition` command before calling `show_menu` would work.

balloonpopper:
Yep, it's just wrapping the menu call in transitions.
```
transition curtain out
show_menu main
transition curtain in
```
Cheers!

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -282,8 +282,8 @@ func _on_event_finished(finished_statement: ESCStatement, return_code: int, chan
 		escoria.save_manager.save_enabled = true
 
 	if return_code == ESCExecution.RC_CANCEL:
-		return_code = ESCExecution.RC_OK
-
+	  return_code = ESCExecution.RC_OK
```
StraToN, line 285: 1 tab turned into 2 spaces. Please set it back to 1 tab :)

StraToN (changes requested):

StraToN:
Thanks for the explanation of the change in esc_transition!

BHSDuncan:
Fixes #198

StraToN (approved):

PR #586, opened 2022-04-24
fix: avoids crashing when selected_action is null

BHSDuncan:


balloonpopper:
https://github.com/godot-escoria/escoria-issues/issues/247

balloonpopper (approved):
I'm approving this as it fixes the bug and stops Escoria crashing (yay!)

@BHSDuncan as the left click works as expected if you click in the window when a different application is active, is it possible to make right click work correctly? My expected behavior would be that either neither button works (clicking either button on the window just makes it the active window), or clicking back on the window does the correct action for the left and right buttons respectively.
If it is possible, do you want to handle that as an enhancement for 4.1?

BHSDuncan:
> I'm approving this as it fixes the bug and stops Escoria crashing (yay!)
> 
> @BHSDuncan as the left click works as expected if you click in the window when a different application is active, is it possible to make right click work correctly? My expected behavior would be that either neither button works (clicking either button on the window just makes it the active window), or clicking back on the window does the correct action for the left and right buttons respectively. If it is possible, do you want to handle that as an enhancement for 4.1?

I'm not sure, to be honest. The difference is that right-clicking on an item requires a current action of some kind to already be present, and I can't remember how right-click works normally in terms of an action not already being selected (since it obviously doesn't crash there). I'm sure there's a straight-forward answer, but let's get @StraToN to weigh in.

StraToN:
I am not certain to understand the issue here as I am unable to reproduce it under Linux. I guess it's a specificity of Windows, maybe, but for now I don't get the exact problem.

9-verbs UI is supposed to mimic LucasArts' 9-verb UI (mostly): each element of the room has a default action defined (ESCItem.default_action), right-click is supposed to execute that action on the focused item.

StraToN (approved):

StraToN:
On a side note: this issue does not happen on Linux. 
This is because of the OS window manager. 

Windows won't sent any mouse input event to a window that not the active one.
Linux will.

PR #588, opened 2022-04-24
fix: prevents the same event from being queued multiple times in a row

BHSDuncan:
Fixes https://github.com/godot-escoria/escoria-issues/issues/245

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -192,6 +192,27 @@ func queue_event_from_esc(script_object: ESCScript, event: String,
 # #### Parameters
 # - event: Event to run
 func queue_event(event: ESCEvent) -> void:
+	# Don't queue the same event more than once in a row.
+	var last_event = _get_last_event_queued(CHANNEL_FRONT)
+
+	# Check the queue first to see if appending the event will result in
+	# consecutive occurrences of the event. If not, be sure to check if the same
+	# event is currently running.
+	if last_event != null and last_event.name == event.name:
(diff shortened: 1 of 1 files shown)


BHSDuncan:
Please test extensively. This PR should block multiple consecutive occurrences in an event queue and, if it's empty, also disallow the same event if it's currently already running.

This shouldn't affect ESC commands/events coming from ESC scripts.

Best test it out and make sure, though.

balloonpopper (changes requested):
I can't reproduce the multiple events being queued issue with this fix. 

I did find a bug that I assume is related - if it's a new bug we can raise it accordingly.
If I right click the right painting, and while the character is walking to the painting I right click the left painting, Mark changes direction and walks to the left painting, however the interaction is the right painting's interaction.  (I can also do this in reverse and get painting 1's response talking to painting 2)

image

BHSDuncan:
> I can't reproduce the multiple events being queued issue with this fix.
> 
> I did find a bug that I assume is related - if it's a new bug we can raise it accordingly. If I right click the right painting, and while the character is walking to the painting I right click the left painting, Mark changes direction and walks to the left painting, however the interaction is the right painting's interaction. (I can also do this in reverse and get painting 1's response talking to painting 2)
> 
> image

I don't think it would be related as this seems to be more related to interrupting. One way to be sure would be to see if this issue exists in `develop`. Could you try that and let us know? (I would but I need to look at a couple other things and I won't have time after to get to this.)

balloonpopper:
The multiple talk issue is now https://github.com/godot-escoria/escoria-issues/issues/258

balloonpopper (approved):

StraToN (approved):

PR #593, opened 2022-05-04
feature: Added ability to use Escoria variables with the custom command

balloonpopper:
It annoyed me that after the change to allow parameters to be passed to "custom" that I couldn't pass Escoria variables.
Fixes : https://github.com/godot-escoria/escoria-issues/issues/261

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/custom.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/custom.gd
@@ -81,6 +81,11 @@ func run(command_params: Array) -> int:
 	var object = escoria.object_manager.get_object(
 		command_params[0]
 	)
+	# Global variables can be substituted into the command arguments by wrapping the global
+	# name in braces.
+	for loop in command_params[3].size():
+		command_params[3][loop] = escoria.esc_compiler.replace_globals(command_params[3][loop])
+
 	object.node.get_node(command_params[1]).call(
 		command_params[2],
 		command_params[3]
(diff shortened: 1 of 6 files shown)


StraToN (approved):

addons/escoria-core/game/core-scripts/esc/commands/custom.gd
```diff
@@ -81,6 +81,11 @@ func run(command_params: Array) -> int:
 	var object = escoria.object_manager.get_object(
 		command_params[0]
 	)
+	# Global variables can be substituted into the command arguments by wrapping the global
+	# name in braces.
+	for loop in command_params[3].size():
+		command_params[3][loop] = escoria.logger.replace_globals(command_params[3][loop])
```
BHSDuncan, line 87: If we're going to be replacing globals for things other than log-related commands, then I'd like to see `replace_globals` extracted from `logger` and put into a more appropriate place, e.g. `ESCCompiler`. @StraToN would you agree with that location? Or would you prefer it put somewhere else?

BHSDuncan (changes requested):

addons/escoria-core/game/core-scripts/esc/commands/custom.gd
StraToN: I totally agree, `replace_globals` will move out of ESCLogger.

balloonpopper:
@StraToN @BHSDuncan I've moved the globals_replace function to esc_compiler.
Tested the print, say and custom functions

BHSDuncan (approved):

StraToN (approved):

StraToN:
Looks good to me. Thanks a lot!

PR #595, opened 2022-05-07
fix: prevents improper events/commands from being queued while changing scenes

BHSDuncan:
Fixes https://github.com/godot-escoria/escoria-issues/issues/249
Fixes https://github.com/godot-escoria/escoria-issues/issues/257
Fixes https://github.com/godot-escoria/escoria-issues/issues/161
Fixes https://github.com/godot-escoria/escoria-issues/issues/263

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -54,6 +54,14 @@ var _channels_state: Dictionary = {}
 # Whether we're currently waiting for an async event to complete, per channel
 var _yielding: Dictionary = {}
 
+# Whether we're currently changing the scene.
+var _changing_scene: bool = false setget set_changing_scene
+
+var _change_scene: ChangeSceneCommand
+
+func _init():
+	_change_scene = ChangeSceneCommand.new()
+
 
 # Make sure to stop when pausing the game
 func _ready():
@@ -70,7 +78,7 @@ func _process(delta: float) -> void:
 	for channel_name in events_queue.keys():
 		channel_yielding = _yielding.get(channel_name, false)
 
-		if events_queue[channel_name].size() == 0 or channel_yielding:
+		if events_queue[channel_name].size() == 0 or channel_yielding or _changing_scene:
 			continue
 		if is_channel_free(channel_name):
 			_channels_state[channel_name] = false
@@ -159,6 +167,9 @@ func _process(delta: float) -> void:
 func queue_event_from_esc(script_object: ESCScript, event: String,
 	channel: String, block: bool) -> int:
 
+	if _changing_scene:
+		return ESCExecution.RC_WONT_QUEUE
+
 	if channel == CHANNEL_FRONT:
 		escoria.event_manager.queue_event(script_object.events[event])
 	else:
@@ -191,7 +202,13 @@ func queue_event_from_esc(script_object: ESCScript, event: String,
 #
 # #### Parameters
 # - event: Event to run
-func queue_event(event: ESCEvent) -> void:
+func queue_event(event: ESCEvent, force: bool = false) -> void:
+	if _changing_scene and not force:
+		escoria.logger.info(
+			"Changing scenes. Won't queue event '%s'." % event.name
+		)
+		return
+	
 	# Don't queue the same event more than once in a row.
 	var last_event = _get_last_event_queued(CHANNEL_FRONT)
 
@@ -260,9 +277,12 @@ func queue_background_event(channel_name: String, event: ESCEvent) -> void:
 
 
 # Interrupt the events currently running.
-func interrupt_running_event():
+#
+# #### Parameters
+# - exceptions: an optional list of events which should be left running
+func interrupt_running_event(exceptions: PoolStringArray = []):
 	for channel_name in _running_events.keys():
-		if _running_events[channel_name] != null:
+		if _running_events[channel_name] != null and not channel_name in exceptions:
 			_running_events[channel_name].interrupt()
 			_channels_state[channel_name] = true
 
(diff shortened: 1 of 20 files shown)


addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -54,6 +54,14 @@ var _channels_state: Dictionary = {}
 # Whether we're currently waiting for an async event to complete, per channel
 var _yielding: Dictionary = {}
 
+# Whether we're currently changing the scene.
+var _changing_scene: bool = false setget set_changing_scene
+
+var _change_scene: ChangeSceneCommand
+
+func _init():
```
StraToN, line 62: Missing comment here

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -54,6 +54,14 @@ var _channels_state: Dictionary = {}
 # Whether we're currently waiting for an async event to complete, per channel
 var _yielding: Dictionary = {}
 
+# Whether we're currently changing the scene.
+var _changing_scene: bool = false setget set_changing_scene
+
+var _change_scene: ChangeSceneCommand
```
StraToN, line 60: Missing comment here

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -291,6 +326,21 @@ func get_running_event(name: String) -> ESCEvent:
...
+# #### Parameterse
+# - value: boolean value to set _changing_scene to
+func set_changing_scene(value: bool) -> void:
+	escoria.logger.trace("Setting _changing_scene to %s." % value)
+
+	_changing_scene = value
+	
+	# If we're changing scenes, interrupt any (other) running events and purge
+	# all event queues.
+	if value:
```
StraToN, line 340: This changes makes the test clearer:
```suggestion
	if _changing_scene:
```
Also possible to rename the `value` variable for `p_is_changing_scene` instead.

addons/escoria-core/game/core-scripts/esc/types/esc_base_command.gd
```diff
@@ -33,10 +33,15 @@ func validate(arguments: Array) -> bool:
...
 
 
 # Return the name of the command based on the script's filename
 func get_command_name() -> String:
 	return command_name_regex.search(get_script().get_path()).get_string(FILE_REGEX_GROUP)
+
+
+# Function called when the command is interrupted.
+func interrupt():
+	escoria.logger.info("Command %s did not override interrupt." % get_command_name())
```
StraToN, line 47: This could be a TRACE log, I think.
```suggestion
	escoria.logger.trace("Command %s did not override interrupt." % get_command_name())
```

StraToN (changes requested):

balloonpopper:
With the latest PR, clicking while transitioning causes issues. 
1) Start in room 3. Click to room 2. While the transition out is happening, right click somewhere on the right platform. When room 2 transitions in, Mark walks to the mouse position you clicked while in room 3.
2) If you start in room 3, click the 3->2 doorway. As the room change is happening, click door 3->4. He's now stuck. The transition doesn't run and you can't move. "Changing scenes. Won't queue event 'exit_scene'."
You can change the active verb (9verb) but clicking on the button in the room does nothing. Console log : "Changing scenes. Won't queue event 'look'."
3) In room 3, click the 3->4 door then click the 3->2 door while he's walking. Room 4 loads when room 2 is expected.

balloonpopper (changes requested):
See my comment for the issues I found

balloonpopper:
Observations from testing this PR.
1) I got this debug output when I interrupted a 3->2 transition. I'm not sure if the current behavior is that you should be able to interrupt change_scenes, but I didn't expect I should be able to.
`ESC 2022-05-10T23:03:03 (I)	Activated action exit_scene on [Node:37641] 	[]
ESC 2022-05-10T23:03:03 (D)	Queueing exit_scene in channel _front. 	
ESC 2022-05-10T23:03:03 (I)	Left click on background at  	[(219, 474)]
ESC 2022-05-10T23:03:03 (D)	Interrupting queued event exit_scene in channel _front... 	
ESC 2022-05-10T23:03:03 (I)	Interrupting event exit_scene ([Object:37654]) 	[]
 ESC 2022-05-10T23:03:03 (W)	Warnings in file change_scene
Interrupt() function not implemented
`
2) I can't force the wrong room to load anymore
3) It pauses before it starts transitioning out, and between transition out and in that it never used to. Particularly with "instant" transitions, the transitions aren't the most "instant" any more. 
To clarify from further testing. I generally play on TRACE (because it's ended up set like that - probably from some previous test - and I just don't think to change it :D ) - under TRACE there's more of a lag then there was (maybe we're logging more now with blocked interruptions?) The response time under "DEBUG" is certainly acceptable - so I won't fault you if you put this in the "will not fix" category.
4) Loading directly into room 1 (play scene) - if I click on the right hand side of the screen (walkable area) while curtain transition-in is running, it crashes.
image

image
5) I have done the room 1->2 transition multiple times and haven't managed to break it.
6) Pressing escape during 1->2  transition out (curtain) it crashes.
image

image

The bugs from yesterday all appear fixed.

BHSDuncan:
RE the above:

1. It's actually interrupting the `exit_scene` event which is internal, but really should have no effect. As long as the behaviour that happens is ok, I'd prefer to leave this for now. Otherwise, we can look at adding `exit_scene` to the exemption list, although it's saying it's interrupting something that's queued and not running, so I'd still think it should be ok.

2. Awesome!

3. Maybe it's just the added logging, as we haven't added (or taken away) anything that's time sensitive like `wait`. The code should bear that out. There's always been a `wait` command after transitioning out and in, but that's by design. I suppose it's a matter of opinion whether it should be there (it predates my joining, though).

4. This should now be fixed.

5. Awesome!

6. This should now be fixed. In fact, we've decided to move the initiating of the pause screen to the UI. I also fixed something (pre-existing) where you could pause during the intro and the main menu (fun fact: unpausing would cause the whole UI itself to show up; whoops). We'll leave this in place until we have a way to properly skip animations/cutscenes.

balloonpopper (approved):
I can't break the latest version - all the previously reported bugs are gone as far as I can find.

StraToN (approved):

PR #596, opened 2022-05-11
fix: handles translation keys that have only one word (i.e. no spaces…

BHSDuncan:
…) in the actual dialogue

Fixes https://github.com/godot-escoria/escoria-issues/issues/264
Fixes https://github.com/godot-escoria/escoria-issues/issues/267
Fixes https://github.com/godot-escoria/escoria-issues/issues/266

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/types/esc_command.gd
+++ b/addons/escoria-core/game/core-scripts/esc/types/esc_command.gd
@@ -46,7 +46,9 @@ func _init(command_string):
 						parameters.append(
 							parameter
 						)
-					elif ":" in parameter and '"' in parameter and not parameter.ends_with('"'):
+					elif ':"' in parameter and (parameter.ends_with(':"') or not parameter.ends_with('"')):
+						# The second clause in this helps to handle dialogue that starts with a space
+						# and also allowing single-word dialogue to be handled in a separate elif.
 						quote_open = true
 						parameter_values.append(parameter)
 					elif not quote_open and parameter.begins_with('"'):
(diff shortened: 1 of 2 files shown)


balloonpopper:
The text still breaks if it's preceded by a space - do you want to fix that under this PR?
i.e.
`say player ROOM1_look_wall_item_1:" Borked"`

balloonpopper (approved):

balloonpopper:
Tried all the combos I could think of. The only ones I had issues with are (and I appreciate these are outside the original scope of the ticket, and we might want to leave the crashing ones crashing as they're actually an error) below. 

* If the speech is all blank with an invalid label, you get talking animation but there's nothing to say. Warnings in log.
```
say player ROOM:"           "
```
* Blank text, valid label works fine
* Leading spaces work fine
* Trailing are fine
* Leading and trailing are fine
* If there is no text at all it crashes, with OR without a valid label, for both avatar/non-avatar
```
say player ROOM1_look_wall_item_2:""
say player ROOM1_smells:""
```
* Missing label but a colon crashes.
```
say player  :"tweet"
```
* Assigning the text (valid label and text) to an ESClocation crashes as it doesn't have a text colour assigned. The sound plays though. **Non avatar mode**.
```
say r1_start ROOM1_look_wall_item_2:"I REALLY don't know what that stuff is."
```
* Assigning the text (valid label and text) to an ESClocation in **avatar mode** has no error message. The sound plays, and the text is displayed, BUT the avatar isn't shown in the text box.
```
say r1_start ROOM1_look_wall_item_2:"I REALLY don't know what that stuff is." avatar
```
* A label without a colon crashes with/without avatar
```
say player ROOM1_look_wall_item_2"a"
```

StraToN (approved):

PR #601, opened 2022-06-18
WIP : Player helper2

balloonpopper:
Needs code cleanup but ready for testing


StraToN:
Some feedback from my test:

== On export ==
- The AnimatedSprite node has a default animation pre-selected. Is it possible to remove that one and select the first custom animation instead?
- The generated Resource seems to have put all animations under the Directions field (walk animations, but also idles and speaks)

StraToN (changes requested):

balloonpopper:
I've just pushed a commit for the character creator. The changes : 
* Removed commented code found by @StraToN 
* Documented the instance variables in each .gd file
* Added tooltips
* Added a plugin description
* Also in case I didn't mention it, I fixed earlier the 2 issues @StraToN found (idle/talk animations are saved correctly, and the default animation is now "idle_down")
* I've also added a progress bar on export, but either due to @Dev Orion's refactor, or because of 🤷 , the export is over before the progress bar has time to display. Export isn't instant (maybe 0.2 seconds) but it seems to take the progress bar window the same amount of time to popup... My thought is to leave the code in so if exports are ever slow, there's a progress bar even though it looks like it won't appear.
* I've added a popup window that suggests some things the player might want to do with the new scene once the export completes.

PR #606, opened 2022-07-11
Remove some trace log happening too often

StraToN:
@balloonpopper already told me about that one. It is a normal trace log, but happening too often and slows down execution.


balloonpopper (approved):
Tested and approved. With the number of these that appear (and no reference to what triggered them) these errors don't currently look to provide much benefit from my perspective. I agree with @BHSDuncan in relation to removing them and re-adding them if necessary.

PR #605, opened 2022-07-11
Fixed non-interactive objects still being interactive

dploeger:
(as discussed on Discord)

related to https://github.com/godot-escoria/escoria-issues/issues/104

PR #609, opened 2022-07-14
Fix when ESCPlayer is not selectable or interactive manage click through

StraToN:
Supersedes https://github.com/godot-escoria/escoria-demo-game/pull/605
Fixes https://github.com/godot-escoria/escoria-demo-game/issues/1071

This change verifies the content of the hover stack before performing the action.

Adds a new test room 18 that allows testing superposed ESCItems.

If the clicked item is an unselectable ESCPlayer then we consider the action went through it and applies to:
- the item behind (if any in the hover stack)
- on the background if no item is behind

To test:
1) ESCPlayer superposed to ESCItem
- set mark as NOT selectable (uncheck in the inspector)
- run room 1
- hover mark should do nothing
- left/right/double click through mark
- put mark in front of and item on the wall and do the same (left/right/double click on the item through Mark)

2) Superposed ESCItems
- run room18

### DIFF
--- a/addons/escoria-core/game/esc_inputs_manager.gd
+++ b/addons/escoria-core/game/esc_inputs_manager.gd
@@ -296,7 +296,7 @@ func _on_mouse_exited_inventory_item() -> void:
 func _on_mouse_entered_item(item: ESCItem) -> void:
 	var object: ESCObject = escoria.object_manager.get_object(item.global_id)
 	if object and not object.interactive:
-		_hover_stack_erase_item(item)
+		hover_stack_erase_item(item)
 		hotspot_focused = ""
 		escoria.main.current_scene.game.element_unfocused()
 		return
@@ -331,7 +331,7 @@ func _on_mouse_entered_item(item: ESCItem) -> void:
 func _on_mouse_exited_item(item: ESCItem) -> void:
 	var object: ESCObject = escoria.object_manager.get_object(item.global_id)
 	if object and not object.interactive:
-		_hover_stack_erase_item(item)
+		hover_stack_erase_item(item)
 		hotspot_focused = ""
 		escoria.main.current_scene.game.element_unfocused()
 		return
@@ -344,7 +344,7 @@ func _on_mouse_exited_item(item: ESCItem) -> void:
 		self,
 		"Item unfocused: %s" % item.global_id
 	)
-	_hover_stack_erase_item(item)
+	hover_stack_erase_item(item)
 	if hover_stack.empty():
 		hotspot_focused = ""
 		escoria.main.current_scene.game.element_unfocused()
@@ -500,5 +500,5 @@ func _clean_hover_stack():
 #
 # #### Parameters
 # - item: the item to remove from the hover stack
-func _hover_stack_erase_item(item):
+func hover_stack_erase_item(item):
 	hover_stack.erase(item)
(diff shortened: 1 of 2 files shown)


StraToN:
Special shoutout to @dploeger : i'd like to check if this fix works for your suitcase scene.

addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -318,6 +329,17 @@ func _on_mouse_entered_item(item: ESCItem) -> void:
 #
 # - item: The Escoria item hovered
 func _on_mouse_exited_item(item: ESCItem) -> void:
+	if escoria.object_manager.get_object(item.global_id) and \
+			not escoria.object_manager.get_object(item.global_id).interactive:
+		_hover_stack_erase_item(item)
+		hotspot_focused = ""
+		escoria.main.current_scene.game.element_unfocused()
+		return
+	var object: ESCObject = escoria.object_manager.get_object(item.global_id)
```
BHSDuncan, line 338: Ditto here.

BHSDuncan (approved):

balloonpopper:
@StraToN I found 2 issues with this. 
Issue 1
1) Turn "selectable" off
2) Hover over a picture in room 1
3) Click the picture. The tooltip will say "Artwork" or "Artwork 2". Don't move the mouse
4) Mark will walk to the picture
5) When mark's body covers the mouse, the tooltip disappears - you need to move off the artwork then back on again to register the artwork.

Issue 2
1) Make mark walk to a picture in room 1
2) Move the mouse away then back over the picture (and over Mark). The tooltip should say "look artwork"
2) Right click. The "look artwork" will not happen, instead Mark will walk closer to the artwork.

StraToN:
@balloonpopper 

> Issue 2
>     1. Make mark walk to a picture in room 1
>     2. Move the mouse away then back over the picture (and over Mark). The tooltip should say "look artwork"
>     3. Right click. The "look artwork" will not happen, instead Mark will walk closer to the artwork.

Thank you for this report. It is fixed in new commit.

> Issue 1
>     1. Turn "selectable" off
>     2. Hover over a picture in room 1
>     3. Click the picture. The tooltip will say "Artwork" or "Artwork 2". Don't move the mouse
>     4. Mark will walk to the picture
>     5. When mark's body covers the mouse, the tooltip disappears - you need to move off the artwork then back on again to register the artwork.

Not a bug. This was the behavior before this PR (I just checked this to ensure).
This relates to how the action is managed in the currently enabled game.gd (so in 9-verbs or simple-mouse). If this requires a change, I propose to submit a new issue and/or discuss it on the Discord maybe.

balloonpopper:
The right-click through the character works, but the behaviour of room 18 is not quite what I'd expect - it still registers items that you've moved away depending on where you move the mouse. I've approved it on the basis that the person bit is working but I'll post the room 18 issue below. if it's a separate issue we can tackle it later.

1 is the upper item, 2 is the middle, 3 is the bottom.

Correct behavior that currently happens : Move over the upper item only (i.e the mouse is over 1 and not over 2 or 3) and "use" it, item 1 moves out of the way and the tooltip clears as it should. 

Issue
* As above
* Once item has finished its animation, move over item 2.
* Tooltip says "upper item" but should say "middle item"
* Tooltip will only register middle item correctly after you move completely off items 1,2 and 3 then back on to item 2.

Issue 2)
* Restart room 18
* Use upper item
* Esc to bring up the menu
* Position your mouse over the top left hand corner of the "continue" button and click "continue" (this is a spot that is over item 2 but not over item 3"
* Note that the tooltip has disappeared
* If you play around with moving the mouse and clicking continue, or leaving the mouse where it is and clicking continue you'll see that sometimes the tooltip is still active as it should be, and sometimes it isn't

balloonpopper (approved):

dploeger:
> Special shoutout to @dploeger : i'd like to check if this fix works for your suitcase scene.

Sorry for the late reply. I just updated escoria core and it works like a charm. Thanks!

PR #610, opened 2022-07-16
fix: changes 'scenery' node from Node to Node2D so it will respect th…

BHSDuncan:
…e visibility of the root (i.e. the room).

fixes https://github.com/godot-escoria/escoria-issues/issues/287

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_animation_player.gd
+++ b/addons/escoria-core/game/core-scripts/esc_animation_player.gd
@@ -95,13 +95,19 @@ func play(name: String, backwards: bool = false):
 		_animation_player.seek(0)
 	elif not _is_animation_player:
 		_animated_sprite.frame = 0
+
 	_current_animation = name
+
 	if backwards and _is_animation_player:
 		_animation_player.play_backwards(name)
 	elif backwards:
 		_animated_sprite.play(name, true)
 	else:
 		_player_node.play(name)
+
+		# Instead of waiting for the next frame, start the animation now.
+		if _is_animation_player:
+			_player_node.advance(0)
 
 
 # Play the given animation backwards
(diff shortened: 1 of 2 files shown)


balloonpopper:
@BHSDuncan - This wasn't obvious to me (so I'm expecting it won't be obvious to new users). Rather than correcting just this scene, is this issue possible to detect and alert on? (I'm not sure why a Node2D is necessary)

BHSDuncan:
Node2D has a visibility property that Node does not. Because the visibility property propagates down, if a parent node doesn’t have it (ie it’s of type Node), it won’t be propagated to its children. 

I’d suggest this should be mentioned in the docs. Alerting on it won’t work, I don’t think, as there may be parts of the scene that could legitimately be a Node, and we can’t glean context.

balloonpopper:
That makes sense. I've added it to the (long outstanding) documentation ticket to be added to the documentation at some point.

I notice though that the door appears a frame after the rest of the room is drawn. As always, I'm in trace so it's more noticeable. The door is "closed" via an animation with immediate enabled ("set_state r6_r_exit close_door") in ":setup" so I would expect it to complete before you see the room (even though transition is immediate). If you don't agree then I can fix that issue by just setting the default visibility on the door to visible, but if there's a timing issue here then it would be good to fix.

BHSDuncan:
I have it on TRACE as well but I don't notice that frame at all...It looks to be in its proper state when the room loads. I don't think there's a timing issue here, but if there is, I'll need to be able to reproduce it.

Given the above, I'd suggest that we just keep it in the back of our minds and get this PR merged!

StraToN (approved):

BHSDuncan:
RE the "room popping" issue above, I can recreate it (barely) if I start room 6 directly, but if, say, I go from room 5->6, I can't see it at all; however, @balloonpopper has sent a video of it happening even with 5->6.

I'll spend a few more days looking into this, but if I can't find a satisfactory solution, this issue is small enough that it can be made into its own and pushed until after the 4.0 release.

PR #611, opened 2022-07-16
fix: puts a default avatar path back as it had gone missing for some …

BHSDuncan:
…reason

fixes https://github.com/godot-escoria/escoria-issues/issues/286


balloonpopper:
@BHSDuncan I would suggest that as part of this ticket, we should add an error message that picks up when you try to use "say" with an avatar and the avatar can't be loaded.

PR #615, opened 2022-07-22
feat: validates updated camera position against current camera limits…

BHSDuncan:
…; also some messaging updates

Fixes https://github.com/godot-escoria/escoria-issues/issues/127

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_push.gd
@@ -54,6 +54,24 @@ func validate(arguments: Array):
 					% [get_command_name(), arguments[0]]
 		)
 		return false
+
+	var target_pos = (escoria.object_manager.get_object(arguments[0]).node as ESCItem).global_position
+	var camera: ESCCamera = escoria.object_manager.get_object(escoria.object_manager.CAMERA).node as ESCCamera
+	var camera_limit: Rect2 = Rect2(camera.limit_left, camera.limit_top, camera.limit_right - camera.limit_left, camera.limit_bottom - camera.limit_top)
+
+	if not camera_limit.has_point(target_pos):
+		escoria.logger.warn(
+			self,
+			"[%s]: invalid camera position. Camera cannot be moved to %s at position %s as this is outside the current camera limit %s."
+				% [
+					get_command_name(),
+					arguments[0],
+					target_pos,
+					camera_limit
+				]
+		)
+		return false
+
 	if not arguments[2] in SUPPORTED_TRANSITIONS:
 		escoria.logger.error(
 			self,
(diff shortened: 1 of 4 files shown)


balloonpopper:
* Camera_push to a location outside the limits : it moves until it hits the limit then stops.  What are your thoughts on adding a warning saying "couldn't reach point x as it's outside the limits?"

* Shift has a bug - the error function is expecting two arguments instead of 1.
```
	if not camera_limit.has_point(new_pos):
		escoria.logger.error(
			"[%s]: invalid camera position. Camera cannot be moved by %s to %s as this is outside the current camera limit %s."
				% [
					get_command_name(),
					shift_by,
					new_pos,
					camera_limit
				]
		)
```

* Set_pos worked correctly except for one thing I'm not sure of. With an x limit of 2260 pixels, shifting to 2259 is fine, and 4000 breaks as expected, but 2260 also breaks. Should it be up to and including the limit, or limit -1 as it currently is?

BHSDuncan:
Will fix the shift bug. Must've gotten through after the rebase.

@StraToN what do you think about the other two questions above?

StraToN:
> Camera_push to a location outside the limits : it moves until it hits the limit then stops. What are your thoughts on adding a warning saying "couldn't reach point x as it's outside the limits?"

I agree on this proposal. That sound like an important information to provide to gamedev.

> Set_pos worked correctly except for one thing I'm not sure of. With an x limit of 2260 pixels, shifting to 2259 is fine, and 4000 breaks as expected, but 2260 also breaks. Should it be up to and including the limit, or limit -1 as it currently is?

That *should* be the case: not breaking on 2260, as per Godot's Camera2D documentation.
Can you check if we do have that kind of limitation in our code?

balloonpopper:
@StraToN I can't spot anywhere we're doing a -1 on the coordinates to cause the behavior.
What looks to be happening is 
* we determine the screen limits from the size of the ESC Background's texture. 
* If the texture is 10 pixels wide and 10 high, that means the pixels it covers are from (0,0) to (9,9) 
As a result of the above, this means that if we think we have a 10x10 background and try to set the camera to (10,10) it fails as the limit is effectively (9,9)
To me, though it doesn't make intuitive sense when looking at the camera limits command, this makes sense as if you actually did set the camera to (10,10) you'd be looking at blackness, so you'd want the camera to stop at (9,9).

Unless anyone disagrees (or thinks I'm misreading the code), I think this is therefore fixed by documentation on the camera limits/movement commands.

balloonpopper (approved):
Tested and works well!

balloonpopper:
Latest code works correctly for trying to move past camera limits, and correctly works up to and including camera limits.

balloonpopper (approved):

StraToN (approved):

StraToN:
Thanks a lot !

PR #617, opened 2022-07-28
feat: adds Escoria wizard

balloonpopper:
Features:

Character creator
* Creates a character from one or more spritesheets for 4 or 8 direction characters
* Requires creation of walk, speak, idle animations to avoid errors in Escoria when playing a game and an animation  is missing
* A collision box is created automatically based on largest animation frame
* A dialog ESC Location node created automatically (currently 20%  of the character's height above their sprite)

Room creator
* Rooms will be automatically created in the folder specified in "Room Selection Room Dir" Escoria setting, but can be modified from the interface. (If having the wizard using and change this setting is problematic as it will break the initial intent of this configuration setting, I'll externalise the setting)
* A folder will be created for the room called [room name], with "scripts" and "objects" subfolders.
* An ESCCharacter will be linked to the ESCRoom if selected
* An ESCBackground will be created. A background sprite will be configured if selected
* If the option to use a room ESC script is chosen, a template file will be copied into the scripts folder with the name selected
* An ESCTerrain will be created with an empty Navigation Polygon 2D child
* An empty Node2D "RoomObjects" node will be created for non-inventory items

Item creator
* Allows for both inventory and room object items to be created. They only differ in where the resulting objects are created.
* Inventory items will be created in the configured inventory item folder (which can be modified in the interface)
* Room object items will be created as a child of the currently selected node in the current scene
* An ESCLocation will be created (2 x sprite size below the image) for an interaction location
* A rectangular collision shape will be created, matching the sprite size

I was planning on having a "new game" helper, but I'll leave that for v2.

Thoughts/feedback/bug fixes welcome.


balloonpopper:
Export of characters is currently buggy. Also need to retest items create correctly.

StraToN (approved):

BHSDuncan (approved):

PR #618, opened 2022-07-29
Issue 165 - no indication when destination is unreachable

balloonpopper:
I'm not sure if this is the right way to solve this problem with scheduling events etc, but it looked right to me and worked for my testing.

To test, take an item (eg. the left hand painting in room 1) and move its related ESCLocation out of the navigation polygon.

Standard behavior is a warning debug message.
If you add a "cant_reach" event to the object's script, this will fire. e.g.
```
:cant_reach
say player "Sorry, I can't reach the painting."
```
Fixes https://github.com/godot-escoria/escoria-issues/issues/165

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
@@ -540,6 +540,20 @@ func perform_inputevent_on_object(
 
 		if not player_global_pos.is_equal_approx(destination_position):
 			dont_interact = true
+			escoria.logger.info(
+				self,
+				"Player could not reach destination coordinates %s. "  % str(destination_position) \
+					+ "Any requested action for %s will not fire." % obj.global_id
+			)
+			if escoria.event_manager.EVENT_CANT_REACH in obj.events:
+				escoria.event_manager.queue_event(obj.events[escoria.event_manager.EVENT_CANT_REACH])
+			else:
+				escoria.logger.info(
+					self,
+					"%s event not found for object %s so nothing to do." % \
+						[escoria.event_manager.EVENT_CANT_REACH, obj.global_id]
+				)
+
 
 	# If no interaction should happen after player has arrived, leave
 	# immediately.
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -31,6 +31,7 @@ const EVENT_ROOM_SELECTOR = "room_selector"
 const EVENT_SETUP = "setup"
 const EVENT_TRANSITION_IN = "transition_in"
 const EVENT_TRANSITION_OUT = "transition_out"
+const EVENT_CANT_REACH = "cant_reach"
 
 
 # Event channel names


addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
```diff
@@ -540,6 +540,19 @@ func perform_inputevent_on_object(
 
 		if not player_global_pos.is_equal_approx(destination_position):
 			dont_interact = true
+			escoria.logger.warn(
```
BHSDuncan, line 543: I'd argue this should be `info` and not `warn` as nothing is "potentially going wrong" in the system itself, just the game.

addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
```diff
@@ -540,6 +540,19 @@ func perform_inputevent_on_object(
...
+				self,
+				"Player could not reach destination coordinates %s. "  % str(destination_position) \
+					+ "Any requested action for %s will not fire." % obj.global_id
+			)
+			if "cantreach" in obj.events:
```
BHSDuncan, line 548: Please add `cantreach` to the list of consts in the event manager class itself and reference it that way.

BHSDuncan (changes requested):
Just a couple small things, but to my eyes, this actually looks like a good solution. Granted, I probably haven't thought through the myriad cases that could come up, but for starters, this looks to address the issue at hand.

addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
balloonpopper: I don't feel particularly strongly here so don't mind changing it. The reason I'd suggest it's a warning is that I've answered questions from new users who can't figure out why their interaction isn't working. They're more likely to see warning's than info messages. 
If you're happy that they'll go straight for debug when trying to work out a problem then I'm happy to make it info :-)

balloonpopper: Made it info as discussed in discord.

BHSDuncan (approved):

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -31,6 +31,7 @@ const EVENT_ROOM_SELECTOR = "room_selector"
 const EVENT_SETUP = "setup"
 const EVENT_TRANSITION_IN = "transition_in"
 const EVENT_TRANSITION_OUT = "transition_out"
+const EVENT_CANT_REACH = "cant_reach"
```
StraToN, line 34: So according to this const, the event to use is `:cant_reach`, not `:cantreach` as you explain in the PR description?

balloonpopper: @StraToN Sorry, I changed it from "cantreach" to "cant_reach" during testing to make it more readable. I've updated the issue's description to correct this.

StraToN (approved):

StraToN:
Thanks!
This addition will also require a documentation change.

Issue #176, opened 2018-05-25
[Feature Request] Ability to slide items


mjtorn:
In [Iron Sky: A Lunar Adventure](https://gamejolt.com/games/ironsky-lunar-adventure/339236) I have an NPC that gets moved to a disjoint segment of the terrain. I did not have time to implement `slide` and `slide_block` so I tried my best to coordinate animation and teleportation.

---

It could be that an item without separate `directions` animations and using `walk` could lead to the same
effect, but that's not enough.

Say the player walks in front of a big turbine, but the project dosn't have the budget to animate a flailing character, but simply slides him back. Or my case, where an NPC (not a simple item!) experiences low gravity and floats.

The simplest solution would be to `set_state` for the desired animation or sprite and `slide` the object.

This is also future-proof when we finally get npc.gd done.

It would have to remember the scale/z-index of exiting the terrain, if that happens.

fleskesvor:
Couldn't this be solved by having a trigger area in front of the turbine and using the `cut_scene` command to play an animation on the player character, which would have tracks for both the flailing and the changing of position? The last key frame of the flailing track would set the player sprite to one of the idle sprites. The flailing could be optional.

I don't remember if it's possible to have an animation without an initial key frame though. If not, this would result in the player being repositioned to that position at the start of the cut_scene. Alternatively, if we could implement a more general purpose `tween` command for more ad-hoc cutscenes. I'm not sure if there's a good way to pre-configure tweens without knowing their start value up front though.

According to the documentation, `Tween` is "useful for animations with unknown start and end points" (quoted from the 2.1 docs) though, so maybe I'll just have to read up on it:

http://docs.godotengine.org/en/3.0/classes/class_tween.html

I think tween was requested once before too, but I'm not able to find where at the moment.

mjtorn:
As I said above, "doesn't have the budget [...] simply slides him back". We have this case where an NPC is floating upward but we don't want the player to be able to walk there, so the character is `teleport`ed.

Although Spine isn't officially supported, I'm sad to say I don't have the time to put the effort into developing a solution that's very strictly tied to not-Spine :( If a simple esc command won't happen, I'd put the effort into making custom esc commands and implement for my own use :(

fleskesvor:
Yes, but couldn't the low budget version just be `cut_scene` with an animation that just moves and rotates the player node? That should only take a few minutes to implement in an `AnimationPlayer` and can already be triggered with a simple esc command.

[...]

PR #621, opened 2022-08-08
fix: "Give xyz with" should say "give xyz to" (#176)

ArturM:
Added ACTION_PREPS dictionary to settings/escoria/ui
Modified `esc_tooltip.gd` to make use of this dictionary.

Fixes https://github.com/godot-escoria/escoria-issues/issues/176

### DIFF
--- a/addons/escoria-ui-9verbs/tooltip/tooltip_action_target.gd
+++ b/addons/escoria-ui-9verbs/tooltip/tooltip_action_target.gd
@@ -1,5 +1,7 @@
 extends ESCTooltip
 
+export var prepositions = {"use": "with", "give": "to"}
+
 func update_tooltip_text():
 	bbcode_text = "[center]"
 	bbcode_text += "[color=#" + color.to_html(false) + "]"
@@ -8,6 +10,7 @@ func update_tooltip_text():
 	bbcode_text += current_target
 
 	if waiting_for_target2 and current_target2.empty():
+ 		current_prep = prepositions.get(current_action, current_prep)		
 		bbcode_text += "\t" + current_prep
 
 	if !current_target2.empty():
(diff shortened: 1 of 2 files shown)


BHSDuncan (changes requested):
The core code should have zero knowledge of the implementation. This issue is strictly related to the UI, and all UIs in Escoria are to be provided by the game developer using Escoria.

If you look in the 9-verbs UI we provide, you'll see that it's meant to have code that extends `ESCTooltip` and overrides `update_tooltip_text()`. This is likely where any manipulation of the tooltip text should go, since `ESCTooltip` itself prints nothing and has no language (in fact all it does is manage targets and display, really).

I would suggest the following:

- Remove any code from `escoria-core`.
- Update the 9-verbs UI as mentioned above.
- Even though this is meant to be an example of a UI, I'd still like to make sure we do things "properly". Specifically, I'm not sure if the prepositions belong in the project settings, especially if we ever want to localize those prepositions.

A good first attempt, but let's see about making the changes above.

Happy to discuss if needed.

BHSDuncan:
Meant to fix #176.

balloonpopper:
Can "preps" be renamed to "prepositions" too please?

StraToN (changes requested):
I agree with @BHSDuncan's comment : your change should happen in the update_tooltip_text() function instead.
You can either make the prepositions dictionary a Project Setting related to 9-verbs UI plugin, or you can also set it as an exported variable of 9-verbs's tooltip so you can change it directly in the inspector.

ArturM:
> I agree with @BHSDuncan's comment : your change should happen in the update_tooltip_text() function instead. You can either make the prepositions dictionary a Project Setting related to 9-verbs UI plugin, or you can also set it as an exported variable of 9-verbs's tooltip so you can change it directly in the inspector.

Reverted my changes in core. Made exported variable in `tooltip_action_changes.gd` and changed `update_tooltip_text()` to use this dictionary.

StraToN (approved):

StraToN:
LGTM. Thanks!

BHSDuncan:
Changes look ok to me, too, but before I accept them I'd like @balloonpopper to test this and get his sign-off before we merge. Thanks!

BHSDuncan (approved):

PR #623, opened 2022-08-11
Adds a small fix on walking interruption to avoid freed reference

StraToN:


BHSDuncan (approved):

balloonpopper (approved):
Wasn't quite sure how to get an invalid instance to test this, so I just did a quick general test with both the player walking and the NPC in room 14 and it didn't break anything. The code looks fine.

PR #626, opened 2022-08-24
fix: disallows the use of specified strings in the name argument for th…

BHSDuncan:
…ese commands

Fixes https://github.com/godot-escoria/escoria-issues/issues/154


balloonpopper (approved):
Tested all 3 commands which work as expected when an "/" is included.

StraToN (approved):

StraToN:
Thanks for this change. It is far better than my previous version.

PR #627, opened 2022-08-24
fix: guards against scenes with no player set or loaded

BHSDuncan:
Fixes https://github.com/godot-escoria/escoria-issues/issues/110

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_game.gd
+++ b/addons/escoria-core/game/core-scripts/esc_game.gd
@@ -162,11 +162,17 @@ func do_walk(destination, params: Array = [], can_interrupt: bool = false) -> vo
 #
 # - position: Position clicked
 func left_click_on_bg(position: Vector2) -> void:
-	do_walk(
-		position,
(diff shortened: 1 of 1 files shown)


balloonpopper:
Tested this by loading room 13 which has no player scene configured and debug set to TRACE. I'm not seeing the messages I'm expecting to see based on the source code changes. Am I testing it incorrectly @BHSDuncan ?

BHSDuncan:
Actually, fair point: I'm not 100% sure how this can be tested unless we have a UI that doesn't extend `ESCGame`. The "fix" was already in for the 9-verbs UI; however, what I did applies to the base `ESCGame` and, as such, is unlikely to be hit.

StraToN (approved):

StraToN:
Ok on the principle.
To fix @balloonpopper, the only way is to explicitely call the overriden methods using:
```
.left_click_on_bg(position)
```
But this has to be done by gamedev.

balloonpopper (approved):

PR #628, opened 2022-08-27
fix: clears tooltip after any action has finished regardless of success

BHSDuncan:
Fixes https://github.com/godot-escoria/escoria-issues/issues/293

### DIFF
--- a/addons/escoria-ui-9verbs/game.gd
+++ b/addons/escoria-ui-9verbs/game.gd
@@ -114,14 +114,18 @@ func element_focused(element_id: String) -> void:
 		ESCActionManager.ACTION_INPUT_STATE.COMPLETED:
 			return
 
-		ESCActionManager.ACTION_INPUT_STATE.AWAITING_VERB_OR_ITEM, \
-		ESCActionManager.ACTION_INPUT_STATE.AWAITING_ITEM:
+		ESCActionManager.ACTION_INPUT_STATE.AWAITING_VERB_OR_ITEM:
 			tooltip.set_target(target_obj.tooltip_name)
 
 			# Hovering an ESCItem highlights its default action
 			if escoria.action_manager.current_action != VERB_USE \
 					and target_obj is ESCItem:
 				verbs_menu.set_by_name(target_obj.default_action)
+
+		ESCActionManager.ACTION_INPUT_STATE.AWAITING_ITEM:
+			tooltip.set_target(target_obj.tooltip_name)
+
+			verbs_menu.set_by_name(escoria.action_manager.current_action)
 
 		ESCActionManager.ACTION_INPUT_STATE.AWAITING_TARGET_ITEM:
 			tooltip.set_target2(target_obj.tooltip_name)


BHSDuncan:
This will need a good bit of testing and a close review of the actual code. The changes made should now clear the current action no matter what an action's return result happens to be. And the character will still walk to the location of the item despite the invalid action, but that is to be addressed in another issue.

balloonpopper:
I think this behavior isn't quite right. If there is a valid tooltip I think it should remain after the current action completes. e.g. if I have a look action selected on the wrench and I click it to be told "it's a wrench", then without moving the mouse, "Look at wrench" should still be the active tooltip, and I should be able to click to run the "look at wrench" action again.

If I move Mark to stand in front of the wrench, and left click look then left click the wrench (wrench is the active tooltip), I get told "it's me Mark-io".

If I pick up the pen, then "use pen with pipe" he walks in front of the pipe and doesn't perform an action (as expected). If I then move the mouse away, then highlight the pipe again and right click (which should "look at pipe") - nothing happens. (I can left click "look" then left click the "pipe" to make the look work (without having moved), but right clicking should also work). If I walk away then right click the pipe, the look action works correctly.

BHSDuncan:
> I think this behavior isn't quite right. If there is a valid tooltip I think it should remain after the current action completes. e.g. if I have a look action selected on the wrench and I click it to be told "it's a wrench", then without moving the mouse, "Look at wrench" should still be the active tooltip, and I should be able to click to run the "look at wrench" action again.

I disagree. In MI2, if you click "Look at" and then an object, it does the action, but the action clears (even if you don't move the mouse), although it does show the object in the tooltip again, which we don't currently do. So this would seem to be a matter of preference in terms of what we should do.

> If I move Mark to stand in front of the wrench, and left click look then left click the wrench (wrench is the active tooltip), I get told "it's me Mark-io".

@StraToN this sounds like a (separate) issue/bug, does it not?

> If I pick up the pen, then "use pen with pipe" he walks in front of the pipe and doesn't perform an action (as expected). If I then move the mouse away, then highlight the pipe again and right click (which should "look at pipe") - nothing happens. (I can left click "look" then left click the "pipe" to make the look work (without having moved), but right clicking should also work). If I walk away then right click the pipe, the look action works correctly.

This sounds related to the issue above, i.e. the player covering an object. @StraToN ?

StraToN (approved):

StraToN:
Ok to me. Thanks!

balloonpopper:
Issue - if you load room 5 directly you can "pick up" the wrench and this works correctly (walks over, picks it up). Instead "look at" the wrench first. Once he's walked over and looked at it, try and pick it up. Trying to get the mouse in the right position (where it's over the wrench but not obscured by Mark) is difficult to get first time as it's only about 1px wide, and if you go over Mark the tooltip resets to "look at".

balloonpopper:
Looks good!

balloonpopper (approved):

StraToN:
Thanks! 🥳

PR #629, opened 2022-08-27
feat: allows for default/fallback actions to be specified

BHSDuncan:
Closes https://github.com/godot-escoria/escoria-issues/issues/9


addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
```diff
@@ -230,27 +230,36 @@ func clear_current_tool():
...
 	action: String,
 	target: ESCObject,
 	combine_with: ESCObject = null
-) -> int:
+) -> ESCEvent:
+
 	escoria.logger.info(
 		self,
-		"Activated action %s on %s." % [action, target]
+		"Checking action %s on %s..." % [action, target]
```
balloonpopper, line 254:
```suggestion
		"Checking if action '%s on %s' has been defined..." % [action, target]
```

balloonpopper (changes requested):

balloonpopper:
11:07 AM] Tootle: Under normal behavior in room 6, the default action on the left door is "use".  Based on some testing I just did (removing the "use" default from the file), it appears that "using" the door doesn't actually do anything (a stuff up on my behalf from when I coded the room behavior). So that's the default  behavior - use left door = nothing happens.

Now I put the default use action back in your file. 
:use | TK
say player "I can't use this!"

Run room 6 and make him walk to the right side of the ship. Left click the left doorway, then right click it. He'll walk towards the doorway, saying the line and ignoring the TK parameter. Once he gets to the door, even though the default "use" action has obviously fired, the previously actioned left click fires and changes you to room 5.
[11:07 AM] Tootle: i.e default actions don't interrupt running ones.
[11:09 AM] Dev Orion: Hmm that’s interesting. @StraToN what do you think? I’m not sure we’ve really defined what should happen in this case.
[11:09 AM] Dev Orion: I think what’s being said is ok, but maybe it shouldn’t be?
[11:11 AM] Tootle: In the above case at least, I would expect that the minute a new event is fired, it takes precedence and cancels anything running.
[11:14 AM] Dev Orion: Default actions aren’t currently setup like that. They’re meant to be used if a current action can’t be found.
[11:15 AM] Tootle: I still think it's valid. If I was in room 5 for example, I've changed my mind about walking to the door, and now want to try and "use" the pipe on the wall. I don't want the game to decide I still want to leave the room.
[11:15 AM] Dev Orion: We should def get @StraToN ‘s input in terms of how interrupting in this case should work. ESP w how it used to be.

StraToN (changes requested):

addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
```diff
@@ -362,33 +339,45 @@ func _activate(
...
+					"Event for action %s on object %s not found." % [
+						action,
+						target.global_id
+					]
+			)
+
+	return event_to_return
+
+
+func _run_event(event: ESCEvent) -> int:
```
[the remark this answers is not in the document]
StraToN: Great job overall, I just have one tiny request regarding a missing comment. Thanks a lot, this is amazing.
---
Concerning above interrogation @balloonpopper: as I said on Discord, no matter the nature of the latest action requested (default or not), it should always interrupt any running action.

StraToN:
Great job overall, I just have one tiny request regarding a missing comment. Thanks a lot, this is amazing.

---
Concerning above interrogation @balloonpopper: as I said on Discord, no matter the nature of the latest action requested (default or not), it should always interrupt any running action.

StraToN (approved):

StraToN (approved):

balloonpopper:
It's working well. Awesome work!!!
2 thoughts on improvements. These can go to new features if you don't want feature creep (though they would be nice to have :-) )
* Items you have to have in your possession to use - I think we need an event that we can react to so we could do something like "I need to be holding that". Currently it's like you didn't click anything as there's no default response, and no way to define one.
* I think it would be nice to have the say command default to the "CURRENT_PLAYER" unless otherwise specified. This would make writing conversations quicker - particularly as "CURRENT_PLAYER" is long and will be repeated everywhere in every conversation.

One bug I've found
Run room 5 (f6). 
Click "use" then "wrench". The command is ignored as expected.
Right click the wrench. Wait for the 'look' text to complete. 
Click "use" then "wrench" - Mark now says the default "use" text. If you make him walk anywhere then try it again, the use command is ignored.

BHSDuncan:
> It's working well. Awesome work!!! 2 thoughts on improvements. These can go to new features if you don't want feature creep (though they would be nice to have :-) )
> 
> * Items you have to have in your possession to use - I think we need an event that we can react to so we could do something like "I need to be holding that". Currently it's like you didn't click anything as there's no default response, and no way to define one.

That comes back to the question of ease of use: We could easily put in a signal, but it'll require more GDScript on behalf of the gamedev to use. We still intend to have a discussion around making it so that GDScript can hook back into ESC script to allow gamedevs to do such things as above. So, this would be a good feature, but it's definitely post-4.0.

> * I think it would be nice to have the say command default to the "CURRENT_PLAYER" unless otherwise specified. This would make writing conversations quicker - particularly as "CURRENT_PLAYER" is long and will be repeated everywhere in every conversation.

Default arguments for ESC commands right now are a pain unless they appear at the end of the arg list, much like a lot of other languages. Plus, we don't have named parameters either, so we can't easily circumvent this. A larger feature for another day. :)

> 
> One bug I've found Run room 5 (f6). Click "use" then "wrench". The command is ignored as expected. Right click the wrench. Wait for the 'look' text to complete. Click "use" then "wrench" - Mark now says the default "use" text. If you make him walk anywhere then try it again, the use command is ignored.

Actually, it shouldn't ignore the command in the first case. There's a default action for `use`, so this definitely seems like a bug. I'll look into it.

BHSDuncan:
It looks like there's an issue with overlapping items, in this case the player and the wrench.

I looks like both get the focus, but the order in which they come in can vary, with the most-recent item with the focus keeping it. And then, if you move the mouse outside of the wrench's collision shape but that's still inside the player's, and then move the mouse back over top of the wrench (but still inside the player's collision shape), the wrench comes into focus and the tooltip is updated.

To make matters worse, even with the wrench focused, the `target` being sent to the object manager can still be the player.

Looking through the inputs manager, the `hover_stack` seems to be reliant on the "back of the queue" for matching the clicked on item. Given the above, I think this needs to be addressed @StraToN

StraToN:
> It looks like there's an issue with overlapping items, in this case the player and the wrench.
> 
> I looks like both get the focus, but the order in which they come in can vary, with the most-recent item with the focus keeping it. And then, if you move the mouse outside of the wrench's collision shape but that's still inside the player's, and then move the mouse back over top of the wrench (but still inside the player's collision shape), the wrench comes into focus and the tooltip is updated.
> 
> To make matters worse, even with the wrench focused, the `target` being sent to the object manager can still be the player.
> 
> Looking through the inputs manager, the `hover_stack` seems to be reliant on the "back of the queue" for matching the clicked on item. Given the above, I think this needs to be addressed @StraToN

This has to do with both `ESCInputsManager._on_mouse_entered_item()` and `ESCInputsManager._on_mouse_left_clicked_item()` functions. The only function that modifies the content of the hover_stack is the former one, and this happens when the mouse cursor hovers a new item. 
I am not exactly certain what should be done here, apart from ensuring that the label is always correctly set: when player and wrench are overlapping, focusing the wrench does show the "wrench" label, but the click applies on the player. In this case, I'd say that the click happening on player is "normal" because it IS the top-level item here, and the label should have been "Me".

BHSDuncan:
But the problem here is that the order in which the items are loaded into `hover_stack` aren't consistent, i.e. it could be `person -> wrench` or `wrench -> person`. Which one it applies to seems to be dependent on this ordering.

balloonpopper:
1) Minor gripe - missing space in debug output
ESC (2022-09-08T09:06:49) W esc_action_manager.gd: Invalid action on **itemTrying** to run action use on object r5_wrench, but item must be in inventory.

2) I edit the "mark" scene and turn off "selectable". I then run room 5 directly and stand in front of the wrench. The wrench won't come up in a tooltip. In my test there was 1 pixel on the right hand side where Mark's collision volume didn't obstruct the wrench's collision volume and that still worked as expected, but Mark's collision volume swallows the input and doesn't let the wrench be activated.
According to the debug, sometimes the player and sometimes the wrench is the last focussed item, but the wrench tooltip never appears.

addons/escoria-core/game/core-scripts/esc/commands/say.gd
```diff
@@ -98,8 +105,12 @@ func run(command_params: Array) -> int:
 	# Replace the names of any globals in "{ }" with their value
 	command_params[1] = escoria.globals_manager.replace_globals(command_params[1])
 
+	var player_character_global_id = escoria.main.current_scene.player.global_id \
+		if command_params[0].to_upper() == CURRENT_PLAYER_KEYWORD \
+		else command_params[0]
+
 	escoria.dialog_player.say(
-		command_params[0],
+		player_character_global_id,
```
StraToN, line 113:
```suggestion
		speaking_character_global_id ,
```

addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -346,6 +348,18 @@ func _on_mouse_exited_item(item: ESCItem) -> void:
 # - item: The Escoria item clicked
 # - event: The input event from the click
 func _on_mouse_left_clicked_item(item: ESCItem, event: InputEvent) -> void:
+	if item as ESCPlayer and not (item as ESCPlayer).selectable:
```
StraToN, line 351: Shouldn't this block happen after test `if input_mode == INPUT_ALL:` that happens below?

addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -380,6 +394,18 @@ func _on_mouse_left_double_clicked_item(
 	item: ESCItem,
 	event: InputEvent
 ) -> void:
+	if item as ESCPlayer and not (item as ESCPlayer).selectable:
```
StraToN, line 397: Same as above:
Shouldn't this block happen after test if input_mode == INPUT_ALL: that happens below?

addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -410,6 +436,18 @@ func _on_mouse_left_double_clicked_item(
 # - item: The Escoria item clicked
 # - event: The input event from the click
 func _on_mouse_right_clicked_item(item: ESCItem, event: InputEvent) -> void:
+	if item as ESCPlayer and not (item as ESCPlayer).selectable:
```
StraToN, line 439: Same as above:
Shouldn't this block happen after test if input_mode == INPUT_ALL: that happens below?

StraToN (changes requested):
Excellent changes overall. I just have a few remarks to be discussed or fixed. Thanks a lot!

addons/escoria-core/game/esc_inputs_manager.gd
BHSDuncan: Yeah, that would probably save some checks. Can't remember why I put the block there but will change.

addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -347,6 +349,18 @@ func _on_mouse_exited_item(item: ESCItem) -> void:
 # - event: The input event from the click
 func _on_mouse_left_clicked_item(item: ESCItem, event: InputEvent) -> void:
 	if input_mode == INPUT_ALL:
+		if item as ESCPlayer and not (item as ESCPlayer).selectable:
+				escoria.logger.debug(
```
StraToN, line 353: Indentation issue here : 2 tabs instead of 1

addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -381,6 +395,18 @@ func _on_mouse_left_double_clicked_item(
 	event: InputEvent
 ) -> void:
 	if input_mode == INPUT_ALL:
+		if item as ESCPlayer and not (item as ESCPlayer).selectable:
+				escoria.logger.debug(
```
StraToN, line 399: Indentation issue as well

addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -411,6 +437,18 @@ func _on_mouse_left_double_clicked_item(
 # - event: The input event from the click
 func _on_mouse_right_clicked_item(item: ESCItem, event: InputEvent) -> void:
 	if input_mode == INPUT_ALL:
+		if item as ESCPlayer and not (item as ESCPlayer).selectable:
+				escoria.logger.debug(
```
StraToN, line 441: Indentation issue

StraToN (changes requested):

StraToN (approved):

balloonpopper (approved):

PR #632, opened 2022-09-10
fix: fixes inconsistent return type from queue_event_from_esc

balloonpopper:
Expands on the queue_event demo in room 14 by adding a second worker. It shows what happens when two events are queued and one contains a blocking command.
It also corrects issues with the return code from the queue_event command when a queue and blocking parameter are provided (thanks @BHSDuncan!)


StraToN (approved):

StraToN:
Looks good to me. Thanks!

BHSDuncan (approved):

PR #633, opened 2022-09-15
fix: Support for both int and real for set_angle command

balloonpopper:
Support for both real and int to be accepted for the angle for the player to face, and for the amount of time the character will face each intervening direction as they turn. Currently only ints are allowed for direction, and real for wait time.

Questions for discussion
1) Duration is expected to be a real, but I used an int. This is easily fixed by just changing the validate function to 
[TYPE_STRING,  TYPE_INT,  [TYPE_REAL, TYPE_INT]]. 

The second parmeter is the angle you want the player to face. I've tested and this also works as a real, i.e.
[TYPE_STRING,  [TYPE_REAL, TYPE_INT],  [TYPE_REAL, TYPE_INT]],

So can anyone see a reason the developer shouldn't be allowed to specify an angle as a real? I can't see anyone ever needing it, but at least it wouldn't break things if they accidently used a real
2) The function definition says
> - *degrees*: Number of degrees by which `object` is to be turned

Testing shows that this is incorrect. The behavior is actually that the angle specified is the angle the character will end up facing. So which is correct?
3) If the final angle is 180 degrees away from where you want it to be, or you want the character to turn the long way, this isn't possible. Should we have a way of specifying clockwise/anti-clockwise rotation?

4) To see which frames of the animation would play during the turn, I added some additional frames to the idle directions before testing. I noticed some unexpected behavior with this script while testing if the set_angle was blocking
    say player "It's a leaking pipe."
    set_angle player 270 2.0
    say player "Nooooooooooo."
    set_global r5_dialog_advance 1

He says the first line, then starts the turn. As set_angle isn't blocking, he immediately says the second line, turns to 270 for a split seconds, then ends up facing back upwards again - something to do with the 2 second timing.
So should this command be blocking? A blocking and non-blocking version might be a good idea, but I'm not sure how we treat the behavior in my example
5) From my (4) example, it appears the first 2 second pause is facing the direction he's already facing (upwards). Does this seem right. It probably won't matter in 99% of cases as I'd expect developers to set the duration to 0.1 or something so the character can be seen to turn rather than immediately facing the new direction. Larger numbers don't serve much purpose (that doesn't mean we shouldn't support them though as you might want a game character to do a very slow, dramatic turn).

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/set_angle.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/set_angle.gd
@@ -22,7 +22,7 @@ class_name SetAngleCommand
 func configure() -> ESCCommandArgumentDescriptor:
 	return ESCCommandArgumentDescriptor.new(
 		2,
-		[TYPE_STRING, TYPE_INT, TYPE_REAL],
+		[TYPE_STRING,  [TYPE_REAL, TYPE_INT],  [TYPE_REAL, TYPE_INT]],
 		[null, null, 0.0]
 	)
 


StraToN:
@balloonpopper 

> 1) So can anyone see a reason the developer shouldn't be allowed to specify an angle as a real? I can't see anyone ever needing it, but at least it wouldn't break things if they accidently used a real.

If I am not wrong, the things was that you'd want actually set your player/npc to *turn* so that he faces a direction that is defined in the [ESCAnimationResource](https://github.com/godot-escoria/escoria-demo-game/blob/develop/addons/escoria-core/game/core-scripts/resources/esc_animationresource.gd) that is attached to it. Such direction is actually defined by [2 elements](https://github.com/godot-escoria/escoria-demo-game/blob/develop/addons/escoria-core/game/core-scripts/resources/esc_directionangle.gd) : a starting angle and a size. These values vary in function of the actual number of directions (usually 4 and 8, but could also be 2 for example).

Of course, ESC already featured a command to turn the player/npc/item to face a direction, that was the `set_angle` command. I had 2 solutions for its parameters:
- Use cardinal directions or directions:
```
# Facing up
set_angle player NORTH 0
# Also facing up
set_angle player UP 0 
```
But I felt this was a bit too arbitrary because it would assume that directions always respect cardinal directions (N, NE, E, SE, S, SW, W, NW). What if someone makes a game using isometric representation (eg: Sanitorium)? Or custom?
- Use a target angle value and determine the target animation by checking which angle interval the provided value lies. This is the chosen solution. This means that all commands below have the same result (provided the player has an UP direction defined for [315°, 45°]):
```
set_angle player 0 0 # 0° is in [315°, 45°] => UP
set_angle player 316 0 # 316° is in [315°, 45°] => UP
set_angle player 32 0 # 32° is in interval [315°, 45°] => UP
```

So, why only ints? Well, since the actual angle value is not as important as the interval it is included in, I thought it would be easier to allow only int values. Technically, real values are not a problem since the test is still possible (assuming all tested values are converted to real) but I don't think this is necessary.

> 2) The function definition says
>- degrees: Number of degrees by which object is to be turned
>
>Testing shows that this is incorrect. The behavior is actually that the angle specified is the angle the character will end up facing.  So which is correct?

As per my explanation above, you understand that the definition is not exact. This should actually be the "target angle" that the player/npc/item should be turned TO. 

> 3) If the final angle is 180 degrees away from where you want it to be, or you want the character to turn the long way, this isn't possible. Should we have a way of specifying clockwise/anti-clockwise rotation?

We probably had discussed this with @dploeger back when I coded that behavior. 
There are actually 2 things:
- this mechanism of choosing the shortest way was indeed done on purpose, I remember it very well. I did it that was because it is automatically used when the player/npc/item walks around with user input clicks. In that cas, shortest way has to be used, otherwise the movement would not feel coherent.
- the `set_angle` ESC command : of course being able to define whether turn is perfored (anti-)clockwise could be useful, I suppose we just decided that this would be added later when/if necessary.

> So should this command be blocking? A blocking and non-blocking version might be a good idea, but I'm not sure how we treat the behavior in my example

I am not certain of the reason why this command was not duplicated with a blocking version: maybe you can to achieve the desired result using `wait` command after the `set_angle` call? If so, that definitely would be the reason. 
However, the `wait` argument seems to have a bug, as per your description...

> 5) From my (4) example, it appears the first 2 second pause is facing the direction he's already facing (upwards). Does this seem right. It probably won't matter in 99% of cases as I'd expect developers to set the duration to 0.1 or something so the character can be seen to turn rather than immediately facing the new direction. Larger numbers don't serve much purpose (that doesn't mean we shouldn't support them though as you might want a game character to do a very slow, dramatic turn).

Ah, no this wasn't the desired behavior, you're right... The `wait` duration should happen after the first "new" animation. That is, if player is currently facing RIGHT, and you want to face DOWN LEFT (SOUTH WEST), then (assuming you have 8 directions defined):
```
# target face down right (240° should do)
set_angle player 240 2.0
```
Then player should 
1. face DOWN_RIGHT (SOUTH EAST)
2. wait 2.0s
3. face DOWN (SOUTH)
4. wait 2.0s
5. turn DOWN LEFT (SOUTH WEST)
6. and return immediately as the target direction is reached, no more waiting.

I hope this long comment clears your interrogations.

StraToN (approved):

BHSDuncan (approved):

PR #635, opened 2022-10-28
fix: ensures a failed command/event doesn't leave the game in a state…

BHSDuncan:
… where it isn't accepting inputs.

Fixes https://github.com/godot-escoria/escoria-issues/issues/314


StraToN (approved):

balloonpopper (approved):
Great work as always!

BHSDuncan:
As pointed out, there's an issue with the error message, namely that `final_statement` looks to actually be an event, and we can see that "events", "statements", and "commands" are often used interchangeably which shouldn't be done.

@StraToN we need to discuss this to discuss what should be the desired nomenclature as well as purpose as the "event_finished" handler in the event manager.

PR #637, opened 2022-10-31
feat: includes most-currently run statement when signalling a stateme…

BHSDuncan:
…nt's/event's completion to aid in debugging messages

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -410,11 +410,8 @@ func _on_event_finished(finished_event: ESCStatement, finished_statement: ESCSta
 	if return_code == ESCExecution.RC_CANCEL:
 		return_code = ESCExecution.RC_OK
 	elif return_code == ESCExecution.RC_ERROR:
-		escoria.logger.warn(
-			self,
-			"Statement '%s' returned an error in event '%s'. Resetting input mode to 'ALL'."
-				% [finished_statement.name, event.name]
-		)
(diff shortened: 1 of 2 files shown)


balloonpopper (approved):

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -372,16 +372,18 @@ func set_changing_scene(p_is_changing_scene: bool) -> void:
...
 #
 # #### Parameters
-# - finished_statement: statement object that finished
+# - finished_event: statement object representing the event that finished
+# - finished_statement: statement object representing the "deepest" statement (most likely a command)
+#   that just completed; this is useful for interrupted or failed statements especially
 # - return_code: Return code of the finished event
 # - channel_name: Name of the channel that the event came from
-func _on_event_finished(finished_statement: ESCStatement, return_code: int, channel_name: String) -> void:
+func _on_event_finished(finished_event: ESCStatement, finished_statement: ESCStatement, return_code: int, channel_name: String) -> void:
```
StraToN, line 380: Why isn't `finished_event` of type `ESCEvent` ?

BHSDuncan: Because that's how `ESCStatement` works: When it's finished its `run()` method, it signals into the event manager using `self` as the `finished_event` argument. I don't think `ESCStatement` has a reference to an event or even shares the same structure.

This is partly what I'm talking about when I say "statement", "event", and "command" are confusing.

Another example: `ESCCommand` and `ESCBaseCommand` serve entirely different purposes.

StraToN (approved):

PR #638, opened 2022-11-01
fix: Stops error message when disabling simple dialog plugin

balloonpopper:
You'll see this error if disabling the simple dialog plugin.
```
 core/project_settings.cpp:961 - Condition "!p_info.has("type")" is true.
```
For every time an empty dictionary is passed to clear a setting from Godot's project settings in simple dialog

```
	ESCProjectSettingsManager.register_setting(
		ESCProjectSettingsManager.MAX_TIME_TO_DISAPPEAR,
		null,
		{}
	)
```

### DIFF
--- a/addons/escoria-core/game/esc_project_settings_manager.gd
+++ b/addons/escoria-core/game/esc_project_settings_manager.gd
@@ -84,9 +84,6 @@ const FULLSCREEN = "%s/%s/%s/fullscreen" % [DISPLAY, WINDOW, SIZE]
 # - default_value: Default value
 # - info: Property info for the setting
 static func register_setting(name: String, default_value, info: Dictionary) -> void:
-	if ProjectSettings.has_setting(name):
-		push_error("Cannot add project setting %s - it already exists." % name)
-		return
 	if default_value == null:
 		push_error("Default_value cannot be null. Use remove_setting function to remove settings.")
 		assert(false)
--- a/addons/escoria-core/game/esc_project_settings_manager.gd
+++ b/addons/escoria-core/game/esc_project_settings_manager.gd
@@ -81,15 +81,42 @@ const FULLSCREEN = "%s/%s/%s/fullscreen" % [DISPLAY, WINDOW, SIZE]
 # #### Parameters
 #
 # - name: Name of the project setting
-# - default: Default value
+# - default_value: Default value
 # - info: Property info for the setting
-static func register_setting(name: String, default, info: Dictionary) -> void:
+static func register_setting(name: String, default_value, info: Dictionary) -> void:
+	if ProjectSettings.has_setting(name):
+		push_error("Cannot add project setting %s - it already exists." % name)
+		return
+	if default_value == null:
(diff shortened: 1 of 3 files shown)


addons/escoria-core/game/esc_project_settings_manager.gd
```diff
@@ -81,15 +81,30 @@ const FULLSCREEN = "%s/%s/%s/fullscreen" % [DISPLAY, WINDOW, SIZE]
...
-static func register_setting(name: String, default, info: Dictionary) -> void:
+static func register_setting(name: String, default_value, info: Dictionary) -> void:
 	ProjectSettings.set_setting(
 		name,
-		default
+		default_value
 	)
-	info.name = name
-	ProjectSettings.add_property_info(info)
+	if ! default_value == null:
```
BHSDuncan, line 91:
```suggestion
	if default_value != null:
```

addons/escoria-core/game/esc_project_settings_manager.gd
```diff
@@ -81,15 +81,30 @@ const FULLSCREEN = "%s/%s/%s/fullscreen" % [DISPLAY, WINDOW, SIZE]
...
+		info.name = name
+		
+		# Project settings require a "type" to be set
+		if ! "type" in info:
```
BHSDuncan, line 95:
```suggestion
		if not "type" in info:
```

BHSDuncan (changes requested):
Just a couple small changes.

Please also update the doc string to show that the method can also be used to unregister/remove a project setting by sending in a `default_value` of null. Thanks!

addons/escoria-core/game/esc_project_settings_manager.gd
balloonpopper: Fixed

addons/escoria-core/game/esc_project_settings_manager.gd
balloonpopper: Fixed

balloonpopper:
PR updated
The project_settings_manager function for creating Godot settings has been split into explicit set_setting and remove_setting functions as using set_setting to unset a setting was a code smell.
Error checking was added to the functions which then highlighted that the DEFAULT_DIALOG_TYPE was being set outside the simple_dialog plugin.

@StraToN my removal of the code that sets a default value for the setting from escoria-core/plugin.gd might cause issues with the code that is expecting to find the setting (in say.gd and esc_dialog_player.gd). If you don't have a dialog plugin active however, things aren't going to work anyway, so it seemed right to me to not be configuring the simple dialog plugin from plugin.gd. Alternatively I put it back in plugin.gd and remove it being created from simple dialog?

balloonpopper:
Damn. Realised the set_setting function needs to be able to change settings as well as create them, so the does-it-already-exist check is wrong (#showerthoughts). I'll remove it when I'm on the computer tomorrow.
The question as to whether we should set the default dialogue type from plugin.gd is still valid though.

PR #640, opened 2022-11-07
fix: Removing dialog settings from core

balloonpopper:
This PR is to separate dialog settings from core to make the dialog plugin standalone.
This also fixes the error messages seen when disabling and enabling core relating to info.type missing in the call in esc_project_settings_manager

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/say.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/say.gd
@@ -78,13 +78,6 @@ func validate(arguments: Array):
 		)
 		return false
 
-	if ESCProjectSettingsManager.get_setting(ESCProjectSettingsManager.DEFAULT_DIALOG_TYPE) == "" \
-			and arguments[2] == "":
-		escoria.logger.error(
-			self,
-			"[%s]: Project setting '%s' is not set. Please set a default dialog type."
-					% [get_command_name(), ESCProjectSettingsManager.DEFAULT_DIALOG_TYPE]
-		)
 	return true
 
 
--- a/addons/escoria-dialog-simple/plugin.gd
+++ b/addons/escoria-dialog-simple/plugin.gd
@@ -3,7 +3,14 @@ tool
 extends EditorPlugin
 
 const MANAGER_CLASS="res://addons/escoria-dialog-simple/esc_dialog_simple.gd"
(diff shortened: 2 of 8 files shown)


balloonpopper:
@StraToN As discussed in Slack, this PR needs your input please. We need a way for plugins outside core to store settings that players might want to interact with. e.g. The dialogue plugin might create a "text speed" setting that the developer might want to expose in the game's settings menu. The suggestion was that it would interact with the save system to store these settings. (Settings can't be stored using the "custom settings" functionality as the custom settings dictionary is not available at the time the plugin is initialised).

addons/escoria-core/game/core-scripts/esc/commands/say.gd
```diff
@@ -78,13 +78,13 @@ func validate(arguments: Array):
...
-			"[%s]: Project setting '%s' is not set. Please set a default dialog type."
-					% [get_command_name(), ESCProjectSettingsManager.DEFAULT_DIALOG_TYPE]
-		)
+#	if ESCProjectSettingsManager.get_setting(ESCProjectSettingsManager.DEFAULT_DIALOG_TYPE) == "" \
+#			and arguments[2] == "":
+#		escoria.logger.error(
+#			self,
+#			"[%s]: Project setting '%s' is not set. Please set a default dialog type."
+#					% [get_command_name(), ESCProjectSettingsManager.DEFAULT_DIALOG_TYPE]
+#		)
```
balloonpopper, line 87:
```suggestion
```

addons/escoria-dialog-simple/plugin.gd
```diff
@@ -13,41 +19,35 @@ func get_plugin_name():
...
-		ESCProjectSettingsManager.MAX_TIME_TO_DISAPPEAR,
-		null,
-		{}
+	ESCProjectSettingsManager.remove_setting(
+		MAX_TIME_TO_DISAPPEAR
 	)
-
+	
+#	if escoria.settings_manager.custom_settings.has(SKIP_DIALOGS):
+#		escoria.settings_manager.custom_settings.erase(SKIP_DIALOGS)
```
balloonpopper, line 43:
```suggestion
```

StraToN (approved):

BHSDuncan (approved):

PR #641, opened 2022-11-08
Fix: unwanted error messages when opening a room in editor

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/290


BHSDuncan (approved):

balloonpopper (changes requested):
I tested the code and it worked fine. I would like to request a change though (if it's not feasible/too hard/you don't agree etc then I'm happy to approve the PR as-is.)

This code :
```				if Engine.is_editor_hint():
					logger.warn(
						self,
						"Multiple NavigationPolygonInstances enabled " + \
						"at the same time."```

Checks for this error condition (2+ polygons enabled) only when the room is loaded in. When you enable the 2nd polygon in the Godot UI, no error is triggered. Is it possible to have this condition turn the ESCTerrain node's icon in the scene tree into a warning ("!") flag ideally, or at least print a warning message to the console every time you enable a 2nd+ navigation polygon?

StraToN:
> [copied code]
> Checks for this error condition (2+ polygons enabled) only when the room is loaded in. When you enable the 2nd polygon in the Godot UI, no error is triggered. Is it possible to have this condition turn the ESCTerrain node's icon in the scene tree into a warning ("!") flag ideally, or at least print a warning message to the console every time you enable a 2nd+ navigation polygon?

You are right, I didn't explain in this PR why I didn't do what you're mentioning as it was very late. Doing so now.

Actually I did try since Godot allows setting that warining icon in the editor quite easily. Unfortunately, this doesn't seem to work because `ESCTerrain` extends `Navigation2D` which will be deprecated in Godot 4.0 (and maybe in Godot 3.6?). Therefore that node *already* shows a warning, and Godot seems to refuse adding my custom warning next to it.

The solution is "simply" to move from Navigation2D to Navigation2DServer and remove use of `get_simple_path()` as advised on [the docs page](https://docs.godotengine.org/en/stable/classes/class_navigation2d.html?highlight=Navigation2D).
This will require a few changes in the path calculation, I don't expect it to be **too** hard, but I lacked examples to use this correctly.

Once this is done, it will be easy to enable that warning icon in the editor.

Until then, I'll check and print this warning every time you enable a 2nd+ navigation polygon as you mention.

BHSDuncan (approved):

balloonpopper (approved):
Works well in my testing.

PR #642, opened 2022-11-08
fix: Turn interrupt warning to debug in change_scene.gd

balloonpopper:
Changes warning :
W change_scene.gd: [change_scene] interrupt() function not implemented
to debug as discussed in https://discordapp.com/channels/884336424780984330/884346430314143745/1039283919255781456


BHSDuncan:
Should we maybe see if there are other commands with a similar warning message that can also be switched to `debug`?

PR #643, opened 2022-11-08
fix: interrupt not implemented messages changed to debug from warn

balloonpopper:
Updated version of previous PR to include changing all messages for "interrupt() function not implemented." from warning to debug level.

I'm not sure if the message in addons/escoria-core/game/core-scripts/esc/types/esc_base_command.gd should be a stronger warning level (error maybe?)


BHSDuncan:
I think `debug` even in the base command is just fine.

StraToN (approved):

StraToN:
Thanks!

PR #645, opened 2022-11-08
feat: adds blocking versions of camera commands that can tween/transition

BHSDuncan:
Closes https://github.com/godot-escoria/escoria-issues/issues/138

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_block.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_block.gd
@@ -7,6 +7,9 @@
 # would result in no change. The zoom will happen over the given time period.
 # Blocks until the command completes.
 #
+# Zoom operations might not be as smooth as desired if the requested zoom
+# level results in an edge of the camera meeting any defined camera limits.
+#
 # **Parameters**
 #
 # - *magnitude*: Magnitude of zoom
@@ -55,7 +58,10 @@ func run(command_params: Array) -> int:
 
 	if command_params[1] > 0.0:
 		yield(_camera_tween, "tween_completed")
-
+	escoria.logger.debug(
+			self,
+			"camera_set_zoom_block tween complete."
+		)
 	return ESCExecution.RC_OK
 
 
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_block.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_block.gd
@@ -1,0 +1,67 @@
+# `camera_set_zoom_block magnitude [time]`
+#
+# Zooms the camera in/out to the desired `magnitude`. Values larger than '1' zoom
+# the camera out while smaller values zoom in. These values are relative to the
+# default zoom value of '1', not the current value. As such, while using a value
+# of '0.5' would double the size of the graphics, running the same command again
+# would result in no change. The zoom will happen over the given time period.
+# Blocks until the command completes.
+#
+# **Parameters**
+#
+# - *magnitude*: Magnitude of zoom
+# - *time*: Number of seconds the transition should take, with a value of `0`
+#   meaning the zoom should happen instantly (default: `0`)
+#
+# For more details see: https://docs.escoria-framework.org/camera
+#
+# @ESC
+extends ESCCameraBaseCommand
+class_name CameraSetZoomBlockCommand
+
+
+var _camera_tween: Tween
+
+
+# Return the descriptor of the arguments of this command
+func configure() -> ESCCommandArgumentDescriptor:
+	return ESCCommandArgumentDescriptor.new(
+		1,
+		[[TYPE_REAL, TYPE_INT], [TYPE_REAL, TYPE_INT]],
+		[null, 0.0]
+	)
+
+
+# Validate whether the given arguments match the command descriptor
+func validate(arguments: Array):
+	if not .validate(arguments):
+		return false
+
+	var camera: ESCCamera = escoria.object_manager.get_object(escoria.object_manager.CAMERA).node as ESCCamera
+	_camera_tween = camera.get_tween()
+
+	return true
+
+
+# Run the command
+func run(command_params: Array) -> int:
+	var camera: ESCCamera = escoria.object_manager.get_object(escoria.object_manager.CAMERA).node as ESCCamera
+
+	camera\
+		.set_camera_zoom(
+			command_params[0],
+			command_params[1]
+		)
+
+	if command_params[1] > 0.0:
+		yield(_camera_tween, "tween_completed")
+
+	return ESCExecution.RC_OK
+
+
+# Function called when the command is interrupted.
+func interrupt():
+	escoria.logger.debug(
+		self,
+		"[%s] interrupt() function not implemented." % get_command_name()
+	)
--- a/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_height_block.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_height_block.gd
@@ -1,19 +1,19 @@
-# `camera_set_pos_block time x y`
+# `camera_set_zoom_height_block pixels [time]`
 #
-# Moves the camera to the given absolute position over a time period. Blocks
-# until the command completes.
+# Zooms the camera in/out so it occupies the given height in pixels.
+# Blocks until the command completes.
 #
 # **Parameters**
 #
-# - *time*: Number of seconds the transition should take
-# - *x*: Target X coordinate
-# - "y*: Target Y coordinate
+# - *pixels*: Target height in pixels
+# - *time*: Number of seconds the transition should take, with a value of `0`
+#   meaning the zoom should happen instantly (default: `0`)
 #
 # For more details see: https://docs.escoria-framework.org/camera
 #
 # @ESC
-extends ESCCameraBaseCommand
-class_name CameraSetPosBlockCommand
+extends ESCBaseCommand
+class_name CameraSetZoomHeightBlockCommand
 
 
 # Tween for blocking
@@ -23,9 +23,9 @@ var _camera_tween: Tween
 # Return the descriptor of the arguments of this command
 func configure() -> ESCCommandArgumentDescriptor:
 	return ESCCommandArgumentDescriptor.new(
-		3,
-		[[TYPE_REAL, TYPE_INT], TYPE_INT, TYPE_INT],
-		[null, null, null]
+		1,
+		[TYPE_INT, [TYPE_INT, TYPE_REAL]],
+		[null, 0.0]
 	)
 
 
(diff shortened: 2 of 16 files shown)


balloonpopper:
Findings : 
1) Change button_set_pos (the current command is outside the room's limits) to 
```
camera_set_pos_block 2 1460 1050
```
And all 3 camera limits for the room to 0,0,2260,2000.
After this, use the set_pos button. As the camera is resetting back to the player, click once to make them walk somewhere else. When Mark stops walking and the camera has stopped updating, click to walk somewhere else. There is an extremely minor camera jump (a handful of pixels). Maybe too little to worry about, but the perfectionist in me noticed it :-)
Happens also after camera_push_block

2) Camera_push_block I noticed moves to the object and not the child ESCLocation. I'm not sure if this is the default behavior or not. It does respect the camera node setting if configured though.

3) I'm not sure if this is fixable (I can't think of any way to fix it), but camera_set_target tweens to where the target was when the tween started. If you set the target back to the player then move the player while it's tweening, the tween doesn't update to match the player's updated position, so jumps when it gets to the destination of the tween.

4) It's easier to watch this bug out than me to try and describe it. Change the set_pos button action to the 
```
camera_set_zoom 1.3 2
wait 2
camera_set_pos_block 2 1460 1050
```

5) And this looks like one of Dante's 9 circles of hell... While I expected a failure, this doesn't fail how I'd expected it to - the zoom doesn't end up where it would if you did the zoom individually (presumably as it keeps zooming until the block finishes.
```
camera_set_zoom 1.3 2
camera_set_pos_block 2 1460 1050
```

6) Another interesting combo. The set_target doesn't fire correctly so it doesn't reset back to the player until you move the character again which appears to be a bug.
```
camera_set_pos 2 1460 1050
camera_shift_block 1000 -400 2 LINEAR
camera_set_target 1 player
```

This is another version of that bug
```
camera_set_pos 2 1460 1050
wait 1
camera_set_target 1 player
```

Basically, as long as I don't try and mix non-blocking and blocking commands, it works fine. A non-blocking walk with blocking camera worked, but mixing camera commands leads to some unexpected situations.

BHSDuncan:
> 1. Change button_set_pos (the current command is outside the room's limits) to
> 
> ```
> camera_set_pos_block 2 1460 1050
> ```
> 
> And all 3 camera limits for the room to 0,0,2260,2000. After this, use the set_pos button. As the camera is resetting back to the player, click once to make them walk somewhere else. When Mark stops walking and the camera has stopped updating, click to walk somewhere else. There is an extremely minor camera jump (a handful of pixels). Maybe too little to worry about, but the perfectionist in me noticed it :-) Happens also after camera_push_block

Yeah, this is pretty difficult to reproduce, and is, like you said, very minor. And actually, given that `accept_input NONE` should still be in effect when that call happens, it looks like the player being able to walk as above is a bug. (I'll raise it.) 

Using `camera_set_target_block` fixes this anyway.
 
> 2. Camera_push_block I noticed moves to the object and not the child ESCLocation. I'm not sure if this is the default behavior or not. It does respect the camera node setting if configured though.

By design. That's how `camera_push` was originally created as far as I can tell (looking at the code in `ESCCamera`.)

> 3. I'm not sure if this is fixable (I can't think of any way to fix it - and I know this is in relation to a non-blocking camera command sorry :D ), but camera_set_target tweens to where the target was when the tween started. If you set the target back to the player then move the player while it's tweening, the tween doesn't update to match the player's updated position, so jumps when it gets to the destination of the tween.
> 
> 3a) Should we have a blocking set_target command?

There already is `camera_set_target_block` ...maybe you missed it? :)

> 4. Then there's this jumpy camera where the set_pos_block unexpectedly zooms...
> 
> ```
> camera_set_zoom 1.3 2
> wait 2
> camera_set_pos_block 2 1460 1050
> ```

This is because the `wait` command is finishing too close to the set time for the zoom. Using `wait 3` (or even `wait 2.5`) corrects this. Probably a good idea to document it, or use the blocking version. :)

> 5. And this looks like one of Dante's 9 circles of hell... While I expected a failure, this doesn't fail how I'd expected it to - the zoom doesn't end up where it would if you did the zoom individually (presumably as it keeps zooming until the block finishes.)
> 
> ```
> camera_set_zoom 1.3 2
> camera_set_pos_block 2 1460 1050
> camera_set_target 0 player
> ```

No. Once a blocking camera command runs, it ends any currently-running tweens, in this case, the unfinished zoom. In fact, every camera command when run will stop any already-running tweening, if any are occurring.

> 6. Another interesting combo. The set_target doesn't fire correctly so it doesn't reset back to the player until you move the character again which appears to be a bug.
> 
> ```
> camera_set_pos 2 1460 1050
> camera_shift_block 1000 -400 2 LINEAR
> camera_set_target 1 player
> ```

Actually this is because `camer_shift_block` is failing validation (going outside limits) and so `camera_set_target` doesn't actually fire. However, expanding the limits corrects this, although that has actually revealed another bug (at the very end, the camera transitions to the top-right of the screen for some reason). I'll fix that by delaying a frame since it looks as though the cause is the tween cancelling juuuuust before it finishes.

> This is another version of that bug
> 
> ```
> camera_set_pos 2 1460 1050
> wait 1
> camera_set_target 1 player
> ```

Similar deal here, although the wait also isn't long enough (1s waiting for a 2s tween).

> Basically, as long as I don't try and mix non-blocking and blocking commands, it works fine. A non-blocking walk with blocking camera worked, but mixing camera commands leads to some unexpected situations. Couldn't think of any other interesting commands to try and mix in with blocking cameras that would be worth testing.

Good stuff. I think we should really call out the dangers of mixing blocking with non-blocking camera commands, and that the blocking commands will work for most people in most cases. Also call out that mixing those calls can result in unexpected behaviour so use at own risk.

balloonpopper:
I tried the following script. This works correctly if the set_pos_block is 2 seconds (i.e. longer than the walk takes to complete) but the set_target goes pear shaped if the walk is still going. Other than this one (and the ones that I've run that worked) I'm running out of synthetic tests for things that might break.
```
[ reset all room limits to 0,0,2260,2000 ] 
accept_input NONE
walk player r7_lower_stairs
camera_set_pos_block 1 1260 550
camera_set_target_block 1 player
accept_input ALL
```

addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_block.gd
```diff
@@ -0,0 +1,67 @@
+# `camera_set_zoom_block magnitude [time]`
+#
+# Zooms the camera in/out to the desired `magnitude`. Values larger than '1' zoom
+# the camera out while smaller values zoom in. These values are relative to the
+# default zoom value of '1', not the current value. As such, while using a value
+# of '0.5' would double the size of the graphics, running the same command again
+# would result in no change. The zoom will happen over the given time period.
+# Blocks until the command completes.
+#
```
balloonpopper, line 9:
```suggestion
# Zooms the camera in/out to the desired `magnitude`. Values larger than '1' zoom
# the camera out while smaller values zoom in. These values are relative to the
# default zoom value of '1', not the current value. As such, while using a value
# of '0.5' would double the size of the graphics, running the same command again
# would result in no change. The zoom will happen over the given time period.
# Blocks until the command completes.
#
# Zoom operations might not be as smooth as desired if the requested zoom
# level results in an edge of the camera meeting any defined camera limits.
#
```

addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_height_block.gd
```diff
@@ -0,0 +1,70 @@
...
+	if not .validate(arguments):
+		return false
+
+	if arguments[0] < 0:
+		escoria.logger.error(
+			self,
+			"[%s]: invalid height. Can't zoom to a negative height (%d)."
+					% [get_command_name(), arguments[0]]
+		)
+		return false
```
balloonpopper, line 43: Setting zoom level to 0 for this command results in 
 ESC (2022-11-10T23:06:44) E esc_camera.gd: Tried to set negative or zero zoom level.
So possibly this should be <= 0?

addons/escoria-core/game/core-scripts/esc/commands/camera_set_zoom_height_block.gd
```diff
@@ -0,0 +1,70 @@
+# `camera_set_zoom_height_block pixels [time]`
+#
+# Zooms the camera in/out so it occupies the given height in pixels.
+# Blocks until the command completes.
+#
+# **Parameters**
+#
+# - *pixels*: Target height in pixels
```
balloonpopper, line 8:
```suggestion
# - *pixels*: Target height in pixels (integer values only)
```

addons/escoria-core/game/core-scripts/esc/commands/camera_shift_block.gd
```diff
@@ -0,0 +1,106 @@
+# `camera_shift_block x y [time] [type]`
+#
+# Shifts the camera by the given horizontal and vertical amounts relative to the
+# current location. Blocks until the command completes.
+#
```
balloonpopper, line 5:
```suggestion
# Shifts the camera by the given horizontal and vertical amounts relative to the
# current location. Blocks until the command completes.
#
# Make sure the destination coordinates are reachable if 
# camera limits have been configured.
#
```

addons/escoria-core/game/core-scripts/esc/commands/camera_shift_block.gd
```diff
@@ -0,0 +1,106 @@
...
+# - *x*: Shift by x pixels along the x-axis
+# - *y*: Shift by y pixels along the y-axis
+# - *time*: Number of seconds the transition should take, with a value of `0`
+#   meaning the zoom should happen instantly (default: `1`)
+# - *type*: Transition type to use (default: `QUAD`)
+#
+# Supported transitions include the names of the values used
+# in the "TransitionType" enum of the "Tween" type (without the "TRANS_" prefix):
+#
+# https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations
```
balloonpopper, line 17: For consistency with the earlier command:
```suggestion
# in the "TransitionType" enum of the "Tween" type (without the "TRANS_" prefix).
#
# See https://docs.godotengine.org/en/stable/classes/class_tween.html?highlight=tween#enumerations
```

balloonpopper:
I wonder if animated gifs work with Github...
set_zoom_height

BHSDuncan:
> Nice! Animated gifs work with Github. I'm going to screen2gif all the bug reports now!!! :D
> 
> Here's the set_zoom_height command. It should centre on his feet (presumably) not wherever it ends up. I've never used the command before, but I can't see why a zoom shouldn't zoom on the player. set_zoom_height [ set_zoom_height ](https://user-images.githubusercontent.com/5151242/201268706-09fa7494-3e75-4ed6-9730-4c99f3683da8.gif) [ ](https://user-images.githubusercontent.com/5151242/201268706-09fa7494-3e75-4ed6-9730-4c99f3683da8.gif)

Nice!

As for where it zooms, I think it's always just on where the screen is centered

For `camera_set_zoom_height`, its meant to zoom in so the camera occupies the given height in pixels (according to the docstring).

For `camera_set_zoom`, there is no target, either...just wherever the camera happens to be centered at the time.

@StraToN will likely know more, but I don't think the intent was to zoom in on a target/character, given this is the zoom target code:

`_zoom_target = Vector2(1, 1) * p_zoom_level`

So it looks like this is working as desired, at least based on the original intent.

balloonpopper (approved):

StraToN (approved):

StraToN:
Looks good to me. Thanks!

PR #646, opened 2022-11-09
fix: Restores default game state after ending a dialog tree, to prevent input from hanging.

MatteoPiovanelli-Laser:
Fixes https://github.com/godot-escoria/escoria-issues/issues/318

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/types/esc_dialog.gd
+++ b/addons/escoria-core/game/core-scripts/esc/types/esc_dialog.gd
@@ -95,4 +95,6 @@ func run():
 		rc = yield(rc, "completed")
 	if rc != ESCExecution.RC_CANCEL:
 		return self.run()
+	# make sure we are resetting GAME_STATE to prevent the game from hanging
+	escoria.current_state = escoria.GAME_STATE.DEFAULT
 	return rc
--- a/addons/escoria-core/game/core-scripts/esc/types/esc_dialog.gd
+++ b/addons/escoria-core/game/core-scripts/esc/types/esc_dialog.gd
@@ -80,6 +80,8 @@ func run():
 		self,
 		"Starting dialog."
 	)
+	# Value of game state when dialog starts running
+	var initial_game_state = escoria.current_state
 	escoria.current_state = escoria.GAME_STATE.DIALOG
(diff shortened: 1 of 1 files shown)


addons/escoria-core/game/core-scripts/esc/types/esc_dialog.gd
```diff
@@ -95,4 +95,6 @@ func run():
 		rc = yield(rc, "completed")
 	if rc != ESCExecution.RC_CANCEL:
 		return self.run()
+	# make sure we are resetting GAME_STATE to prevent the game from hanging
+	escoria.current_state = escoria.GAME_STATE.DEFAULT
```
BHSDuncan, line 99: It might make some sense to store the previous state as a local variable prior to setting the state to `DIALOG` and then use that saved state to restore to here.

BHSDuncan (changes requested):

addons/escoria-core/game/core-scripts/esc/types/esc_dialog.gd
MatteoPiovanelli-Laser: I thought about that. This does the same thing the `say` command does. I preferred to not over complicate things.

MatteoPiovanelli-Laser: for reference: here
https://github.com/godot-escoria/escoria-demo-game/blob/998c9e2535f27889b8a9dbe85af27b726ff7d319/addons/escoria-core/game/core-scripts/esc/commands/say.gd#L118

BHSDuncan: State needs to be encapsulated. Changing it in one place and then expecting/hoping another, external call will otherwise change it isn't a good idea.

And if after restoring the state something else breaks as a result, we should deal with it then. Leaking state changes is a code smell and I can't find anything in the codebase saying that it has to be that way.

BHSDuncan (changes requested):

BHSDuncan:
Also fixed by https://github.com/godot-escoria/escoria-demo-game/pull/652 so closing this one.

PR #648, opened 2022-11-15
feat: allows for the skipping of dialogue after the text is fully vis…

BHSDuncan:
…ible; still need to look at integrating with user options

Closes https://github.com/godot-escoria/escoria-issues/issues/321


BHSDuncan:
REMOVED THE FOLLOWING OPTIONS:

- "Skip dialogs"
- "Max disappear time"

ADDED THE FOLLOWING OPTIONS:

- "Clear text by click only": If true, the only way to clear fully-visible text is to click. It won't disappear after a timeout.
- "Reading speed in WPM": Used to determine how long fully-visible text remains on screen before disappearing. This setting is ignored if "Clear text by click only" is set.
- "Left click action": If "Speed up", left clicking during text rendering will speed up the rendering. If "Instant finish", left clicking during text rendering will force the entire block of text to complete immediately. If "None", left clicking has no effect when text is rendering, but note that, once rendered, you can still click to clear the text.

balloonpopper:
What I've found is below. Half of it is nitpicky, with a couple of exceptions this is very solid (and awesome to have the ability to click through demo scene text that I've read a million times and I'm happy I can make disappear quickly now!)

* The number scale feels wrong. Leave it as-is if you disagree, but it feels like 0-0.2 is the usable (normal) text speed and this would be more intuitive to me as 0-10, or 0-100. I know it's still infinitely configurable as I could use a speed of 0.13579 if I wanted, but the 0-0.2 usable range feels "clunky" to use.
* I was also expecting larger number = faster. It's probably not worth the effort to modify (and it doesn't take effort much for the developer to reverse the number if they go the wrong way) - just a thought.
* I think it's wrong that if normal = 0.1 and fast = 0.1 that they're not both the same speed. 0.1 as a fast speed should be double the speed of a 0.2 normal speed. I expect you've inherited the way the code works, I think we should fix it though.
* Can you please error on a negative speed number? The text never progresses if you use a negative number.
* If I turn on click to progress, the talking animation continues indefinately until I progress the text (and looks very silly :-) ) . I think we should calculate [WPM * words = duration] and stop the talking animation after 'duration'
* WPM = 0 leads to a division by zero error which would be nice to detect and give a better error for. Negative numbers doesn't error but detecting it would be good for warning the user of a bad setting

BHSDuncan:
All of the above have been addressed/implemented with the latest push, with the (agreed upon) exception that "speed" has been changed to "time" so that lower values mean faster rendering. Also the time is now specified in milliseconds and not seconds.

balloonpopper (approved):
Awesome!

StraToN (approved):
Wow :o This is such a huge change, really amazing!!

PR #652, opened 2022-12-03
fix: returns an appropriate RC code when no more options are left at …

BHSDuncan:
…a particular dialog level; this has the somewhat beneficial side effect of moving up a level in dialog if there is one, although this requires a 'stop' command to be present in at least one of the top-most dialog options.

Fixes https://github.com/godot-escoria/escoria-issues/issues/328

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/types/esc_dialog.gd
+++ b/addons/escoria-core/game/core-scripts/esc/types/esc_dialog.gd
@@ -90,9 +90,17 @@ func run():
 		escoria.dialog_player,
 		"option_chosen"
 	) as ESCDialogOption
-	var rc = option.run()
-	if rc is GDScriptFunctionState:
-		rc = yield(rc, "completed")
-	if rc != ESCExecution.RC_CANCEL:
-		return self.run()
+
+	var rc = ESCExecution.RC_OK
+
+	# If no valid option was returned, it means this level of dialog is done.
+	# If this is the case and the current level of dialog has a parent, it means
+	# it is still yielding and so will be shown again.
+	if option:
+		rc = option.run()
+		if rc is GDScriptFunctionState:
+			rc = yield(rc, "completed")
+		if rc != ESCExecution.RC_CANCEL:
+			return self.run()
+
 	return rc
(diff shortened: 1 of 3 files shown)


BHSDuncan:
The upshot here is that nested dialogs are essentially nested `yield`s which makes even going back up levels via an explicit command extremely difficult. That said, this PR looks for an exhaustion of dialog options at the current level and, instead of running the next "option", will return an RC code which will effectively end the current level of yielding and go up one level.

Note that this won't work if you're at the top-most level of options, and so will need at least one option in the top-most level of dialog that has a `stop` command in it.

Also, a workaround for going back up a level in a dialog tree would be to use booleans creatively, e.g.

```
:talk
turn_to worker player
?
    - "Outer option 1"
        #set_global done false # uncomment this line if you want the inner dialog options to be presented again
        say player     "Outer option 1"
        ?
        - "Inner option 1" [!option1,!done]
            say player    "Inner option 1"
            set_global option1 true
        - "Inner option 2" [!option2,!done]
            say player    "Inner option 2"
            set_global option2 true
        - "Inner option 3" [!done]
            set_global done true
        !
    - "Outer option 2"
        say player    "Outer option 2"
        stop
!
```

Will go up a level from the inner options to the outer options if you select "Inner option 3". And, to re-visit the inner options, the outer options could reset the `done` variable to `false` before entering the nested "inner option" dialog block (since anything after the "inner option" block will be ignored).

balloonpopper (approved):
Tested with the below successfully

:talk
?
	- "This stop wouldnt work without this PR"
		stop
	- "Blah"
		set_global done false
		?
		- "Done" [!done]
			say player	"1a"
			set_global done true

		- "2" [!done]
			say player	"2a"
			set_global hide_this_option true
		- "Once only" [hide_this_option]
			say player "only once"
			set_global hide_this_option false
		- "Quit" [!done]
			say player	"4a"
			turn_to worker worker_face_down
			stop
		!
	- "Not blah"
		say player "not blah"
	- "quit"
		say player "s"
		stop
!

BHSDuncan:
@balloonpopper feel like updating room 6 to make use of this?

StraToN (approved):

StraToN:
Thank you very much! 🥳

PR #653, opened 2022-12-04
feat: Added 1 and 2 direction animations to wizard character creator

balloonpopper:
This lets the wizard create 1 or 2 direction characters as well as the existing 4/8. Helpful when creating an NPC that you'll talk to but otherwise doesn't do anything.


StraToN (approved):
Looks good to me

BHSDuncan (approved):

PR #654, opened 2022-12-04
fix: disallows dialog blocks that aren't properly indented; however, …

BHSDuncan:
…this also changes how dialog options are parsed and will require substantial testing.

Fixes https://github.com/godot-escoria/escoria-issues/issues/335


StraToN (approved):

StraToN:
Sorry for the delay!
Thanks for the fix 👍 !

PR #657, opened 2022-12-21
fix: Fix p_child is null exit on quit

balloonpopper:
This is to fix the bug where if you open a room (e.g. room 4) and go into room 3, then quit the game by force closing the window, you get the error:
E 0:00:05.946   remove_child: Parameter "p_child" is null.
  <C++ Source>  scene/main/node.cpp:1308 @ remove_child()
  <Stack Trace> esc_camera.gd:27 @ _exit_tree()

Only seems to show up with simplemouse, 9 verb didn't show it.

### DIFF
--- a/addons/escoria-core/game/scenes/camera_player/esc_camera.gd
+++ b/addons/escoria-core/game/scenes/camera_player/esc_camera.gd
@@ -24,8 +24,8 @@ func _ready():
 
 
 func _exit_tree():
-	remove_child(_tween)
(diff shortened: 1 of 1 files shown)


StraToN (approved):

StraToN:
> @StraToN this fix seems to make sense, but can you think of a reason why this would only happen for Simple Mouse and not 9verbs?

No idea, sorry :(

PR #660, opened 2022-12-25
fix: Correct error message for missing animation player

balloonpopper:
This error fires when there's no path, so removing the reference to the path from the output

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_item.gd
+++ b/addons/escoria-core/game/core-scripts/esc_item.gd
@@ -476,7 +476,12 @@ func get_animation_player() -> Node:
 				if child is AnimatedSprite or \
 						child is AnimationPlayer:
 					player_node_path = child.get_path()
-		if not has_node(player_node_path):
+		if player_node_path == "":
+			escoria.logger.warn(
+				self,
+				"Can not find animation_player or animated sprite for %s." % global_id
+			)
+		elif not has_node(player_node_path):
 			escoria.logger.warn(
 				self,
 				"Can not find animation_player node at path %s for %s." % [player_node_path, global_id]


BHSDuncan (commented):
Doesn't it make sense to show the path anyway for debugging purposes? E.g. informing the developer that they put in path $x/y/z and can quickly spot pathing mistakes?

StraToN (approved):

balloonpopper:
@BHSDuncan when this code fires there is no path. In the full code block, the player_node_path will be set to a valid path based on the children of the ESCItem being the right node type. The developer isn't putting in a path, they're just doing something like"set_state frog jumping" and there is no node that can provide a "jumping" animation. Unless I'm missing something, the "if not has_node(player_node_path):" could be also be rewritten as "if player_node_path == null:". I'm not sure what other metadata we can provide the developer but happy to add something if there's something you can see that would be good to provide.

		if player_node_path == "":
			for child in self.get_children():
				if child is AnimatedSprite or \
						child is AnimationPlayer:
					player_node_path = child.get_path()
		if not has_node(player_node_path):
			escoria.logger.warn( self, "Can not find animation_player or animated sprite for %s." % global_id )

I can't see a way that the top logic could return a path and the bottom code would be valid.

BHSDuncan:
> @BHSDuncan when this code fires there is no path. In the full code block, the player_node_path will be set to a valid path based on the children of the ESCItem being the right node type. The developer isn't putting in a path, they're just doing something like"set_state frog jumping" and there is no node that can provide a "jumping" animation. Unless I'm missing something, the "if not has_node(player_node_path):" could be also be rewritten as "if player_node_path == null:". I'm not sure what other metadata we can provide the developer but happy to add something if there's something you can see that would be good to provide.
> 
> ```
> 	if player_node_path == "":
> 		for child in self.get_children():
> 			if child is AnimatedSprite or \
> 					child is AnimationPlayer:
> 				player_node_path = child.get_path()
> 	if not has_node(player_node_path):
> 		escoria.logger.warn( self, "Can not find animation_player or animated sprite for %s." % global_id )
> ```
> 
> I can't see a way that the top logic could return a path and the bottom code would be valid.

Maybe we should discuss further as `player_node_path` is an exported instance variable that has nothing to do with the method being called: It looks entirely possible that the variable in question could be valid and not empty.

balloonpopper:
Thanks @BHSDuncan - let me know if you are happy with the change I just pushed.

BHSDuncan (approved):

PR #662, opened 2022-12-28
fix: allows for string joining that's compatible with pre-3.5 Godot

BHSDuncan:


balloonpopper:
(Very lightly) tested with 3.3 and we're backwards compatible back to 3.3 again with this patch. 3.2 is not supported due to calls in the code to "emit_changed()" which isn't supported pre v3.3

balloonpopper (approved):

StraToN (approved):

StraToN:
Thanks!

PR #664, opened 2022-12-29
fix: Movable commands on non-movable items will be handled gracefully.

balloonpopper:
Movable commands on non-movable items will be handled gracefully.
Non-movable ESCItems will default to direction[0] for talk animations if they exist.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_item.gd
+++ b/addons/escoria-core/game/core-scripts/esc_item.gd
@@ -86,7 +86,8 @@ export(String) var trigger_out_verb = "trigger_out"
 # If true, the player can interact with this item
 export(bool) var is_interactive = true
 
-# Whether this item is movable
+# Whether this item is movable. A movable item will be scaled with the terrain
+# and be moved with commands like teleport and turn_to.
 export(bool) var is_movable = false
 
 # If true, player orients towards 'interaction_direction' as
@@ -284,15 +285,15 @@ func _ready():
 			if not self.is_connected("body_exited", self, "element_exited"):
 				connect("body_exited", self, "element_exited")
 
-	# If object can be in the inventory, set default_action_inventory to same as
-	# default_action, if default_action_inventory is not set
-	if use_from_inventory_only and default_action_inventory.empty():
-		default_action_inventory = default_action
+		# If object can be in the inventory, set default_action_inventory to same as
+		# default_action, if default_action_inventory is not set
+		if use_from_inventory_only and default_action_inventory.empty():
+			default_action_inventory = default_action
 
-	# Perform a first terrain scaling if we have to.
-	if (!is_exit or dont_apply_terrain_scaling) and is_movable:
-		_movable.last_scale = scale
-		_movable.update_terrain()
+		# Perform a first terrain scaling if we have to.
+		if (not is_exit or dont_apply_terrain_scaling) and is_movable:
+			_movable.last_scale = scale
+			_movable.update_terrain()
 
 
 # Mouse exited happens on any item that mouse cursor exited, even those UNDER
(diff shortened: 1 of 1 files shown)


addons/escoria-core/game/core-scripts/esc_item.gd
```diff
@@ -642,51 +673,91 @@ func set_angle(deg: int, wait: float = 0.0):
...
 
-		if animation_player.is_playing():
-			animation_player.stop()
+	if animation_player.is_playing():
+		animation_player.stop()
 
-		if animations.speaks[_movable.last_dir].mirrored \
-				and not _movable.is_mirrored:
-			_sprite_node.scale.x *= -1
+	if is_movable and animations.speaks[_movable.last_dir].mirrored \
```
StraToN, line 729: This test of `is_movable` value is included in `check_talk_possible()`. Is it necessary again here?

addons/escoria-core/game/core-scripts/esc_item.gd
```diff
@@ -642,51 +673,91 @@ func set_angle(deg: int, wait: float = 0.0):
...
+		return
+		
+	var animation_player = get_animation_player()
 
-		if animation_player.is_playing():
-			animation_player.stop()
+	if animation_player.is_playing():
+		animation_player.stop()
 
+	if is_movable:
```
StraToN, line 749: This test of is_movable value is included in check_talk_possible(). Is it necessary again here?

StraToN (approved):

addons/escoria-core/game/core-scripts/esc_item.gd
```diff
@@ -642,51 +673,91 @@ func set_angle(deg: int, wait: float = 0.0):
...
+func check_talk_possible():
+	if is_movable and (_movable.last_dir < 0 \
+			or _movable.last_dir >= animations.speaks.size()):
+		escoria.logger.warn(
+			self,
+			"Node %s cannot talk. Its \"last_dir\" parameter is invalid : %s." \
+			%[self, _movable.last_dir]
+		)
+		return false
+	if animations == null:
```
BHSDuncan, line 695:
```suggestion
	if not is_instance_valid(animations):
```

addons/escoria-core/game/core-scripts/esc_item.gd
```diff
@@ -642,51 +673,91 @@ func set_angle(deg: int, wait: float = 0.0):
...
+		)
+		return false
+	if animations.speaks.size() == 0:
+		escoria.logger.warn(
+			self,
+			"Node %s cannot talk. Its \"animations.speaks\" array is empty : %s." \
+			%self
+		)
+		return false
+	if not get_animation_player():
```
BHSDuncan, line 709: Above you're using `!` instead of `not`.  I don't have a preference either way, but let's try to be consistent with what we use. Please pick whichever one you want to use and update the other usages of it for this PR. Thanks!

BHSDuncan (changes requested):

addons/escoria-core/game/core-scripts/esc_item.gd
balloonpopper: This one is necessary as you only want to check animations...mirrored when is_movable is true otherwise this check will fail.

addons/escoria-core/game/core-scripts/esc_item.gd
balloonpopper: This check is required for the script logic rather than whether the item is movable.

addons/escoria-core/game/core-scripts/esc_item.gd
balloonpopper: The only one I could spot was this one "if (!is_exit or dont_apply_terrain_scaling)" which I hadn't touched from the previous codebase. I've updated it to be "not" for consistency.

StraToN: Indeed, we historically chose to use `not` instead of `!`. I agree this is not the common usage in other languages though, but that's the choice we made :)

BHSDuncan (approved):

PR #668, opened 2022-12-31
fix: Remove call to stop dialog to correct player mirroring

balloonpopper:
Without this line being removed, during a conversation, stop talking gets called by both :
image
and
image

The result of this is that this code is called twice
```
		if animations.speaks[_movable.last_dir].mirrored \
				and not _movable.is_mirrored:
			_sprite_node.scale.x *= -1
```
Which, when you have a mirrored talking animation, means that the character gets turned to face the wrong way when they finish talking.


StraToN:
Looks correct. Good catch!

StraToN (approved):

BHSDuncan (approved):

PR #673, opened 2023-01-24
fix: Fix bug where inventory is invisible by default

balloonpopper:
Fixes https://github.com/godot-escoria/escoria-issues/issues/351

### DIFF
--- a/addons/escoria-core/game/escoria.gd
+++ b/addons/escoria-core/game/escoria.gd
@@ -168,6 +168,7 @@ func run_event_from_script(script: ESCScript, event_name: String):
 
 # Called from escoria autoload to start a new game.
 func new_game():
+	escoria.game_scene.escoria_show_ui()
 	run_event_from_script(escoria.start_script, escoria.event_manager.EVENT_NEW_GAME)
 
 


BHSDuncan:
Would it make more sense to call this out in documentation and have the devs use the `set_gui_visible` command? I only ask that in case there's ever a situation where a game dev doesn't want the UI immediately shown? E.g. they might want to show it after some cutscene or other intro.

balloonpopper:
I think users would expect a working inventory by default. 
This fix doesn't stop the user from keeping the UI hidden. I tested and confirmed that if you put in your first room 
```
:setup
set_gui_visible false
```
the inventory will stay invisible from when you click "New game" until you re-enable it.

StraToN (approved):

BHSDuncan:
> I think users would expect a working inventory by default. This fix doesn't stop the user from keeping the UI hidden. I tested and confirmed that if you put in your first room
> 
> ```
> :setup
> set_gui_visible false
> ```
> 
> the inventory will stay invisible from when you click "New game" until you re-enable it.

Fair enough.

BHSDuncan (approved):

PR #675, opened 2023-01-29
Let speak animations be optional

romanzes:
I was following Escoria's step-by-step guide. It [says](https://docs.escoria-framework.org/en/devel/getting_started/step_by_step/2_create_player_character.html#telling-escoria-about-the-walkcycle) that the speak animations are optional:

> The Speak animations are optional and only required if your game needs them.

However, when I tried running the game without them, it crashed with the message `8 animation angles specified but 0 'speaks' animation(s) given`. I think it makes sense to do this validation only when any speak animations are defined, i.e. their number is greater than 0.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc_item.gd
+++ b/addons/escoria-core/game/core-scripts/esc_item.gd
@@ -442,9 +442,10 @@ func validate_animations(animations_resource: ESCAnimationResource) -> void:
 		else:
 			_validate_animations_property_all_not_null(animations_resource.idles, "idles")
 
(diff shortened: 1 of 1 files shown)


balloonpopper:
@romanzes Thanks for the PR! Unfortunately that's an issue with the documentation being out of date. From previous experience there will be issues if there aren't speak, walk and talk animations defined, hence why the check is there. It's a bigger piece of work to make them optional.

romanzes:
@balloonpopper thanks for the clarification. I guess I'll open a PR against the docs repo then. As far as I can see, this check is present in the published version of Escoria too.

PR #676, opened 2023-02-11
fix: Issue 336 - unclickable area with Simplemouse

balloonpopper:
Fixes Simplemouse related issues I discovered: 
- The area at the bottom of the screen is unclickable
- the completed form couldn't be given to the worker
- Right clicks on the background of most demo screens were being ignored because the backgrounds didn't cover the whole window
- Right clicks in the inventory window when not on an inventory item are no longer ignored
- Fast right clicks in the inventory window were being ignored as they were being treated as an (unhandled) double click.

Fixes : https://github.com/godot-escoria/escoria-issues/issues/336
Duplicate : https://github.com/godot-escoria/escoria-issues/issues/279


BHSDuncan (approved):

StraToN (approved):

StraToN:
Thanks a lot!

PR #678, opened 2023-02-14
move dialog settings to another script to fix exported game issues

skogach:
There's a problem when you try to export game and launch it, you'll get a grey screen. In the log there's errors about  `Unknown class: "EditorPlugin"` presumably because editor classes aren't exported. The problem is that `escoria-dialog-simple` plugin exports it's class name to other scripts which get settings via it. So I've moved settings to separate script that extends `Resource` same as `ESCProjectSettingsManager`. It fixes the problem.

### DIFF
--- a/addons/escoria-dialog-simple/esc_dialo_simple_settings.gd
+++ b/addons/escoria-dialog-simple/esc_dialo_simple_settings.gd
@@ -1,0 +1,20 @@
+extends Resource
+class_name ESCSimpleDialogSettingsManager
+
+const SETTINGS_ROOT="escoria/dialog_simple"
+
+const AVATARS_PATH = "%s/avatars_path" % SETTINGS_ROOT
+const TEXT_TIME_PER_LETTER_MS = "%s/text_time_per_letter_ms" % SETTINGS_ROOT
+const TEXT_TIME_PER_LETTER_MS_FAST = "%s/text_time_per_fast_letter_ms" % SETTINGS_ROOT
+const READING_SPEED_IN_WPM = "%s/reading_speed_in_wpm" % SETTINGS_ROOT
+const CLEAR_TEXT_BY_CLICK_ONLY = "%s/clear_text_by_click_only" % SETTINGS_ROOT
+const LEFT_CLICK_ACTION = "%s/left_click_action" % SETTINGS_ROOT
+
+const STOP_TALKING_ANIMATION_ON = "%s/stop_talking_animation_on" % SETTINGS_ROOT
+
+const LEFT_CLICK_ACTION_SPEED_UP = "Speed up"
+const LEFT_CLICK_ACTION_INSTANT_FINISH = "Instant finish"
+const LEFT_CLICK_ACTION_NOTHING = "None"
+
+const STOP_TALKING_ANIMATION_ON_END_OF_TEXT = "End of text"
+const STOP_TALKING_ANIMATION_ON_END_OF_AUDIO = "End of audio"
--- a/addons/escoria-core/game/scenes/dialogs/state_machine/dialog_say.gd
+++ b/addons/escoria-core/game/scenes/dialogs/state_machine/dialog_say.gd
@@ -39,7 +39,7 @@ func initialize(character: String, type: String, text: String) -> void:
 	_type = type
 	_text = text
 	_stop_talking_animation_on_option = \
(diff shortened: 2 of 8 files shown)


BHSDuncan:
Thanks for this. I have a couple comments, but first, `addons/escoria-dialog-simple/esc_dialo_simple_settings.gd` has a typo in the filename (I'm sure you can see it now).

The issue you bring up is valid but also highlights something I think we overlooked: The core has an explicit reference to a class located in a plugin, which is not a good idea, e.g. what if someone wants to implement their own dialog manager?

@StraToN @balloonpopper can you see any issues or challenges in moving `ESCDialogPlayer` (and its state machine) into the Simple Dialog addon? One thing we'll need to do is make sure that `ESCProjectSettingsManager` is somehow exposed so that the Simple Dialog addon can get at things like the speech folder, speech file extension, etc.

BHSDuncan (changes requested):
`addons/escoria-dialog-simple/esc_dialo_simple_settings.gd` has a typo in the filename. Please correct.

addons/escoria-dialog-simple/esc_dialo_simple_settings.gd
```diff
@@ -0,0 +1,20 @@
+extends Resource
+class_name ESCSimpleDialogSettingsManager
+
+const SETTINGS_ROOT="escoria/dialog_simple"
```
BHSDuncan, line 4:
```suggestion
const SETTINGS_ROOT = "escoria/dialog_simple"
```

BHSDuncan (changes requested):

skogach:
I fixed typo and whitespaces, also I forgot to commit changes to project file, now it's there.

BHSDuncan (approved):

StraToN:
> The issue you bring up is valid but also highlights something I think we overlooked: The core has an explicit reference to a class located in a plugin, which is not a good idea, e.g. what if someone wants to implement their own dialog manager?

We have two solutions: 
- either define a hook mechanism for any user to personalize some parts of the dialog manager. But then we'd need to define what hooks would be necessary for each needs.
- or allow a plugin to override the default behavior. Which would still require to design a clear API for this purpose.

> @StraToN @balloonpopper can you see any issues or challenges in moving ESCDialogPlayer (and its state machine) into the Simple Dialog addon? One thing we'll need to do is make sure that ESCProjectSettingsManager is somehow exposed so that the Simple Dialog addon can get at things like the speech folder, speech file extension, etc.

`ESCProjectSettingsManager` is a static class filled with static const values, so it can be used from anywhere, including a plugin.

I am a bit mixed about the idea of moving `ESCDialogPlayer` out of core. That would mean that any plugin that needs to use it will have to include it (as-is or edited for their own purpose). Again in that case I'd go in favor of some mechanism allowing one to override the default ESCDialogPlayer behavior if needed.

BHSDuncan:
Let’s discuss further on Discord when you have time. On Feb 15, 2023, at 10:59 AM, Julian Murgia ***@***.***> wrote:﻿

The issue you bring up is valid but also highlights something I think we overlooked: The core has an explicit reference to a class located in a plugin, which is not a good idea, e.g. what if someone wants to implement their own dialog manager?

We have two solutions:

either define a hook mechanism for any user to personalize some parts of the dialog manager. But then we'd need to define what hooks would be necessary for each needs.
or allow a plugin to override the default behavior. Which would still require to design a clear API for this purpose.


@StraToN @balloonpopper can you see any issues or challenges in moving ESCDialogPlayer (and its state machine) into the Simple Dialog addon? One thing we'll need to do is make sure that ESCProjectSettingsManager is somehow exposed so that the Simple Dialog addon can get at things like the speech folder, speech file extension, etc.

ESCProjectSettingsManager is a static class filled with static const values, so it can be used from anywhere, including a plugin.
I am a bit mixed about the idea of moving ESCDialogPlayer out of core. That would mean that any plugin that needs to use it will have to include it (as-is or edited for their own purpose). Again in that case I'd go in favor of some mechanism allowing one to override the default ESCDialogPlayer behavior if needed.

StraToN (approved):

StraToN:
@skogach I sincerely apologize for the time we took for reviewing and merging this PR. Thanks a lot!
Fixes https://github.com/godot-escoria/escoria-issues/issues/354
Closes https://github.com/godot-escoria/escoria-demo-game/pull/684

PR #681, opened 2023-03-09
feat: add fullscreen option in options menu

enekonieto:
I'm unsure if I should have used `set_pressed` instead of `set_pressed_no_signal` in `addons/escoria-core/ui_library/menus/options/options.gd` line 64.


StraToN:
At first I was surprised we didn't implement fullscreen management yet 🤔 but then I went like "Oh. Indeed, we didn't."
Thanks!

StraToN (approved):

PR #682, opened 2023-03-10
feat: validates for multiple start locations set in an ESCRoom and up…

BHSDuncan:
…dates scene tree with configuration warnings next to the offending ESCLocation nodes

Closes https://github.com/godot-escoria/escoria-issues/issues/323


balloonpopper (approved):
Looks good to me. Indicators get set as expected when multiple start locations are configured - great work!

PR #683, opened 2023-03-20
fix: removes type as ESCResourceCache checks the type of 'res' and ca…

BHSDuncan:
…n expect a non-Resource-derived value, e.g. ResourceInteractiveLoader

### DIFF
--- a/addons/escoria-core/game/core-scripts/resources/esc_resource_descriptor.gd
+++ b/addons/escoria-core/game/core-scripts/resources/esc_resource_descriptor.gd
@@ -4,12 +4,12 @@ class_name ESCResourceDescriptor
 
 
 # The resource being described
-var res: Resource
+var res
 
 # Whether the resource is permanent
 var permanent: bool
 
 
-func _init(res_in: Resource, permanent_in: bool) -> void:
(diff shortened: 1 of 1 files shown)


StraToN:
Can you explain the requirement or provide a link to it? I don't get why you would put a `ResourceInteractiveLoader` instance as the resource defined by ESCResourceCache.

BHSDuncan:
It's how the cache was designed (not by me).

If a resource to load is queued up and isn't already loading or in the queue, the following line is executed:

`var res = ResourceLoader.load_interactive(path)`

This returns a `ResourceInteractiveLoader` that starts loading the specified resource by chunks via its `poll()` method.

The use of this makes sense, but for whatever reason, the same field for the resource itself is also used to store the `ResourceInteractiveLoader`.  The code itself works just fine with the fix I've put in.

Here's the docs for `load_interactive()`: https://docs.godotengine.org/en/3.5/classes/class_resourceloader.html#class-resourceloader-method-load-interactive

StraToN (approved):

StraToN:
Alright :) 
Seems a bit weird that the ResourceLoader works like so, but I must admit I don't edit that code very often. Thanks!

PR #687, opened 2023-04-22
fix: delegates all dialog logic entirely to dialog plugin

BHSDuncan:
This allows for more decoupling between the core scripts and the dialog-handling to allow game devs more flexibility. For now, the interface/contract remains relatively the same (see `ESCDialogManager`) with the dialog player in the core only needing to call the appropriate method(s) in the plugin without needing to manage state, etc. All that is handled by the plugin.


addons/escoria-core/game/scenes/dialogs/esc_dialog_player.gd
```diff
@@ -67,41 +37,93 @@ func _add_states_to_machine() -> void:
...
 
+	_determine_say_dialog_manager(type)
 
-# Called when a dialogue line is to be sped up.
-func speedup() -> void:
-	_change_state("say_fast")
+	if is_a_parent_of(_say_dialog_manager):
+		remove_child(_say_dialog_manager)
 
+	add_child(_say_dialog_manager)
```
StraToN, line 50: I am always scared when I see an `add_child()` clause somewhere. This `say()` method is supposed to be called multiple times, right? So it is quite possible that the `_say_dialog_manager` node is added as child each time the `say()` method is called.

addons/escoria-core/game/scenes/dialogs/esc_dialog_player.gd
```diff
@@ -67,41 +37,93 @@ func _add_states_to_machine() -> void:
...
+# - type: The dialog chooser type to use (default: "simple")
 func start_dialog_choices(dialog: ESCDialog, type: String = "simple"):
-	states_map["choices"].initialize(self, dialog, type)
-	_change_state("choices")
+	_determine_choose_dialog_manager(type)
+
+	if is_a_parent_of(_choose_dialog_manager):
+		remove_child(_choose_dialog_manager)
+
+	add_child(_choose_dialog_manager)
```
StraToN, line 67: Same as above: nodes duplication is possible with this `add_child()` call here.

StraToN (changes requested):

StraToN (approved):

StraToN:
All good! Thanks!

PR #692, opened 2023-05-21
Make ESCBackground node be the 1st child of ESCRoom on run

StraToN:
Fixes the issue with using buttons in room10 (click on verb then on item, not the default action)


StraToN:
Superseded by https://github.com/godot-escoria/escoria-demo-game/pull/693

PR #694, opened 2023-05-23
fix: fixes incorrect handler signature; also checks for null since it…

BHSDuncan:
… appears that sometimes the tween parameter is null when the handler is called


StraToN (approved):

StraToN:
All good! Good fix, thanks!

PR #701, opened 2023-07-08
fix: add .cfg files and R/W permissions to android export

StraToN:
Fixes https://github.com/godot-escoria/escoria-issues/issues/158


BHSDuncan (approved):
Just a reminder that your personal/local values in the `keystore/debug*` entries remain, but you're ok with that for the time being.

PR #702, opened 2023-08-26
Change interaction_direction to interaction_angle

StraToN:
And some validation checks on exported parameters

Fixes https://github.com/godot-escoria/escoria-issues/issues/304


addons/escoria-core/game/core-scripts/behaviors/esc_movable.gd
```diff
@@ -260,14 +260,25 @@ func walk_stop(pos: Vector2) -> void:
...
 	# orient towards the defined interaction direction set on the object
 	# (if any), can be ESCItem or ESCLocation
 	if walk_context.target_object and \
 			walk_context.target_object.node.player_orients_on_arrival:
-		var orientation = walk_context.target_object.node.interaction_direction
+		
+		var orientation = _get_dir_deg(walk_context.target_object.node.interaction_angle + 90,
+			parent.animations)
+		if orientation >= parent.animations.dir_angles.size() or orientation < 0:
+			escoria.logger.warn(self, "Orientation is invalid for item %s (received value: %s)\nPath to item: %s.\nDefaulting to 0."
```
BHSDuncan, line 274:
```suggestion
			escoria.logger.warn(self, "Orientation is invalid for item '%s' (received value: %s degrees)\nPath to item: '%s'.\nDefaulting to 0 degrees." 
```

BHSDuncan (changes requested):
Looks good, but in the future let's try to avoid mixing fixes in a single commit/PR. That way we improve traceability and make things easier to find/roll back if necessary.

BHSDuncan (approved):

balloonpopper (changes requested):
I'm not sure how this is supposed to work now. I set the interaction angle for both the wrench and pipe in scene 5. Nothing seems to change regardless of the angle I've put. I've tried 0,1,2,30,45,90,8000 and the interaction is still the same each time.

balloonpopper (approved):

PR #704, opened 2023-09-12
fix: allows queue_event to execute events located in a room instead o…

BHSDuncan:
…f only objects in a room

Fixes https://github.com/godot-escoria/escoria-issues/issues/366

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/queue_event.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/queue_event.gd
@@ -37,42 +37,50 @@ func validate(arguments: Array):
 	if not .validate(arguments):
 		return false
 
-	if not escoria.object_manager.has(arguments[0]):
+	if not escoria.object_manager.has(arguments[0]) and not _is_current_room(arguments[0]):
 		escoria.logger.error(
 			self,
-			"Object with global id %s not found." % arguments[0]
+			"No object or room with global id '%s' found." % arguments[0]
 		)
 		return false
-	var node = escoria.object_manager.get_object(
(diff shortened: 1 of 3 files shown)


addons/escoria-core/game/core-scripts/esc/commands/queue_event.gd
```diff
@@ -87,3 +95,27 @@ func run(arguments: Array) -> int:
...
+# The object node or the current room.
+func _get_script_node(global_id: String):
+	var node = null
+
+	if escoria.object_manager.has(global_id):
+		node = escoria.object_manager.get_object(
+			global_id
+		).node
+	else:
+		node = escoria.main.current_scene
```
StraToN, line 119: In this code, we'll return the room node script when `global_id` doesn't refer to a registered item. This will work, but do we want a check for this ?

StraToN (changes requested):

addons/escoria-core/game/core-scripts/esc/commands/queue_event.gd
BHSDuncan: I had thought about that, but given that we've already done that check prior to calling this method, I put a precondition ("PRE") in the docstring. I'm not a fan of duplicating code if possible. Happy to discuss, though.

StraToN: Nothing against using a PRE in docstring here, but I find it rather incomplete then: I'd add a notice whether provided global_id is null or inexisting would default to returning the room script.

StraToN (approved):

PR #711, opened 2023-10-20
fix: use default text for dialog options if text is untranslated

enekonieto:
At this moment default text is not being using in `esc_dialog_option.gd`, this PR fixes that.

I included some minor fixes also:

- fix: delete `.translation` files (`.translation` is in `.gitignore` but I think they were commited before)
- fix: delete unused translated texts (`ROOM6_dialog_1` and `ROOM6_dialog_2`)
- fix: typo in one translated text (missing tilde in `ROOM1_look_wall_item_4` spanish translation)


enekonieto:
I forgot to mention that I added a default text in the room6 dialog with the worker so the issue can be tested.

StraToN (approved):

StraToN:
Looks good to me. Waiting for other approvals. Thank you very much!

addons/escoria-core/game/core-scripts/esc/types/esc_dialog_option.gd
```diff
@@ -10,7 +10,7 @@ const REGEX = \
 
 
 # Option displayed in the HUD
-var option: String setget ,get_option
+var option: String setget ,get_translated_text
```
BHSDuncan, line 13:
```suggestion
var option: String setget ,get_translated_option
```

addons/escoria-core/game/core-scripts/esc/types/esc_dialog_option.gd
```diff
@@ -51,9 +51,19 @@ func load_string(option_string: String):
 		)
 
 
-func get_option():
+func get_translated_text():
```
BHSDuncan, line 54:
```suggestion
func get_translated_option():
```

BHSDuncan (changes requested):
I think it best to keep the updated method name close to the original intent of the referenced variable. Past that it looks ok to me.

enekonieto:
> I think it best to keep the updated method name close to the original intent of the referenced variable. Past that it looks ok to me.

Done!

BHSDuncan (approved):

PR #713, opened 2023-10-26
fix: move ESCCondition parsing logic out of the constructor

bolinfest:
fix: move ESCCondition parsing logic out of the constructor

This makes it possible to create an `ESCCondition` programmatically.
Unfortunately, this does not fix the issue with `ESCCondition`
where `comparison_regex.compile(REGEX)` is called *every time* an
`ESCCondition` is instantiated.

I have read all sorts of forums, and no one seems to address this
issue head-on. The question of, "Why doesn't GDScript support static
variables" is not addressed clearly in the docs, IMHO.

The only "workaround" I can find is this article on Singletons/Autoload:

https://docs.godotengine.org/en/latest/tutorials/scripting/singletons_autoload.html

Though this seems incredibly heavyweight for caching a `RegEx`.


bolinfest:
Note that the PR does something very wrong, as-is, which is it passes `{}` as the first argument to logger.

I see that sometime in the past 18 months, the `Logger` API was changed to require an `owner` object of some sort. When updating my own project, this seemed to make it impossible to use from a `static` function. I ended up deleting the logger statements because I couldn't find a workaround.

StraToN (approved):

StraToN:
This change looks interesting, thank you for it.
It is right that GDScript is a bit limited when we're starting to require some higher level language features. Hopefully this will be leveraged in the future - either when we move to Godot 4.x, or use C#, which is not in our plans at the moment.

bolinfest:
Though before anyone merges this, nb:

> Note that the PR does something very wrong, as-is, which is it passes {} as the first argument to logger.

How should I fix that?

BHSDuncan:
@StraToN do we want to merge this in despite us moving to the new interpreter?

StraToN:
> Though before anyone merges this, nb:
> 
> > Note that the PR does something very wrong, as-is, which is it passes {} as the first argument to logger.
> 
> How should I fix that?

Logging methods require the calling class so its name is put in the log message. You may simply pass `self`.

StraToN:
@bolinfest Any news on this? I think it only needs the logger parameter to be set. When this is done I am fine merging this @BHSDuncan no matter if we're to moving to the new interpreter.

bolinfest:
> Logging methods require the calling class so its name is put in the log message. You may simply pass self.

Is `self` valid in a `static` function in Godot? ChatGPT doesn't think so:

https://chat.openai.com/share/9d22a90e-c1d4-4ca6-8c34-21f6b6d25370

StraToN:
> > Logging methods require the calling class so its name is put in the log message. You may simply pass self.
> 
> Is `self` valid in a `static` function in Godot? ChatGPT doesn't think so:
> 
> [chat.openai.com/share/9d22a90e-c1d4-4ca6-8c34-21f6b6d25370](https://chat.openai.com/share/9d22a90e-c1d4-4ca6-8c34-21f6b6d25370)

Ah, ok I missed your point.

Here is what I can propose then: add a new error_static() method in ESCLogger that you should be able to call in a static way:
```
	# Error log
	func error(owner: Object, msg: String):
		var context = owner.get_script().resource_path.get_file()
		error_static(context, msg)

	func error_static(context: String, msg: String):
		printerr(formatted_message(context, msg, "E"))
		push_error(formatted_message(context, msg, "E"))
		if ESCProjectSettingsManager.get_setting(
			ESCProjectSettingsManager.TERMINATE_ON_ERRORS
		):
			assert(false)
			escoria.get_tree().quit()
```

You can then call it like so:
```
escoria.logger.error_static(
							"ESCConditionParser",
							"Invalid comparison type detected: %s" %
									comparison_string +
							"Comparison type %s unknown" %
									ESCUtils.get_re_group(
										result,
										"comparison"
									)
						)
```

StraToN:
@BHSDuncan I think this is superseded by ASHES script now. WDYT?

BHSDuncan:
I believe it is, especially since we no longer rely on RegEx for much of anything. There is going to be a fair amount of dead code related to the old interpreter, and we should likely look at each section on a case-by-case basis.

PR #715, opened 2024-02-14
fix: Allow ':' in dialog options default text

enekonieto:
Having a dialog option like `ROOM6_dialog_bye:"I'm done: Bye!"` is shown as `I'm done`. This PR fixes that.


BHSDuncan:
Thanks for this, but let's revisit once the new interpreter is in place and see if it's still an issue there. It very well could be, but I'd like to hold off on applying fixes to anything dialog- or scripting-related until the new interpreter is merged.

Thanks again!

PR #716, opened 2024-02-15
fixed dialogue appearing only on top of the screen

madalv:
When creating a new character using the Escoria Wizard, the floating dialogue appears only on the top of the screen, regardless of the position of the Dialogue Location set in the character scene.

The Wizard creates a instance of a ESCLocation class for the dialog_position:
image

Whereas the floating dialog requires an ESCDialogLocation class (addons/escoria-dialog-simple/types/floating.gd):
image

### DIFF
--- a/addons/escoria-wizard/CharacterCreator.gd
+++ b/addons/escoria-wizard/CharacterCreator.gd
@@ -1338,8 +1338,8 @@ func export_player(scene_name) -> void:
 	yield(get_tree(), "idle_frame")
 
 	# Add Dialog Position to the ESCPlayer
-	var dialog_position = ESCLocation.new()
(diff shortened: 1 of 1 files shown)


BHSDuncan:
@balloonpopper I know you're really busy but whenever you have a minute if you could review this that'd be hugely appreciated. (We don't want to make any assumptions since this addon is your baby.)

Thanks!

StraToN (approved):
I assume this change was required as the plugin was created before the addition of ESCDialogLocation. Besides, ESCDialogLocation inherits ESCLocation, so I believe all is good here. I will therefore accept the merge.

Of course if @balloonpopper arguments otherwise later, then we'll revert this and fix another way.

PR #722, opened 2025-01-19
feat: adds static analysis to the compiler and is also able to be invoked via tool script available under: Project->Tools

BHSDuncan:


addons/escoria-core/tools/script_analysis/analyzers/esc_exit_scene_analyzer.gd
```diff
@@ -0,0 +1,167 @@
...
+
+
+# These are vars because they have to be, but should be treated as consts
+var TRANSITION_COMMAND_NAME = TransitionCommand.new().get_command_name()
+var CHANGE_SCENE_COMMAND_NAME = ChangeSceneCommand.new().get_command_name()
+var ACCEPT_INPUT_COMMAND_NAME = AcceptInputCommand.new().get_command_name()
+
+
+# Returns: Array of strings for any issues detected.
+func analyze(statements: Array) -> void:
```
StraToN, line 17: The prototype of the method is wrong. Should be Array[String] instead of void.

addons/escoria-core/plugin.gd
```diff
@@ -59,6 +64,7 @@ func _on_warning_popup_confirmed():
 func _disable_plugin():
 	remove_autoload_singleton("escoria")
 	set_filesystem_hide_esc_files()
+	#remove_tool_menu_item(ASHES_ANALYZER_MENU_ITEM)
```
StraToN, line 67: This line can be removed as it is called in `_exit_tree()`

StraToN (approved):

PR #723, opened 2025-02-02
Fix simple mouse plugin

StraToN:
Full revamp of the UI.
Still not the best, having to mousewheel through actions is not the best.


BHSDuncan:
Question: If I mousewheel to, say, the "look" action (the eye), and I mouse over the button in room 2, the action automatically changes to what I'm assuming is the default action for that button, in this case, the "use" action (the wrench). Is this on purpose? 

Also, if I hover over the inventory while I'm in windowed mode, and I move the mouse off the edge of the game window (while still over the inventory panel), the inventory panel is stuck in the raised position until I mouse back over and move the mouse off of the panel to a non-inventory panel area of the game window.

StraToN:
@BHSDuncan I forgot your comment regarding inventory close. Fixed.
I also fixed the cursor not getting back to "walk" icon when unhovering inventory item.

BHSDuncan:
If I run room 1 directly (F6), I get a crash on startup: `esc_item.gd#637`. Screenshot below:

image

StraToN:
Strange. Ok I'll check this tonight. Thanks!

BHSDuncan:
Hmm, while the crash is fixed, the issue still exists. Maybe it's ok if the inventory pane stays raised in the scenario I described? We can discuss.

Also, a couple of errors in the debugger:

image

PR #730, opened 2025-03-03
Migration to Godot 4.4 (UID files created)

StraToN:


BHSDuncan (approved):
I'm assuming these are all needed.

Given that this is going into `main` now, should we look to rebase the outstanding branches on `main` instead of `develop`, and then remove `develop` as the default branch?

PR #735, opened 2025-03-04
feat: adds event flag conditionals, specifically to the TK event flag

StraToN:
This new command add a flag on an object's specific event.
Helps making an event as telekinetic (TK) after it was triggered once normally, for example.

note: keeping this one draft as I would like to check if it requires some saving work.

Example:
```
# in plant object script
:use
	if (!plant_moved):
		set_state($plant, "move_left")
		plant_moved = true
		force_event_flag("plant", "use", "TK")
	else:
		turn_to("sophie", "plant")
		say("sophie", "I'm not going to spend my time moving it...")
```
------
edit: another, preferable way to this is to add a predicate to flags, like this:

```
:use | TK<plant_moved>
	if (!plant_moved):
		set_state($plant, "move_left")
		plant_moved = true
	else: # TK flag applying
		turn_to("sophie", "plant")
		say("sophie", "I'm not going to spend my time moving it...")
```
This is a better solution because it relies on the value of a defined global that is already being saved in savegame data, and thus doesn't require more data to be saved - contrarily to this 1st implementation.

## Summary by CodeRabbit

- **New Features**
  - Event flags can now include optional conditions based on global variables, allowing for more dynamic event behavior.
  - Added support for toggling telekinetic (TK) interactions in the wall item dialog, controlled by a new global variable.
  - Introduced a method to directly query global variables in event scripts.

- **Improvements**
  - Standardized event flag naming and usage for better consistency.
  - Enhanced event parsing and flag handling to support condition-based logic.
  - Refined telekinetic applicability checks for input events.
  - Improved layout configuration in test scenes by unifying to a single layout mode.

- **Documentation**
  - Updated plugin descriptions and author information for clarity and accuracy.

- **Style**
  - Minor whitespace and formatting cleanups in scripts and scene files.


BHSDuncan:
The `TK<plant_moved>` approach ended up being implemented.

For the foreseeable future, the part inside of the angle brackets must be limited to a single global variable. Otherwise, we have to deal with the possibility of calling commands as well as dealing with coroutines.

This means that if a comparison or multi-clause predicate is to be used, its result must be stored in a single global variable that is used inside the angle brackets.

BHSDuncan (approved):

PR #736, opened 2025-03-07
[feature] add optional exit_direction property to ESCExit

bolinfest:


BHSDuncan:
`ESCItem` (from which `ESCExit` inherits) does have a `custom_data` Dictionary available which could be used in a manner such as this.

That said, I don't really see any harm in having a new property in `ESCExit` as set out in this PR.

@StraToN what do you think?

addons/escoria-core/game/core-scripts/esc_exit.gd
```diff
@@ -21,6 +21,9 @@ class_name ESCExit
 # Sound effect to play when changing the scene
 @export var switch_sound: Resource # (String, FILE, "*.ogg,*.mp3,*.wav")
 
+# UI hint that can be useful when determining what mouse cursor to use.
```
StraToN, line 24:
```suggestion
## String value to be used as UI hint, for example to determine the mouse cursor to use.
```

StraToN:
`custom_data` is primarily used for including specific data that is not being save by default in savegames. 
I was indeed not too happy I was using a naming convention to manage the mouse cursor for left and right exit, so this solution fits a bit better. Using a naming convention instead for this purpose remains at gamedev's discretion, of course, but currently forcing them into this style is a bit too inelegant, regarding the current other features Escoria proposes. 

I just added a small suggestion regarding the comment. We're now striving to use GDScript docstrings starting by ## so we can have a better documentation to be generated *and* available inside Godot editor.

bolinfest:
@StraToN updated the PR with your recommended comment!

BHSDuncan (approved):

PR #741, opened 2025-03-30
fix: avoid crashing when loading savegame with items in inventory

enekonieto:
fix(savegames): avoid crashing when loading savegame with items in inventory


BHSDuncan:
What's the actual crash message/cause? I agree that the `run` method in `InventoryAddCommand` ultimately ignores anything but the first element in the passed-in array, so the change in this PR should probably be accepted, but given that the change in this PR removes an argument that would only be accessed should `savegame_inventory` have the key in question, I'm not exactly sure what this fixes (i.e. what the crash is).

StraToN (approved):

StraToN:
I feel bad I missed this although I swear I thought I had tested this. My apologies.
Thank you for the fix!

BHSDuncan:
Ah yes, the array element being used as a key. Good find!

PR #746, opened 2025-04-10
fix: click doesn't reset walk animation anymore.

StraToN:
Now if you left click continuously while player is walking, the walk animation won't reset anymore (which was resulting in player moving but animation was frozen).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved how gameplay interruptions work during interactions. Player actions now result in a more predictable halt of in-game movement—ensuring that when you click on background areas or items, your character stops as expected for a smoother, more responsive experience.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
@@ -311,10 +311,11 @@ func queue_background_event(channel_name: String, event: ESCGrammarStmts.Event) 
 #
 # #### Parameters
 # - exceptions: an optional list of events which should be left running or queued
-func interrupt(exceptions: PackedStringArray = []) -> void:
+func interrupt(exceptions: PackedStringArray = [], stop_walking = true) -> void:
 	if escoria.main.current_scene != null \
 			and escoria.main.current_scene.player != null \
-			and escoria.main.current_scene.player.is_moving():
+			and escoria.main.current_scene.player.is_moving() \
+			and stop_walking:
 		escoria.main.current_scene.player.stop_walking_now()
 
 	for channel_name in _running_events.keys():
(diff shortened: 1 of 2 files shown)


addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -311,10 +311,11 @@ func queue_background_event(channel_name: String, event: ESCGrammarStmts.Event)
 #
 # #### Parameters
 # - exceptions: an optional list of events which should be left running or queued
```
BHSDuncan, line 313: Should update parameter list in docstring with `stop_walking`.

BHSDuncan (changes requested):

addons/escoria-core/game/core-scripts/esc/esc_event_manager.gd
```diff
@@ -311,10 +311,13 @@ func queue_background_event(channel_name: String, event: ESCGrammarStmts.Event)
...
-func interrupt(exceptions: PackedStringArray = []) -> void:
+# - stop_walking: boolean value (default true) determining whether the player 
+# (if any) has to be interrupted walking or not.
+func interrupt(exceptions: PackedStringArray = [], stop_walking = true) -> void:
 	if escoria.main.current_scene != null \
 			and escoria.main.current_scene.player != null \
-			and escoria.main.current_scene.player.is_moving():
+			and escoria.main.current_scene.player.is_moving() \
+			and stop_walking:
```
BHSDuncan, line 320: This may seem a bit pedantic, but, what do you think about moving the `stop_walking` clause to the front of the predicate? That is, something like:

```
if stop_walking \
  and escoria.main.current_scene != null \
  and ...
```

...since it's a simple boolean and could short circuit the checks that involved more dereferencing.

What do you think?

BHSDuncan (commented):
One additional small thing. It isn't strictly necessary but it's worth consideration.

BHSDuncan (approved):

PR #750, opened 2025-05-05
fix: tooltip position

clarkjc:
Keep the position of tooltips consistent, even if the viewport is resized, by setting the game size to the design size

## Summary by CodeRabbit

- **Refactor**
  - Updated how the game size is initialized to use fixed project settings instead of dynamically retrieving the viewport size. This ensures consistent sizing based on configured values.

### DIFF
--- a/addons/escoria-core/game/esc_autoload.gd
+++ b/addons/escoria-core/game/esc_autoload.gd
@@ -126,7 +126,9 @@ var creating_new_game: bool = false
 var temp: int = 0
 
 # The game resolution
-@onready var game_size = get_viewport().size
+@onready var game_size = Vector2(
+	ProjectSettings.get_setting("display/window/size/viewport_width"),
+	ProjectSettings.get_setting("display/window/size/viewport_height"))
 
 # The current state of the game
 @onready var current_state = GAME_STATE.DEFAULT


BHSDuncan:
@clarkjc I'm a bit curious as to the nature of the issue. The fix itself should only set `game_size` on startup, and so if the viewport is later resized, it should have no effect on `game_size`. 

Can you share with us a bit more detail of how the fix addresses the tooltip issue you saw?

clarkjc:
> @clarkjc I'm a bit curious as to the nature of the issue. The fix itself should only set `game_size` on startup, and so if the viewport is later resized, it should have no effect on `game_size`.
> 
> Can you share with us a bit more detail of how the fix addresses the tooltip issue you saw?

Prior to this fix, I observed that `game_size` was set to the initial size of the viewport on startup. If the viewport was later resized, `game_size` would not be updated and the tooltip position code in `addons/escoria-ui-simplemouse/game.gd` and `addons/escoria-core/game/core-scripts/esc_tooltip.gd` would use the now incorrect limits to constrain tooltip positions.

Since the stretch mode is set to `canvas_items` in the project settings, I realized that using the actual viewport size is incorrect in this context, and that the design size specified in the project settings is correct.

StraToN:
@clarkjc Do you have a precise procedure to reproduce the issue in the first place? Ie, having your tooltip wrongly positioned (so without your fix). 
I didn't test too much before committing I have to admit, but I'm testing now with and without your fix and I fail to observe a difference.

clarkjc:
@StraToN I found this issue in one of my own projects, but the easiest way to see it here is like so:

1. In the project settings, set the window size overrides to something other than the desired window size. I used 800x600 for testing today

test_settings

2. Hover over the door in the first room and see that the tooltip is positioned incorrectly without my fix

bad


(I noticed this in my project because some window managers (or some OSes) treat the requested window size as a hint rather than a demand, so I was getting a smaller or larger window than the 1280x900 requested)

StraToN:
Thank you for the explanation. That was then due to using the windows size overrides, instead of desired size :+1: .

PR #763, opened 2025-08-06
fix: type check in inc_global and dec_global

Dragonfyle:
There was a bug where `inc_global` and `dec_global` commands expected an int value but the compiler produces floats, which crashes the game whenever you use one of them. I changed the type check at the command level. It seems to make more sense to be able to increment/decrement by a float since you can assign a float to an escoria global.
I've seen some other commands, e.g camera_push expect either float or int, but it looks like the compiler never actually produces an int.
```
esc_scanner.gd

func _number() -> void:
	while _is_digit(_peek()):
		_advance()

	# Fraction part?
	if _peek() == '.' and _is_digit(_peek_next()):
		_advance()

		while _is_digit(_peek()):
			_advance()

	_add_token(ESCTokenType.TokenType.NUMBER, _source.substr(_start, _current - _start).to_float())
```

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/commands/dec_global.gd
+++ b/addons/escoria-core/game/core-scripts/esc/commands/dec_global.gd
@@ -16,7 +16,7 @@ class_name DecGlobalCommand
 func configure() -> ESCCommandArgumentDescriptor:
 	return ESCCommandArgumentDescriptor.new(
 		1,
-		[TYPE_STRING, TYPE_INT],
+		[TYPE_STRING, TYPE_FLOAT],
 		[null, 1]
 	)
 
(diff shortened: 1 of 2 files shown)


BHSDuncan:
@Dragonfyle yeah, since the compiler isn't strongly typed, there's just a "numeric" type. In order to keep things simple (at least for now) and avoiding having to implement things like duck typing, we've left them as floats.

That said, perhaps it makes some degree of sense to remove both `inc_global` and `dec_global`. I only say that because globals can be incremented and decremented using arithmetic operations, making these two functions redundant.

@StraToN thoughts?

StraToN:
> @Dragonfyle yeah, since the compiler isn't strongly typed, there's just a "numeric" type. In order to keep things simple (at least for now) and avoiding having to implement things like duck typing, we've left them as floats.
> 
> That said, perhaps it makes some degree of sense to remove both `inc_global` and `dec_global`. I only say that because globals can be incremented and decremented using arithmetic operations, making these two functions redundant.
> 
> @StraToN thoughts?

Indeed, these 2 functions are redundant. They could simply be removed instead ; would that be an issue @Dragonfyle ? Or do you specifically need these 2 functions?

Dragonfyle:
@BHSDuncan @StraToN I'm fine with that. I used them because I thought they fit the ASHES style better than direct arithmetic

StraToN:
I think we can merge this PR even if `inc_global` and `dec_global` will probably be removed in the future - it's not big deal to keem them in for the time being, as these functions won't do much more that they currently do.

StraToN (approved):

PR #764, opened 2025-08-31
Fix use_from_inventory_only use in action manager

enekonieto:
Make wrench usable in Room5 even if not marked with `use_from_inventory_only`.

Less lines are changed than it appears. Lines from 285 to 370 only lost a tab character.

Tested with simple_mouse and 9verbs.

## Summary by CodeRabbit

* **New Features**
  * More flexible item interactions: items that don't require inventory can be used or combined directly in the scene.
  * Inventory items clicked in-scene can become the active tool for combining.

* **Bug Fixes**
  * Smarter action resolution and safer fallbacks when combination events are missing.
  * Clearer feedback when combinations are blocked (including one-way items).
  * Wrench now usable directly in the scene without taking it into inventory.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
@@ -661,9 +661,6 @@ func perform_inputevent_on_object(
 
 
 func _telekinetic_applies_to(event: ESCGrammarStmts.Event) -> bool:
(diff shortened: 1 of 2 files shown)


BHSDuncan:
@StraToN this is going to cause conflicts with the update we're doing for event tracking to prevent events with the same name being overwritten. So when that PR is created, we should take our time with it.

PR #765, opened 2025-09-03
fix(compiler): avoid crash when target is null

enekonieto:
I don't know why in our game we are getting to the situation where the event has no target name, but it is happening. In any case, safety is never too much.

I'm not entirely comfortable returning an empty string, but since I can't return null, it's the best option I can think of.


StraToN:
I am not totally comfortable with grammar code. That was @BHSDuncan 's work, can you have a look at it and approve if it's ok for you? Thanks!

enekonieto:
Ops, pushed to the wrong branch and this PR is now dirty, closing it.

enekonieto:
Besides, PR #754 fixed this issue.

PR #768, opened 2025-09-06
Hovered inventory texture default to inventory texture

oierbravo:
I think that makes more sense defaulting to the inventory_texture when missing the inventory_texture_hovered.
Avoids having to always fill both variables with the same texture.
This affects only items that have a different texture for the inventory.

## Summary by CodeRabbit

* **Bug Fixes**
  * Hovered inventory icons now correctly fall back to the primary inventory texture when no explicit hover texture is set, ensuring consistent visuals.
  * Prevents mismatched or missing hover textures that could occur on some items.
  * Existing behavior is preserved when no textures are set; automatic texture detection still applies.
  * No changes required for existing content or integrations; improvements apply automatically.


StraToN:
Thanks!

PR #770, opened 2025-09-16
fix(Logger): avoid clash with new Logger native class

enekonieto:
Godot 4.5 introduced new native Logger class which clashed with Escoria Logger class.

Perhaps Escoria Logger should use the new class (https://godotengine.org/releases/4.5/#script-backtracing), but that is out of scope of this PR.


StraToN:
Thanks! I had tested this too when 4.5 was released and that was basically the only change needed for Escoria to work in it :)

PR #771, opened 2025-09-19
fix: Revert of PR767

StraToN:
@oierbravo My apologies, I merged #767 a bit too fast yesterday and I forgot that we are in the process of designing with @BHSDuncan a solution the achieve the same goal without using a compisite key. We'll try to come up with the solution soon, in the meantilme I revert this one. 

## Summary by CodeRabbit

- Bug Fixes
  - Interactions now trigger more reliably when combining items or using actions on specific targets.
  - Resolved cases where the wrong interaction could be matched or some combinations were ignored.
  - Preserved existing fallback behavior when no specific interaction is found.

- Refactor
  - Simplified how interactions are registered and looked up to improve consistency and reduce edge cases.
  - Streamlined combined-action handling for more predictable behavior across objects.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_action_manager.gd
@@ -293,21 +293,25 @@ func _get_event_to_queue(
 					do_combine = false
 
 				if do_combine:
-					var action_name = action
-					if combine_with.global_id:
-						action_name += " " + combine_with.global_id
-				
-					var combine_name = action
-					if combine_with.global_id:
-						combine_name += " " + target.global_id
+#						var target_event = "%s %s" % [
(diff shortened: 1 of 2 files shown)


oierbravo:
I understand the situation and it's your call.
However, in my head, it's a minimal temporal fix to make it work so it's compatible with the old code and ASHES, keeping the ball rolling.

From my understanding the new aproach will be compatible with the actual ASHES implementation and what we expect from the events with target... so, changing the data structure should not break or affect the behavior or the esc scripts being transparent for the actual and new escoria users.

We are using extensively events with targets... it's kind of critical for us... and I think it also blocks new users from starting new proyects because it's a basic functionality for point&clicks.

In any case, whatever it's the decision... thanks for your time and effort maintaining escoria.

PR #790, opened 2025-11-14
feat: support - in identifiers starting with $ in ASHES

bolinfest:
In my own game, I had to change the global id of an item from `closet-door` to `closet_door` so I could write:

```
if $closet_door is "open_door":
```

because previously, the logic to match an identifier in ASHES did not accept `-`.

## Summary by CodeRabbit

* **New Features**
  * Global identifiers (those beginning with $) now accept hyphens (e.g., $closet-door), allowing more readable and expressive names for game entities.
  * This change preserves existing identifier rules for non-global names, so only prefixed global IDs gain hyphen support—no other tokenization or error behavior visible to users is changed.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/compiler/esc_scanner.gd
+++ b/addons/escoria-core/game/core-scripts/esc/compiler/esc_scanner.gd
@@ -7,6 +7,11 @@ extends RefCounted
 class_name ESCScanner
 
 
+## In ASHES, an identifier starting with `$` is interpreted as a lookup in
+## `ESCObjectManager.get_object()` with the value after the `$`.
+const GLOBAL_ID_PREFIX: String = "$"
+
+
 var _tokens: Array = []
 
 var _keywords: Dictionary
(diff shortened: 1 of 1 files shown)


addons/escoria-core/game/core-scripts/esc/compiler/esc_scanner.gd
```diff
@@ -263,7 +263,12 @@ func _is_alphanumeric(c: String) -> bool:
 
 
 func _identifier() -> void:
-	while _is_alphanumeric(_peek()):
+	# Identifiers starting with '$' are used as shorthand for global IDs and
+	# are allowed to contain hyphens (e.g. `$closet-door`). Other identifiers
+	# still treat '-' as the subtraction operator.
+	var allow_hyphen_in_identifier: bool = _source[_start] == '$'
```
BHSDuncan: Please move the `$` character/string into a class-level constant with a descriptive name.

bolinfest:
Added a comment and changed the loop slightly so that `_peek()` is only called once per iteration.

I don't have a strong preference on any of the style aspects here.

BHSDuncan:
> Added a comment and changed the loop slightly so that `_peek()` is only called once per iteration.
> 
> I don't have a strong preference on any of the style aspects here.

Looks good. Last change I'd like to see is giving `c` a more descriptive name, e.g. `current_char` or some such. It'll help improve readability, something which we're slowly trying to improve across the codebase.

Once that's done, we'll merge!

bolinfest:
@BHSDuncan updated

BHSDuncan (approved):

PR #791, opened 2025-11-16
fix: checks inventory variable value

oierbravo:
Adapts `.has` method from the inventory manager to use the value when it's present.
There are cases when an item is placed and taken many times.
In ashes: `in inventory`  uses this method.

### DIFF
--- a/addons/escoria-core/game/core-scripts/esc/esc_inventory_manager.gd
+++ b/addons/escoria-core/game/core-scripts/esc/esc_inventory_manager.gd
@@ -16,7 +16,9 @@ class_name ESCInventoryManager
 ## [br]
 ## Returns a `bool` value. (`bool`)
 func inventory_has(item: String) -> bool:
-	return escoria.globals_manager.has("i/%s" % item)
+	if not escoria.globals_manager.has("i/%s" % item):
+		return false
+	return escoria.globals_manager.get_global("i/%s" % item)
 
 
 ## Retrieves all inventory items.[br]


BHSDuncan:
@oierbravo this looks ok but can you explain the change a bit more? I'm not entirely clear on how it achieves the use case you're describing (although I'm sure it does). Thanks!

oierbravo:
Hi!
The `remove_item(item: String)` function removes the item changing `"i/global_id"` global variable. 
`if $item in inventory:` uses this to determine if an item is in the inventory.

If the item is removed from the inventory... `in inventory`won't work the second time even if the item is not in the inventory.

BHSDuncan:
> Hi! The `remove_item(item: String)` function removes the item changing `"i/global_id"` global variable. `if $item in inventory:` uses this to determine if an item is in the inventory.
> 
> If the item is removed from the inventory... `in inventory`won't work the second time even if the item is not in the inventory.

I'm still not entirely clear (could be a language barrier thing, and that's ok; we'll sort it out!).

I'll try to summarize below what I think you're saying, but if I'm mistaken, please correct it in a reply:

What's currently happening (i.e. what your fix addresses):

- Item is in inventory.
- `if $item in inventory:` is called and returns true.
- `remove_item` is called to remove the item from the inventory.
- `if $item in inventory:` is called again and returns true (?)

Let's keep discussing until I understand. :)

oierbravo:
Exactly! :)
Calling it again returns true even if the inventory is removed.

PR #792, opened 2025-11-16
Makes interpreter factory overridable

oierbravo:
Changes the instantiation from the static class to the escoria singelton.
This way it's overridable.

## Summary by CodeRabbit

* **Refactor**
  * Restructured internal script interpreter factory architecture to use instance-based access instead of static class methods. Updated multiple components to access the interpreter through the centralized manager, improving code organization and maintainability.


BHSDuncan:
@oierbravo I'm not sure I want this change. It potentially makes unit testing much more difficult since it would rely on needing to start up the entire project so autoload could be run. 

One benefit of a factory is to allow other implementations to be used, so long as they follow the same "interface". So the factory is still overrideable when implemented the way it currently is.

I'm happy to keep discussing with you, though, if you have additional thoughts or reasoning.

oierbravo:
Hi!

At the moment unit testing is far from possible with the current code base... however the compiler/interpreter is a very good candidate to start with... I understand your concerns... if you are already planing that and/or you see that the interpreter can have tests... it's super good news.

Right now, as far as I know, there is no way to extend/switch the interpreter. Calling the factory from the static class makes it non feasible.

Theres are other ways... that avoid changing the static side.

### Option 2:
- Use the project.godot to configure the interpreter. This is the approach used for the dialog_player.
- `ui/default_dialog_scene="res://addons/escoria-core/game/scenes/dialogs/esc_dialog_player.gd"`
- This way the Factory can call a custom interpreter.

### Option 3:
- Interpreter Plugin. 
- More expandable, more work... but from a core prespective more clean. Using the same approach as for the ui plugins... this allows to ensure that all the necessary function are present as well as other features that the interpreter may need.


From my perspective... the third option is the best one. The second one it's a middle ground between this PR and the plugin.

BHSDuncan:
The static factory behaves like a singleton. If you want to switch or change the interpreter, I believe you should be able to change the script path string in the factory, although you're right in that that's not very modular and would require a change to the "core" addon. However, one could argue that the interpreter is meant to be the canonical implementation of ASHES and any change to the interpreter/compiler/scanner/whatever could very well require far-reaching changes to the core addon.

Now, admittedly, I'm not 100% sure how Godot 4 handles subtyping (i.e. of `ESCInterpreter`, since that's what's used as the type for most variables that hold/use the intpreter), but it may be worth testing out at some point. I can't honestly see many people wanting to change out the interpreter at this point, though.

@StraToN what do you think about the discussion above? I'm interested in hearing your take on things.

Thanks again, everyone. This is a good discussion to have.

oierbravo:
Good discussion indeed!!😁

From my experience with other frameworks... allowing overrides, plugins, hooks etc... it's beneficial for the core. It makes possible to test new features or experimental features.. within the core itself or within addons. 

Changing the core addon code should be avoided, in my opinion.... it makes very difficult to get new updates once is done... and opens the door to multiversion code maintenance requirement.

From our use case... godot handles well the subtyping. We have an extended [ESCInterpreter](https://git.fosil.eu/gymkhana/gymkhana/src/branch/main/addons/escoria-ui-return-monkey-island/overrides/esc_interpreter.gd) that changes 2 functions... works like a charm.

BHSDuncan:
Even if we go ahead with this, I don't see any updates to the Escoria autoload singleton. Is that missing? Or am I not understanding something?

oierbravo:
@BHSDuncan The escoria autoload is already creating an instance that is overridable.
https://github.com/godot-escoria/escoria-demo-game/blob/7dec6290d3ec64381dfbb20d7de0a40a012987ed/addons/escoria-core/game/escoria.gd#L51-L52

This PR apart from changing the static method to non static unifies the calling of the `interpreter_factory` so every time is called from the singleton.

oierbravo:
@BHSDuncan  it seems a too old branch... created before the static checking action. Too much out of scope linter issues.
I think we can close it and create a new one with the new codebase... just tell me or close this PR directly.

BHSDuncan:
Sure, I'll close the PR and you can create a new one whenever you're ready.

PR #793, opened 2025-11-16
feat: add input mode changed signal

oierbravo:
Useful for giving feedback to the players.

Added wait cursor example to simplemouse_ui.

In our game... we change the cursor to a watch to inform the player that input is disabled.
With play testing we found that many user where trying to do actions when `walk_block` was called. The cursor change solved the issue.

## Summary by CodeRabbit

* **New Features**
  * Input mode now broadcasts change notifications so UI reacts immediately.
  * Cursor updates automatically when input mode changes (wait vs. target).
* **Behavior Changes**
  * Verb-change input is recognized more broadly, improving responsiveness across devices.
  * Double-clicking inventory items no longer triggers actions (double-click disabled).
* **Bug Fixes**
  * Improved gamepad connection and input handling for more consistent behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/godot-escoria/escoria-demo-game/pull/793?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -5,6 +5,9 @@
 extends Resource
 class_name ESCInputsManager
 
+## Emitted when input mode changes.
+signal input_mode_changed(new_mode)
```
BHSDuncan: If possible, please add a type for `new_mode`.

BHSDuncan (changes requested):
Can you please also add an example of this in use? I'd imagine it would have to be in one of the UI implementations, unless you can suggest a better way.

addons/escoria-core/game/esc_inputs_manager.gd
oierbravo: Done!
Added `int` because it's an enum.

BHSDuncan: I can't remember if we made the enum a proper one in Godot 4, but if we haven't then we should at some point because we can then use it as a type.

addons/escoria-core/game/esc_inputs_manager.gd
```diff
@@ -27,7 +30,12 @@ const ESC_UI_PRIMARY_ACTION = "esc_ui_primary_action"
 
 
 ## The current input mode
-var input_mode = INPUT_ALL
+var input_mode = INPUT_ALL:
+	get:
+		return input_mode
+	set(new_mode):
+		input_mode = new_mode
+		input_mode_changed.emit(new_mode)
```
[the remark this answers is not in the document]
BHSDuncan: @oierbravo I've actually never done getters and setters like this in GDScript...I always use methods. Can you check and see if what the comment above is saying actually happens? Or whether GDScript is smart enough to know not to perform an infinite recursion. Thanks!

oierbravo:
> Can you please also add an example of this in use? I'd imagine it would have to be in one of the UI implementations, unless you can suggest a better way.

I've added our use case and how we found it.

I think that the simple mouse plugin can be a good candidate for an example... I could do it in another PR.

addons/escoria-core/game/esc_inputs_manager.gd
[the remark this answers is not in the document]
oierbravo: I'm starting to use them and I kind of like it for cases like this. :)
Tried to follow the documentation https://docs.godotengine.org/es/4.x/tutorials/scripting/gdscript/gdscript_basics.html#when-setter-getter-is-not-called

Extract: _Using the variable's name to set it inside its own setter or to get it inside its own getter will directly access the underlying member, so it won't generate infinite recursion and saves you from explicitly declaring another variable:_

This time the rabbit is wonderland🤭

BHSDuncan:
I think I'd prefer to have the example as part of this PR, It makes sense to keep the example together with the feature, unless you disagree @StraToN ?

BHSDuncan:
@oierbravo just sending a quick nudge here concerning my last comment. Adding that example to this PR would be helpful. Thanks again, and good work.

oierbravo:
Added waiting cursor example.
Changes the cursor to a watch when input is not allowed.
Visible in the first cut scene.

addons/escoria-ui-simplemouse/game.gd
```diff
@@ -77,9 +70,16 @@ var _is_gamepad_connected = false
 # Tracks the mouse's current position onscreen.
 var _current_mouse_pos: Vector2 = Vector2.ZERO
 
+# Cursor resources
+var _cursor_target = load("res://addons/escoria-ui-simplemouse/cursors/new/target_a.png")
+var _cursor_wait = load("res://addons/escoria-ui-simplemouse/cursors/new/wait_b.png")
```
BHSDuncan: Can we please put the pointer images in a better-named directory? e.g. `cusrors/images/` or something other than just `new/`?

addons/escoria-ui-simplemouse/game.gd
```diff
@@ -492,3 +492,14 @@ func _on_event_done(_return_code: int, _event_name: String):
 func _on_room_ready():
 	$mouse_layer/verbs_menu.set_by_name(VERB_WALK)
 	$mouse_layer/verbs_menu.action_manually_changed = false
+
+func _on_input_mode_changed(new_mode):
+	if new_mode == ESCInputsManager.INPUT_NONE:
+		Input.set_custom_mouse_cursor(_cursor_wait, Input.CURSOR_ARROW)
+		return
```
BHSDuncan, line 495: This is a good first attempt at an example; however, what happens if the input mode is later set back and the most recent pointer before the waiting one is, say, `talk` or `use`?

Take a look at `verbs_mouseicons.gd` and see how `set_by_name` makes use of the various cursors/pointers. Maybe figure out a nice way to work with that and use a similar system, just for the sake of consistency.

BHSDuncan (changes requested):

addons/escoria-ui-simplemouse/game.gd
oierbravo: just followed the existing pattern.
It's a good time to reorganize code, scene & assets of this part.
@BHSDuncan  Should I tackle this in this PR?

addons/escoria-ui-simplemouse/game.gd
oierbravo: I'll take a look!

oierbravo: Changed to use the mouse verbs approach.
Now also uses the current action instead of the "walk" verb.

addons/escoria-ui-simplemouse/game.gd
oierbravo: I propose just a PR for reorganizing the images. I think we should delete the old cursors and attribution. New ones don't require attribution... but can be nice to add it... i've found the asset pack! :)

BHSDuncan: Sorry for the delay in responding.

Agreed that any reorg can be done in a separate PR. As for the deletion of the old cursors, etc., I think @StraToN should weigh in, just to be safe.

BHSDuncan (approved):

PR #1221, opened 2026-03-27
fix(wizard): character creator Godot 4 compatibility - .tscn format and AtlasTexture

pfaz69:
## Summary

The Character Creator wizard currently saves character scenes as compressed binary `.scn` files using `ImageTexture` with embedded pixel data. This was appropriate for Godot 3 but creates several issues in Godot 4.

## Problems addressed

1. **Inconsistent file format**: The Godot 4 editor creates scenes as `.tscn` (text), but the wizard generates `.scn` (binary). This makes character scenes impossible to inspect, diff, or version-control, unlike every other scene in the project.

2. **Enormous file sizes**: `ImageTexture.create_from_image()` embeds raw pixel data into the scene file. With high-resolution spritesheets, this produces ~8 MB binary files. Attempting to convert to `.tscn` (e.g. via "Save Scene As") results in files of ~763 MB, which are impractical to work with.

3. **Typed array crash**: Assigning a generic `Array` to typed array properties (`dir_angles = []`) causes a runtime error in Godot 4 during character export.

## Changes

- **`.scn` → `.tscn`**: Output format changed to text scene, consistent with Godot 4 conventions
- **`AtlasTexture` instead of `ImageTexture`**: Animation frames now reference the source `.png` spritesheets with region coordinates, instead of embedding copied pixel data. This reduces a typical character scene from ~763 MB to ~23 KB
- **`FLAG_COMPRESS` removed** from `ResourceSaver.save()` to enable text format output
- **`.clear()` instead of `= []`** for typed arrays to prevent crash

## Testing

Tested with a 4-direction character using 3200x2000 spritesheets (walk, idle, talk). The generated `.tscn` file is ~23 KB and works correctly in-game: walking, idle, talk animations, and direction angles all function as expected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Character creator now exports scenes in the correct Godot 4 text format instead of legacy format.

* **Performance**
  * Optimized animation frame generation to improve resource efficiency and reduce memory overhead.

### DIFF
--- a/addons/escoria-wizard/CharacterCreator.gd
+++ b/addons/escoria-wizard/CharacterCreator.gd
@@ -836,7 +836,8 @@ func spritesheet_on_export_button_pressed() -> void:
 	var anim_name: String = ""
 	var dirnames = []
 
-	var scene_name = "%s/%s.scn" % [get_node(CHARACTER_PATH_NODE).text, get_node(NAME_NODE).get_node("node_name").text]
+	# Fix: use .tscn (text) instead of .scn (binary) for Godot 4 compatibility
+	var scene_name = "%s/%s.tscn" % [get_node(CHARACTER_PATH_NODE).text, get_node(NAME_NODE).get_node("node_name").text]
 
 	if FileAccess.file_exists(scene_name):
 		get_node(GENERIC_ERROR_NODE).dialog_text = \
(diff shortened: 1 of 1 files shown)


oierbravo:
Tested... It works!
- Player creation OK
- NPC creation OK

BHSDuncan:
This will need to be rebased before it can be merged. Unfortunately, it looks like doing so will cause conflicts, so those conflicts will need sorting out before we can accept and merge.

PR #1261, opened 2026-07-28
chore: remove imports files so they are not changed because of uids all the time

StraToN:
## Summary by CodeRabbit

* **Refactor**
  * Cleaned up legacy image asset metadata across character, item, and interface resources.
  * Updated worker animation definitions for improved compatibility with current project formats.
* **Chores**
  * Added an AI-related project file to the ignored files list.
  * Removed obsolete generated asset configuration files to keep project contents leaner.


BHSDuncan:
@StraToN I'm not sure if this is the right approach. I'm pretty sure `.import` files are necessary and should be committed alongside everything else as removing it can break things after pulling (e.g. broken scene files). 

Admittedly, it looks to be a known limitation: https://godotengine.org/article/uid-changes-coming-to-godot-4-4/

Here's a quick result from a Google query:

```
How to Properly Handle UID and Git Churn

The constant shifting of UIDs in Git is a known pain point, usually caused by multiple team members generating files on different machines or external file management.

Commit them anyway: According to the official Godot Engine Article, you must track and commit all .import and .uid files. Only the .godot/ folder itself should be ignored.

Only use the Godot FileSystem dock: Moving, renaming, or deleting assets outside of the engine (e.g., via Windows Explorer or VS Code) splits the asset from its .import/.uid file, forcing Godot to regenerate mismatched UIDs later. Doing these actions strictly inside Godot's editor keeps everything synced perfectly.

Re-save to fix lingering issues: If you are already seeing an endless cascade of UID warnings, open the broken scenes or resources in the Godot editor, make a dummy tweak, and Save All. This forces the engine to flush memory caches and write the stable UIDs explicitly down to disk.
```

BHSDuncan (approved):

PR #1263, opened 2026-08-05
fix: remove this init 'patch' as it doesn't work with statically-typed arrays

BHSDuncan:
…cally-typed arrays in the latest versions of Godot

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when selecting character images in the Character Creator.
  * Improved handling of main-menu confirmation actions in the character workflow.
  * Preserved existing animation settings during character export, reducing the risk of losing direction, idle, speech, or angle configuration.


StraToN (approved):
LGTM

